PR #2, opened 2024-08-13
feat: setup starter gf-1

liza-veis:
Project name placeholders:

- `ProjectName`
- `project-name`
- `project-name-prefix`


.ls-lint.yml
```diff
@@ -0,0 +1,25 @@
+ls:
+  .dir: regex:([.a-z0-9]*)([-.][a-z0-9]+)*
```
what1s1ove, line 2: now you can avoid it 

https://github.com/what1s1ove/whatislove.dev/blob/main/.ls-lint.yml

.ls-lint.yml
```diff
@@ -0,0 +1,25 @@
...
+    .ts: snake_case
+  .github/ISSUE_TEMPLATE:
+    .dir: SCREAMING_SNAKE_CASE
+  apps/**:
+    CHANGELOG.md: SCREAMING_SNAKE_CASE
+  packages/**:
+    CHANGELOG.md: SCREAMING_SNAKE_CASE
+
+ignore:
+  - .git
```
what1s1ove, line 17: ignores as well

what1s1ove (approved):

.ls-lint.yml
liza-veis: This regexp matches kebab-case but also allows dots. I can change it for the .dir, but then we will need to override this for some folders like .github, and for files we will need to leave it because of .helper etc. extentions

liza-veis: updated config

what1s1ove (approved):

PR #28, opened 2024-08-19
feat: add common color variables gf-27

AnutkaGn:
### Summary

This PR implements the common color variables from the design

### DIFF
--- a/apps/frontend/src/assets/css/variables.css
+++ b/apps/frontend/src/assets/css/variables.css
@@ -1,0 +1,25 @@
+:root {
+	/* Background Colors */
+	--background-primary: #121415;
+	--background-secondary: #1d1e21;
+	--background-hover: rgba(255 255 255 / 6%);
+	--border-secondary: #303030;
+	--border-primary: #3e424d;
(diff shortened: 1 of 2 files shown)


apps/frontend/src/assets/css/variables.css
```diff
@@ -0,0 +1,25 @@
+:root {
+	/* Background Colors */
+	--background-primary: #121415;
```
liza-veis, line 3: Let's add a color prefix for all color variables `--color-background-primary`

apps/frontend/src/assets/css/variables.css
```diff
@@ -0,0 +1,25 @@
...
+	--background-secondary: #1d1e21;
+	--background-hover: rgba(255 255 255 / 6%);
```
liza-veis, line 5: There is an ability to add alpha values to hex color, let's use it to keep everything the same style

liza-veis (changes requested):

liza-veis (approved):

what1s1ove (approved):

yevhenii-hladkov (approved):

PR #29, opened 2024-08-19
feat: create loader component gf-9

perekotypole:
Implemented a loader component:
- pulsing animation (similar to https://www.davidhu.io/react-spinners/storybook/?path=/docs/puffloader--docs)

image

### DIFF
--- a/apps/frontend/src/libs/components/app/app.tsx
+++ b/apps/frontend/src/libs/components/app/app.tsx
@@ -1,5 +1,5 @@
 import reactLogo from "~/assets/images/react.svg";
-import { Link, RouterOutlet } from "~/libs/components/components.js";
+import { Link, Loader, RouterOutlet } from "~/libs/components/components.js";
 import { AppRoute } from "~/libs/enums/enums.js";
 import {
 	useAppDispatch,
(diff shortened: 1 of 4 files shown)


apps/frontend/src/libs/components/loader/loader.tsx
liza-veis, line 1: Can we add something more engaging, like this?

Screenshot 2024-08-19 at 11 50 07

liza-veis (changes requested):

apps/frontend/src/libs/components/loader/loader.tsx
perekotypole: I have changed it to another kind of animation (link and screenshot in the PR description). If it's not okay, I will replace it to spinner

liza-veis: Cool, I like the new one

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -47,11 +47,16 @@ const App = (): JSX.Element => {
 				<>
 					<h2>Users:</h2>
 					<h3>Status: {dataStatus}</h3>
-					<ul>
-						{users.map((user) => (
-							<li key={user.id}>{user.email}</li>
-						))}
-					</ul>
+
+					{dataStatus === "pending" ? (
```
liza-veis, line 51: move this to a constant
```suggestion
					{isLoading ? (
```

apps/frontend/src/libs/components/loader/styles.module.css
```diff
@@ -0,0 +1,62 @@
+:root {
+	--size: 50px;
+	--weight: 5px;
+	--animation-time: 2s;
+}
```
liza-veis, line 5: Let's remove variables that are used only once; for animation-duration, it's better to keep this in .loader scope, not in the root
```suggestion
.loader {
	--animation-duration: 2s;
}
```

liza-veis (changes requested):

liza-veis (approved):

what1s1ove (approved):

PR #31, opened 2024-08-19
feat: add route protection gf-14

sandrvvu:
Summary
Created a ProtectedRoute wrapper component that redirects unauthorized users from protected pages to the Sign In page.

### DIFF
--- a/apps/frontend/src/libs/components/protected-route/protected-route.tsx
+++ b/apps/frontend/src/libs/components/protected-route/protected-route.tsx
@@ -1,0 +1,24 @@
+import { type FC, type ReactNode } from "react";
+import { Navigate } from "react-router-dom";
+
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
+import { DataStatus } from "~/libs/enums/data-status.enum.js";
+import { useAppSelector } from "~/libs/hooks/hooks.js";
+
+type ProtectedRouteProperties = {
+	children: ReactNode;
+};
+
+const ProtectedRoute: FC<ProtectedRouteProperties> = ({
+	children,
+}: ProtectedRouteProperties): JSX.Element => {
+	const authStatus = useAppSelector((state) => state.auth.dataStatus);
+
+	return authStatus === DataStatus.FULFILLED ? (
+		<>{children}</>
+	) : (
+		<Navigate replace to={AppRoute.SIGN_IN} />
+	);
+};
+
+export { ProtectedRoute };
--- a/apps/frontend/src/libs/components/protected-route/protected-route.tsx
+++ b/apps/frontend/src/libs/components/protected-route/protected-route.tsx
@@ -15,7 +15,7 @@ const ProtectedRoute: FC<ProtectedRouteProperties> = ({
 		(state) => state.auth.authenticatedUser,
 	);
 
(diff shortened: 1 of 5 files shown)


apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,24 @@
...
+type ProtectedRouteProperties = {
+	children: ReactNode;
+};
+
+const ProtectedRoute: FC<ProtectedRouteProperties> = ({
+	children,
+}: ProtectedRouteProperties): JSX.Element => {
+	const authStatus = useAppSelector((state) => state.auth.dataStatus);
+
+	return authStatus === DataStatus.FULFILLED ? (
```
liza-veis, line 17: Let's check it by user presence, not status. Because there could be other conditions when the status is fulfilled. We will store it in `auth.authenticatedUser` in store

liza-veis (changes requested):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
liza-veis: Yes, you can, use SignUpResponseDto as a type

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const ProtectedRoute: FC<ProtectedRouteProperties> = ({
+	children,
+}: ProtectedRouteProperties): JSX.Element => {
+	const authenticatedUser = useAppSelector(
+		(state) => state.auth.authenticatedUser,
+	);
+
+	return authenticatedUser ? (
```
liza-veis, line 18:
```suggestion
    const hasAuthenticatedUser = Boolean(authenticatedUser);

	return hasAuthenticatedUser ? (
```

liza-veis (changes requested):

what1s1ove (approved):

liza-veis (approved):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,27 @@
+import { type FC, type ReactNode } from "react";
+import { Navigate } from "react-router-dom";
+
+import { AppRoute } from "~/libs/enums/app-route.enum.js";
+import { useAppSelector } from "~/libs/hooks/hooks.js";
+
+type ProtectedRouteProperties = {
```
liza-veis, line 7: It is named as properties in other places
```suggestion
type Properties = {
```

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,27 @@
...
+	children: ReactNode;
+};
+
+const ProtectedRoute: FC<ProtectedRouteProperties> = ({
```
liza-veis, line 11: We use `({}: Properties): JSX.Element =>` notation, there is no need for both FC and JSX.Element 
```suggestion
const ProtectedRoute = ({
```

liza-veis (changes requested):

liza-veis (approved):

PR #34, opened 2024-08-19
feat: generate authorization token gf-8

Vitaliistf:
Implemented authorization token generation. As sign-up and sign-in flows are not fully developed, left token generation only for existing sign-up flow, in order to not clutter the project, writing unuseful code. When flows are developed, devs might use existing line from sign-up method in auth service to work with JWT. Also added environment variables for JWT for flexible setting up.


apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,11 +1,12 @@
+import { type UserResponseDto } from "@git-fit/shared/src/modules/users/libs/types/user-response-dto.type.js";
```
yevhenii-hladkov, line 1: import { type UserResponseDto } from "@git-fit/shared

yevhenii-hladkov, line 1: Better to import it from ./libs/types/types.js

Vitaliistf: Yeah, thanks, forgot about it) Fixed!

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -4,17 +4,26 @@ import {
...
 	}
 
-	public signUp(
+	public async signUp(
 		userRequestDto: UserSignUpRequestDto,
 	): Promise<UserSignUpResponseDto> {
-		return this.userService.create(userRequestDto);
+		// the mock of the method. May be changed according to required sign-up flow
+		const user = await this.userService.create(userRequestDto);
+		// next line should be used to create token also in the sign-in method
```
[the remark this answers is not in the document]
what1s1ove: i would say that we can avoid such comments 😅

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,11 +1,12 @@
+import { type UserResponseDto } from "@git-fit/shared/src/modules/users/libs/types/types.js";
```
what1s1ove, line 1: fix import

what1s1ove (changes requested):

apps/backend/src/modules/auth/auth.service.ts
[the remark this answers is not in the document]
Vitaliistf: So what should I do with this comment? Remove or change to Liza's suggestion?

apps/backend/src/modules/auth/token/token.service.ts
```diff
@@ -0,0 +1,27 @@
...
+class TokenService {
+	private expirationTime: string;
+	private secret: Uint8Array;
+
+	public constructor(secret: string, expirationTime: string) {
+		this.secret = new TextEncoder().encode(secret);
+		this.expirationTime = expirationTime;
+	}
+
+	public async createToken(userId: number): Promise<string> {
```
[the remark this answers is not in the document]
Vitaliistf: I changed this method to a generic one (but wrote generic like this <T extends JWTPayload>, to make it work), and I have a linter error (Type parameter T is used only once in the function signature), should I just ignore it?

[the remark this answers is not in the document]
liza-veis: I think this means, you can use (payload: JWTPayload) instead of generic

apps/backend/src/modules/auth/auth.service.ts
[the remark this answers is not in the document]
liza-veis: I think this commit can be omitted, let's remove it

apps/backend/src/modules/auth/auth.ts
```diff
@@ -1,10 +1,17 @@
+import { config } from "~/libs/modules/config/config.js";
 import { logger } from "~/libs/modules/logger/logger.js";
 import { userService } from "~/modules/users/users.js";
 
+import { TokenService } from "../../libs/modules/token/token.service.js";
```
liza-veis, line 5: let's rename it to `token`, and it should be created the same way as the logger or config module. We should re-export a single instance from this module 
```suggestion
import { token } from "~/libs/modules/token/token.js";
```

liza-veis (changes requested):

apps/backend/src/modules/auth/auth.ts
Vitaliistf: Should I create also a libs/types folder, and declare types for this service? If so, should I name that type as a Token, or a TokenService?

Vitaliistf: Fixed. Also added types for this module, as in the logger.

apps/backend/src/libs/modules/token/libs/types/token.type.ts
```diff
@@ -0,0 +1,12 @@
+import { type JWTPayload } from "jose";
```
what1s1ove, line 1: each type should be in a separate file. or write inline function signatures (best choice)

apps/backend/src/libs/modules/token/token.ts
```diff
@@ -0,0 +1,11 @@
+import { config } from "../config/config.js";
+import { TokenService } from "./token.service.js";
+
+const token = new TokenService(
+	config.ENV.JWT.SECRET,
```
what1s1ove, line 5: pass as an object

what1s1ove (changes requested):

apps/backend/src/libs/modules/token/libs/types/token.type.ts
Vitaliistf: I just saw that in the logger it's done like this, so I thought that this would be the best approach. I'll change it, thanks.

apps/backend/src/libs/modules/token/token.ts
Vitaliistf: Should I create a type for this object? If so, should I put it in token.service.js?

apps/backend/src/libs/modules/token/libs/types/token.type.ts
liza-veis: @what1s1ove I think it's ok here, it is not a common type, and that's the only place it is used

apps/backend/src/libs/modules/token/token.service.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { type Token } from "./token.js";
+
+type TokenServiceOptions = {
+	algorithm: string;
+	expirationTime: string;
+	secret: string;
+};
+
+class TokenService implements Token {
```
liza-veis, line 11:
```suggestion
class BaseToken implements Token {
```

apps/backend/src/libs/modules/token/token.service.ts
```diff
@@ -0,0 +1,41 @@
+import { type JWTPayload, jwtVerify, SignJWT } from "jose";
+
+import { type Token } from "./token.js";
+
+type TokenServiceOptions = {
```
liza-veis, line 5:
```suggestion
type Constructor = {
```

apps/backend/src/libs/modules/token/token.service.ts
```diff
@@ -0,0 +1,41 @@
+import { type JWTPayload, jwtVerify, SignJWT } from "jose";
+
+import { type Token } from "./token.js";
```
liza-veis, line 3: this type should be in token/libs/types/types
```suggestion
import { type Token } from "./libs/types/types.js";
```

liza-veis (changes requested):

liza-veis (approved):

jhladkov (approved):

yevhenii-hladkov (approved):

what1s1ove (approved):

PR #37, opened 2024-08-19
feat: add sign-up flow gf-5

kolibri753:
1. Refactored form validation to include the name property, along with appropriate validation rules and messages.
2. Implemented the design from Figma into the sign-up form, including icon support and styles for common components (button, input).
3. Redirected users to the root path upon successful sign-up and added "Log in" link navigation to the sign-in page.
4. Stored authenticated user data in the Redux store.

Not sure about the logo and bg yet
image

image


apps/frontend/src/assets/icons/eye.png
liza-veis, line 1: Let's keep icons in `assets/images/icons` folder. Also please download them as svg, remove fill or stroke from them, or set it to currentColor, so that we can import them as svg elements and set color using css `color` property. We also will need to create a common Icon component for this.

liza-veis (changes requested):

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -18,36 +23,70 @@ const SignUpForm = ({ onSubmit }: Properties): JSX.Element => {
...
+			void handleSubmit(async (formData: UserSignUpRequestDto) => {
+				await onSubmit(formData);
+				navigate(AppRoute.ROOT);
+			})(event_);
 		},
-		[handleSubmit, onSubmit],
+		[handleSubmit, onSubmit, navigate],
 	);
 
+	const togglePasswordVisibility = useCallback(() => {
```
what1s1ove, line 39: A10

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -20,8 +20,12 @@ const Auth = (): JSX.Element => {
 	}, []);
 
 	const handleSignUpSubmit = useCallback(
-		(payload: UserSignUpRequestDto): void => {
-			void dispatch(authActions.signUp(payload));
+		async (payload: UserSignUpRequestDto): Promise<void> => {
+			const result = await dispatch(authActions.signUp(payload));
+
+			if (result.meta.requestStatus !== "fulfilled") {
```
what1s1ove, line 26: you do not need to handle an error here

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -7,33 +7,46 @@ import {
...
+		<label className={styles["input-label"]}>
+			<span className={styles["input-label-text"]}>{label}</span>
+			<div className={styles["input-container"]}>
+				<input
+					{...field}
+					className={styles["input-field"]}
+					placeholder={placeholder}
+					type={type}
+				/>
+				{icon && <div className={styles["input-icon"]}>{icon}</div>}
```
what1s1ove, line 45: we should create the `Icon` component for this

what1s1ove (changes requested):

apps/frontend/src/assets/icons/eye.png
kolibri753: I'm on it. I'll download the icons as SVGs, adjust them for use with CSS, and I'll also include the common Icon component in this PR.

apps/frontend/src/pages/auth/auth.tsx
kolibri753: should I just leave it like that?
```
async (payload: UserSignUpRequestDto): Promise<void> => {
	await dispatch(authActions.signUp(payload));
},
```

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,29 @@
+import { type SVGProps } from "react";
+
+const Eye = (properties: SVGProps<SVGSVGElement>): JSX.Element => (
```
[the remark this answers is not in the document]
kolibri753: thank you for example :)
should I split it into two different files (icon-map.ts, icon.tsx), and if yes - where should I put icon-map.ts?

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+	strikedEye: StrikedEye,
+};
+
+interface Properties {
+	className?: string | undefined;
+	name: IconName;
+	onClick?: () => void;
+}
+
+const Icon: FC<Properties> = ({ className, name, onClick }: Properties) => {
```
GvoFor, line 21: Other components declared in the following way: `const Component = ({ ... }: Properties): JSX.Element => {` so I think it's better to do the same for your `Icon` to keep code consistent

package.json
```diff
@@ -63,10 +63,14 @@
 		"stylelint-config-recess-order": "5.0.1",
 		"stylelint-config-standard": "36.0.1",
 		"tsc-alias": "1.8.10",
-		"typescript": "5.5.4"
+		"typescript": "5.5.4",
+		"vite-plugin-svgr": "4.2.0"
```
GvoFor, line 67: Not sure, if you installed `vite-plugin-svgr` into the correct place: do we need it for both frontend and backend?! and inot sure if it is a `dev dependency`: shouldn't it be in regular `dependencies` list?

package.json
```diff
@@ -63,10 +63,14 @@
...
 	},
 	"simple-git-hooks": {
 		"pre-commit": "npm run lint:type && npm run git:hook:precommit",
 		"commit-msg": "npm run git:hook:commit"
+	},
+	"dependencies": {
+		"clsx": "2.1.1"
```
GvoFor, line 74: Same for `clsx`: do we need it for both frontend and backend?

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { type FC } from "react";
+
+import Eye from "~/assets/images/icons/eye.svg?react";
+import StrikedEye from "~/assets/images/icons/striked-eye.svg?react";
+
+type IconName = "eye" | "strikedEye";
+
+const iconNameToSvg: Record<IconName, FC<React.SVGProps<SVGSVGElement>>> = {
+	eye: Eye,
```
GvoFor, line 11: I think `IconName` and `iconNameToSvg` should be moved from this file. That's because, they will be changed in the future (cause we will need to add new icons), and if we keep them here, there will be a possibility to break Icon component. So without them Icon component will be fully (at least for now) done and there will be no problem to safely and easily add new icons.

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -18,36 +21,76 @@ const SignUpForm = ({ onSubmit }: Properties): JSX.Element => {
...
+				navigate(AppRoute.ROOT);
+			})(event_);
 		},
-		[handleSubmit, onSubmit],
+		[handleSubmit, onSubmit, navigate],
 	);
 
+	const handleTogglePasswordVisibility = useCallback(() => {
+		setIsPasswordVisible((previousState) => !previousState);
+	}, []);
```
GvoFor, line 39: Couldn't we use here `setIsPasswordVisible(!isPasswordVisible);`

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+	strikedEye: StrikedEye,
+};
+
+interface Properties {
```
what1s1ove, line 15: use a `type` instead for props

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -1,3 +1,4 @@
+import clsx from "clsx";
```
what1s1ove, line 1: create a new helper `getValidaClasses` for it

what1s1ove (changes requested):

apps/frontend/src/libs/components/icon/icon.tsx
kolibri753: Where do you think is the best place to store them?

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
kolibri753: Some problems with ESLint. https://stackoverflow.com/questions/36677733/why-shouldnt-jsx-props-use-arrow-functions-or-bind

apps/frontend/src/libs/components/icon/icon.tsx
GvoFor: For type `IconName`, the best I think would be:
`apps/frontend/src/libs/components/icon/types/types.ts` and `apps/frontend/src/libs/components/icon/types/icon-name.type.ts`
And for `iconNameToSvg` I'm not sure. But maybe good place can be:
`apps/frontend/src/libs/components/icon/helpers/helpers.ts` and `apps/frontend/src/libs/components/icon/helpers/icon-name-to-svg.helper.ts`

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
GvoFor: Then let's leave it as it is

apps/frontend/src/libs/components/button/button.module.css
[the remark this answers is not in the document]
kolibri753: I noticed that someone had already changed it, so I thought I would just merge the changes later

apps/frontend/src/libs/components/input/input.tsx
kolibri753: I'm not sure I understand correctly why do we need it. For readability improve? (for me it's so much easier to read the first version with filter and join) 

To refactor this?
```
const inputClassNames = clsx(
	styles["input-field"],
	leftIcon && styles["with-left-icon"],
	rightIcon && styles["with-right-icon"],
);
```

Should it be something like this?
```
import clsx from "clsx";

type Properties = {
	hasError: boolean;
};

const getValidClassNames = ({ hasError }: Properties): string => {
	return clsx({
		"is-invalid": hasError,
		"is-valid": !hasError,
	});
};

export { getValidClassNames };
```

liza-veis: no, this should be just a wrapper for clsx, like
```javascript
const getValidClassNames = (...inputs) => {
	return clsx(...inputs);
};
```

what1s1ove: I meant something like this 🙂 Instead of importing `clsx ` all the time 

```ts
import { type ClassValue, clsx } from "clsx";

const getValidClassNames = (...inputs: ClassValue[]): string => {
	return clsx(...inputs);
};

export { getValidClassNames };
```

apps/frontend/src/pages/auth/auth.tsx
what1s1ove:
```
async (payload: UserSignUpRequestDto): void => {
	await dispatch(authActions.signUp(payload));
},
```

like this one

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { type FC } from "react";
+
+import Eye from "~/assets/images/icons/eye.svg?react";
+import StrikedEye from "~/assets/images/icons/striked-eye.svg?react";
+
+import { mapIconNameToSvg } from "./helpers/helpers.js";
+import { type IconName, type Icon as Properties } from "./types/types.js";
+
+const iconNameToSvg = mapIconNameToSvg<IconName>({
```
liza-veis, line 11: This map should be in `icon/libs/maps/icon-name-to-svg.map.ts`, also we don't need `mapIconNameToSvg` mapper.

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,22 @@
...
+	eye: Eye,
+	strikedEye: StrikedEye,
+});
+
+const Icon: FC<Properties> = ({ className, name, onClick }) => {
```
liza-veis, line 16: we use this notation - `const Icon = ({}: Properties): JSX.Element =>`

apps/frontend/src/libs/components/icon/types/icon.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type IconName } from "./icon-name.type.js";
+
+type Icon = {
+	className?: string | undefined;
+	name: IconName;
+	onClick?: () => void;
```
liza-veis, line 6: icon shouldn't have className or onClick. If you need to modify the width or something else, pass a specific property. We can add onClick only to interactive elements like buttons, so if you need this, you need to create a separate component like IconButton.   

```suggestion
	name: IconName;
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -5,35 +5,79 @@ import {
...
+		rightIcon && styles["with-right-icon"],
+	);
+
 	return (
-		<label>
-			<span>{label}</span>
-			<input {...field} placeholder={placeholder} type={type} />
-			{hasError && <span>{error as string}</span>}
+		<label className={styles["input-label"]}>
+			<span className={styles["input-label-text"]}>{label}</span>
```
liza-veis, line 46: I think we don't need span here, we can use only label

apps/frontend/src/libs/components/icon/icon.tsx
GvoFor: Oh, about `iconNameToSvg`. I mean that you should move whole object and not introduce that identity function ;)
So the `icon-name-to-svg.helper.ts` should looks like
```ts
/// <reference types="vite-plugin-svgr/client" />

import { type FC } from "react";

import Eye from "~/assets/images/icons/eye.svg?react";
import StrikedEye from "~/assets/images/icons/striked-eye.svg?react";

import { type IconName } from "../types/types.js";

const iconNameToSvg: Record<IconName, FC<React.SVGProps<SVGSVGElement>>> = {
	eye: Eye,
	strikedEye: StrikedEye,
};

export { iconNameToSvg };
```
and then you can import and use this map (i.e. `iconNameToSvg`) inside `icon.tsx` file like:

```ts
import { iconNameToSvg } from "./helpers/helpers.js";
import { type IconName } from "./types/types.js";

type Properties = {
  className?: string;
  name: IconName;
  onClick?: () => void;
};

const Icon = ({ className, name, onClick }: Properties): JSX.Element => {
  const SvgComponent = iconNameToSvg[name];
  
  return <SvgComponent className={className} onClick={onClick} />
};

export { Icon };
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -5,35 +5,79 @@ import {
...
+							styles["input-icon"],
+							styles["input-icon-left"],
+						)}
+					>
+						{leftIcon}
+					</div>
+				)}
+
+				<input
+					{...field}
```
liza-veis, line 60: pass the properties directly
```suggestion
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -5,35 +5,79 @@ import {
...
+
 	return (
-		<label>
-			<span>{label}</span>
-			<input {...field} placeholder={placeholder} type={type} />
-			{hasError && <span>{error as string}</span>}
+		<label className={styles["input-label"]}>
+			<span className={styles["input-label-text"]}>{label}</span>
+			<div className={styles["input-container"]}>
+				{leftIcon && (
```
liza-veis, line 48:
```suggestion
				{hasLeftIcon && (
```

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -5,35 +5,79 @@ import {
...
+				)}
+
+				<input
+					{...field}
+					className={inputClassNames}
+					placeholder={placeholder}
+					type={type}
+				/>
+
+				{rightIcon && (
```
liza-veis, line 66:
```suggestion
				{hasRightIcon && (
```

apps/frontend/src/libs/components/icon/helpers/icon-name-to-svg.helper.ts
GvoFor, line 1: Rewrite this file, as I mentioned in another comment. In that case, when we gonna add new icon, we do not need to change `icon.tsx` -- we need only change this and `icon-name.type.ts` files

apps/frontend/src/pages/auth/components/sign-up-form/styles.module.css
```diff
@@ -1,5 +1,76 @@
-.form {
+.auth-container {
+	display: flex;
+	width: 100vw;
+	height: 100vh;
```
liza-veis, line 4: avoid using vh, let's try to achieve it using 100% or flex
```suggestion
```

apps/frontend/src/pages/auth/components/sign-up-form/styles.module.css
```diff
@@ -1,5 +1,76 @@
...
+}
+
+.form-text > a {
+	color: var(--color-brand-primary);
+	transition: all 0.3s ease;
+}
+
+.form-text > a:hover {
+	color: var(--color-brand-primary-hover);
+}
```
liza-veis, line 61: This is a common style for link, apply it to common link component
```suggestion
```

apps/frontend/src/libs/components/icon/icon.tsx
GvoFor: Also there is no need for `Icon` type, so in `icon.type.ts` file. To keep project consistent, you better keep here `type Properties` (as you can see in my code snippet), and also take a look at component declaration. It must looks like: `const Icon = ({ className, name, onClick }: Properties): JSX.Element => {`.

apps/frontend/src/libs/components/icon/types/icon.type.ts
GvoFor, line 1: No need for this file. Look at some of my comments

apps/frontend/src/libs/components/icon/types/types.ts
GvoFor, line 1: Do not forget to remove `type Icon` export )

liza-veis:
Please remove boilerplate code for `pages/auth/auth.tsx`

apps/frontend/src/libs/components/input/input.tsx
kolibri753: Do you mean like this?
```
<label className={styles["input-label"]}>
	{label}
	...
</label>
```
Is it a better practice? I understand that less html is generally better, but for styling purposes - having an extra element can be more more appealing when debugging. Alternatively, is it better to pass styles to the container (label) and overwrite them? I don't think the difference in performance would be noticeable.

apps/frontend/src/libs/components/icon/libs/maps/icon-name-to-svg.map.ts
```diff
@@ -0,0 +1,15 @@
+/// <reference types="vite-plugin-svgr/client" />
```
what1s1ove, line 1: you need to add it to apps/frontend/src/vite-env.d.ts instead

apps/frontend/src/pages/auth/components/sign-up-form/styles.module.css
kolibri753: In our project common link is NavLink, I thought it would be better to pass from a parent. What is the best approach in this case?

apps/frontend/src/pages/auth/components/sign-up-form/styles.module.css
kolibri753: I'm not sure why, but it's not working as expected.
The .auth-container has a height of 100%, but the .left-side and .right-side are not taking the full height.
Here is the code for the container:
```
.auth-container {
	display: flex;
	flex-grow: 1;
	align-items: stretch;
	width: 100%;
	height: 100%;
}
```
And for the left and right sides:
```
.left-side,
.right-side {
	display: flex;
	width: 50%;
	height: 100%;
}
```
I've also verified body and root:
```
body {
	display: flex;
	flex-direction: column;
	min-width: 320px;
	min-height: 100vh;
	margin: 0;
	font-family: Inter, Arial, sans-serif;
	font-size: 14px;
	line-height: 1.5;
}

#root {
	display: flex;
	flex-direction: column;
	flex-grow: 1;
}
```

kolibri753: Maybe your trained eye already know the answer :sweat_smile:

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -2,10 +2,12 @@
 	padding: 10px 16px;
 	font-size: 16px;
 	font-weight: 500;
+	line-height: 24px;
```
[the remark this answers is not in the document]
kolibri753: Oh.. on web figma and linux version line-height is in px. Or am I looking the wrong way 🤔 ?

apps/frontend/src/libs/components/icon/icon.tsx
kolibri753: I've tried to implement what @GvoFor adviced, can you check it out?

apps/frontend/src/libs/components/button/styles.module.css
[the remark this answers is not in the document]
liza-veis: Does it display it in pixels here for you?
Screenshot 2024-08-21 at 21 39 12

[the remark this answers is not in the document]
kolibri753: My bad, it was in styles... will check ✅

apps/frontend/src/libs/components/input/input.tsx
kolibri753: thank you for this comprehensive answer, really appreciate this 🐾

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -1,5 +1,6 @@
 .button {
 	padding: 10px 16px;
+	font-family: Inter, sans-serif;
```
kolibri753, line 3: Failed to decode the downloaded font (http://localhost:3000/assets/fonts/inter/inter-medium.tt)

P.S. The regular and semi-bold are working fine.

what1s1ove (approved):

apps/frontend/src/libs/components/button/styles.module.css
kolibri753: It was a typo mistake -> src: url("/assets/fonts/inter/inter-medium.tt") format("truetype");
Should be -> src: url("/assets/fonts/inter/inter-medium.ttf") format("truetype");

Can I fix this in this PR?

liza-veis: yes, please

liza-veis (changes requested):

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,20 @@
...
+type Properties = {
+	height?: string;
+	name: IconName;
+	width?: string;
+};
+
+const Icon = ({
+	height = "22px",
+	name,
+	width = "22px",
```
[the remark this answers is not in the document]
kolibri753: Should I better delete default values of width and height here, or create constraints for that in this folder?

P.S. [eslint/no-magic-numbers](https://typescript-eslint.io/rules/no-magic-numbers/)

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -40,12 +38,7 @@ const Auth = (): JSX.Element => {
 		return null;
 	};
 
-	return (
-		<>
-			State: {dataStatus}
-			{handleScreenRender(pathname)}
-		</>
-	);
+	return <main>{handleScreenRender(pathname)}</main>;
```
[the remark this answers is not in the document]
liza-veis: Yes, add everything to make it look like on design

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -18,36 +27,78 @@ const SignUpForm = ({ onSubmit }: Properties): JSX.Element => {
...
-					name="password"
-					placeholder="••••••••"
-					type="text"
-				/>
-				<Button label="Sign up" type="submit" />
-			</form>
-		</>
+		<section className={styles["auth-container"]}>
+			<div className={styles["left-side"]}>
+				{/* //TODO: replace logo */}
```
liza-veis, line 50:
```suggestion
				{/* TODO: replace logo */}
```

apps/frontend/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type Properties = {
+	className?: string | undefined;
+	icon: JSX.Element;
+	label: string;
+	onClick: () => void;
+};
+
+const IconButton = ({
+	className = "",
```
liza-veis, line 13: We don't want to pass className to common components, in case we need something, we will pass it a specific prop
```suggestion
```

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
```diff
@@ -18,36 +27,78 @@ const SignUpForm = ({ onSubmit }: Properties): JSX.Element => {
...
+					<div className={styles["password-container"]}>
+						<Input
+							autoComplete="one-time-code"
+							control={control}
+							errors={errors}
+							label="Password"
+							name="password"
+							rightIcon={
+								<IconButton
+									icon={
```
liza-veis, line 84: We need to pass icon name to IconButton  and properties like width, height, color, not the whole Icon component

liza-veis (changes requested):

apps/frontend/src/libs/components/icon/icon.tsx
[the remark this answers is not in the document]
liza-veis: let's remove default values

apps/frontend/src/libs/components/icon-button/icon-button.tsx
kolibri753: Noted :notebook:

kolibri753: I just thought that maybe we needed the option to change the button background and so on

apps/frontend/src/pages/auth/components/sign-up-form/sign-up-form.tsx
kolibri753: So I assume, that I should also add a color property to Icon
```
type Properties = {
	height?: number;
	name: IconName;
	width?: number;
};
```
and height + width are mandatory (should delete "?")

GvoFor (commented):
All looks fine

apps/frontend/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,41 @@
+import { Icon } from "~/libs/components/components.js";
+
+import { type IconName } from "../icon/types/types.js";
+import styles from "./styles.module.css";
+
+type Properties = {
+	iconColor?: string | undefined;
```
GvoFor, line 7: Do we need explicitly write type as `string | undefined`? maybe better use just `string` as you do in `Input` component?

apps/frontend/src/libs/components/icon/icon.tsx
```diff
@@ -0,0 +1,17 @@
+import { iconNameToSvg } from "./libs/maps/maps.js";
+import { type IconName } from "./types/types.js";
+
+type Properties = {
+	color?: string | undefined;
```
GvoFor, line 5: Do we need explicitly write type as `string | undefined`? maybe better use just `string` as you do in `Input` component?

apps/frontend/src/libs/components/icon-button/icon-button.tsx
GvoFor, line 1: Can you clarify please, what `IconButton` component stands for? does it represent "clickable icon" or "button with icon inside"?

apps/frontend/src/libs/components/icon-button/icon-button.tsx
GvoFor: @liza-veis

kolibri753: I can redo using 

color={iconColor ?? ""}

GvoFor: > I can also redo it using color={iconColor ?? ""}.

No-no. My question is about what is better:
```ts
type Properties = {
  iconColor?: string;
  ...
}
```
or
```ts
type Properties = {
  iconColor?: string | undefined;
  ...
}
```

apps/frontend/src/libs/components/icon/icon.tsx
kolibri753: Yes, here I can leave just `string`, but in this case I should also redo icon-button `using color={iconColor ?? ""}`

kolibri753: I see it more as a matter of preference, so I'm not sure if I should redo it.

apps/frontend/src/libs/components/icon-button/icon-button.tsx
GvoFor: I'm asking because in case it is "clickable icon", then your styles and props make sense, but if it is "button with icon" (like 
image ) then we need to style border for this element and also there will be needs for button dimensions instead of icon dimensions

liza-veis: It's a button with icon, but we will add other styles like border later

apps/frontend/src/libs/components/icon-button/icon-button.tsx
kolibri753: If icon has `color?: string;` and `iconColor?: string;`, then Icon inside IconButton 
```
Type '{ color: string | undefined; height: number; name: IconName; width: number; }' is not assignable to type 'Properties' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'color' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.ts(2375)
```

liza-veis (approved):

apps/frontend/src/libs/components/icon-button/icon-button.tsx
GvoFor: Got it

apps/frontend/src/libs/components/icon/icon.tsx
GvoFor: In this case, leave color type as it is

PR #38, opened 2024-08-19
feat: add sign-up flow gf-6

s1rserg:
1. Added name column to users table through migration.
2. Updated model and service in backend folder.

3. Added encryption service to hash user password before saving to the db. Salt is included with the hash by bcrypt, so idk what to do with password_salt field in database. Salt rounds variable is added to env variables.

4. Unique email constraint and `/auth/sign-up` endpoint were already implemented.
5. Returned object now has user property in response (password_hash and password_salt columns excluded).
6. As for validation, user sign-up validation schema in shared folder was updated.


s1rserg:
It seems that some checks failed due to the change of `UserSignUpRequestDto` in shared folder which triggered errors on front end part.

apps/backend/src/libs/modules/encryption/encryption.service.ts
```diff
@@ -0,0 +1,42 @@
...
+	public async comparePassword(
+		password: string,
+		hash: string,
+	): Promise<boolean> {
+		try {
+			return await bcrypt.compare(password, hash);
+		} catch (error) {
+			throw new ApplicationError({
+				cause: error,
+				message: "Error comparing passwords",
```
yevhenii-hladkov, line 25: Let's store error messages in enums

apps/backend/src/modules/users/user.service.ts
```diff
@@ -9,24 +11,42 @@ import {
...
 	}
 
 	public async create(
 		payload: UserSignUpRequestDto,
 	): Promise<UserSignUpResponseDto> {
+		const existingUser = await this.userRepository.findByEmail(payload.email);
+
+		if (existingUser) {
+			throw new ApplicationError({
+				message: "Email address is already in use.",
```
yevhenii-hladkov, line 32: Same here

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -38,6 +39,12 @@ class UserRepository implements Repository {
 		return users.map((user) => UserEntity.initialize(user));
 	}
 
+	public async findByEmail(email: string): Promise<null | UserEntity> {
+		const user = await this.userModel.query().where({ email }).first();
+
+		return user ? UserEntity.initialize(user) : null;
```
yevhenii-hladkov, line 45:
```suggestion
		return user && UserEntity.initialize(user);
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -38,6 +39,12 @@ class UserRepository implements Repository {
 		return users.map((user) => UserEntity.initialize(user));
 	}
 
+	public async findByEmail(email: string): Promise<null | UserEntity> {
```
yevhenii-hladkov, line 42: It's better to modify `find`method on 36 line and make it universal. For example, it takes query property, and then you use `findOne(query)`

apps/backend/src/modules/users/user.service.ts
```diff
@@ -9,24 +11,42 @@ import {
...
+		encryptionService: EncryptionService,
+	) {
 		this.userRepository = userRepository;
+		this.encryptionService = encryptionService;
 	}
 
 	public async create(
 		payload: UserSignUpRequestDto,
 	): Promise<UserSignUpResponseDto> {
+		const existingUser = await this.userRepository.findByEmail(payload.email);
```
yevhenii-hladkov, line 28:
```suggestion
    const {name, email, password} = payload
		const existingUser = await this.userRepository.find({ email });
```

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -3,6 +3,8 @@ import { UserValidationRule } from "./user-validation-rule.enum.js";
 const UserValidationMessage = {
 	EMAIL_REQUIRED: "Email is required.",
 	EMAIL_WRONG: "Email is wrong.",
+	NAME_REQUIRED: "Name is required.",
+	NAME_TOO_LONG: "Name is too long (> 100 symbols).",
```
yevhenii-hladkov, line 7: NAME_TOO_LONG: `Name is too long (> ${UserValidationRule.NAME_MAXIMUM_LENGTH} symbols).`,

s1rserg:
Addressed all suggested changes. Error messages are stored in enums. Find method is now universal. Error message for max symbols in name is changed.

liza-veis:
> Addressed all suggested changes. Error messages are stored in enums. Find method is now universal. Error message for max symbols in name is changed.

You need to adapt to these changes, because we can't merge PR without these checks passed

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,32 +1,54 @@
...
+		encryptionService: EncryptionService,
+	) {
 		this.userRepository = userRepository;
+		this.encryptionService = encryptionService;
 	}
 
 	public async create(
 		payload: UserSignUpRequestDto,
 	): Promise<UserSignUpResponseDto> {
+		const { email } = payload;
```
yevhenii-hladkov, line 29: You can distract also name and password from payload and use them below instead of payload.name, payload.password

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,32 +1,54 @@
...
+		);
+
 		const item = await this.userRepository.create(
 			UserEntity.initializeNew({
 				email: payload.email,
-				passwordHash: "HASH", // TODO
-				passwordSalt: "SALT", // TODO
+				name: payload.name,
+				passwordHash,
+				passwordSalt: "", // bcrypt handles salt internally
```
what1s1ove, line 45: we need to save the salt to db

what1s1ove (changes requested):

yevhenii-hladkov (approved):

package.json
```diff
@@ -38,6 +38,7 @@
 		"@eslint/js": "9.9.0",
 		"@ls-lint/ls-lint": "2.3.0-beta.1",
 		"@stylistic/eslint-plugin": "2.6.2",
+		"@types/bcrypt": "5.0.2",
```
liza-veis, line 41: this should be in apps/backend/packages.json

apps/backend/src/libs/modules/encryption/encryption.service.ts
```diff
@@ -0,0 +1,55 @@
...
+import { ApplicationError } from "~/libs/exceptions/exceptions.js";
+
+import { EncryptionServiceMessage } from "./libs/enums/enums.js";
+
+const DEFAULT_SALT_ROUNDS = 10;
+
+class EncryptionService {
+	private saltRounds: number;
+
+	public constructor() {
```
liza-veis, line 12: We should pass salt rounds as parameters to constructor. Also we don't use process.env directly, we have a config module for this.

apps/backend/src/libs/modules/encryption/encryption.service.ts
```diff
@@ -0,0 +1,55 @@
+import bcrypt from "bcrypt";
+
+import { ApplicationError } from "~/libs/exceptions/exceptions.js";
+
+import { EncryptionServiceMessage } from "./libs/enums/enums.js";
+
+const DEFAULT_SALT_ROUNDS = 10;
+
+class EncryptionService {
```
liza-veis, line 9: Let's name it encryption, and it should have the same structure as the logger module, with BaseEncryption class
```suggestion
class BaseEncryption implement Encryption {
```

apps/backend/src/libs/types/repository.type.ts
```diff
@@ -1,7 +1,7 @@
 type Repository<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(id: number): Promise<boolean>;
-	find(id: number): Promise<null | T>;
+	find(query: T): Promise<null | T>;
```
liza-veis, line 4: We will use id for find as a common method. If you need anything else this should be a separate method 
```suggestion
	find(id: number): Promise<null | T>;
```

apps/backend/src/modules/users/libs/enums/user-service-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const UserServiceMessage = {
```
liza-veis, line 1: Let's have a common ExceptionMessage enum, not per each service

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -77,10 +92,22 @@ class UserEntity implements Entity {
...
+
+	public toSignUpResponseDto(): UserSignUpResponseDto {
+		return {
+			user: {
+				email: this.email,
+				id: this.id as number,
+				name: this.name,
+			},
 		};
 	}
```
liza-veis, line 112: we don't need this, we will use toObject
```suggestion
```

apps/backend/src/modules/users/user.service.ts
```diff
@@ -1,32 +1,56 @@
...
 	}
 
 	public async create(
 		payload: UserSignUpRequestDto,
 	): Promise<UserSignUpResponseDto> {
+		const { email, name, password } = payload;
+		const existingUser = await this.userRepository.find({ email });
+
+		if (existingUser) {
+			throw new ApplicationError({
```
liza-veis, line 33: Create a custom UserError
```suggestion
			throw new UserError({
```

apps/backend/src/libs/modules/encryption/encryption.service.ts
```diff
@@ -0,0 +1,55 @@
...
+			return await bcrypt.genSalt(this.saltRounds);
+		} catch (error) {
+			throw new ApplicationError({
+				cause: error,
+				message: EncryptionServiceMessage.GENERATE_SALT,
+			});
+		}
+	}
+
+	public async hashPassword(password: string, salt: string): Promise<string> {
```
liza-veis, line 43: Encryption service should know nothing about password. It should have two methods - encrypt and compare. Salt should also be generated here

apps/backend/src/libs/modules/encryption/encryption.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+	public async generateSalt(): Promise<string> {
+		try {
+			return await bcrypt.genSalt(this.saltRounds);
+		} catch (error) {
+			throw new ApplicationError({
+				cause: error,
+				message: EncryptionServiceMessage.GENERATE_SALT,
+			});
+		}
```
liza-veis, line 40: There is no need to wrap each async call into try catch, we will handle this on global or service level. The same for other methods here

```suggestion
	    return await bcrypt.genSalt(this.saltRounds);
```

liza-veis (changes requested):

apps/backend/src/libs/types/repository.type.ts
s1rserg: This change was suggested [here](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/38#discussion_r1722141884)

s1rserg: I changed this

apps/backend/src/libs/modules/config/base-config.module.ts
```diff
@@ -76,6 +76,14 @@ class BaseConfig implements Config {
 					format: Number,
 				},
 			},
+			ENCRYPTION: {
+				SALT_ROUNDS: {
+					default: 10,
```
what1s1ove, line 81:
```suggestion
					default: null,
```
let's avoid default values

apps/backend/src/libs/modules/encryption/libs/types/encryption.type.ts
```diff
@@ -0,0 +1,11 @@
+type EncryptionResult = {
```
what1s1ove, line 1: each type should be in the separate file

apps/backend/src/libs/modules/encryption/encryption.ts
```diff
@@ -0,0 +1,7 @@
+import { config } from "../config/config.js";
+import { BaseEncryption } from "./base-encryption.module.js";
+
+const encryption = new BaseEncryption(config.ENV.ENCRYPTION.SALT_ROUNDS);
```
what1s1ove, line 4: you need to pass the config to `BaseEncryption ` and get all values there

packages/shared/src/modules/users/libs/types/user-get-all-item-response-dto.type.ts
```diff
@@ -1,6 +1,9 @@
 type UserGetAllItemResponseDto = {
-	email: string;
-	id: number;
+	user: {
```
what1s1ove, line 2: items: YourUserType[]

s1rserg: Can you please explain why there should be items: YourUserType[]? Don't really understand why there should be an array .

apps/backend/src/libs/modules/encryption/libs/types/encryption-result.type.ts
```diff
@@ -0,0 +1,6 @@
+type EncryptionResult = {
```
liza-veis, line 1: usually we name such types as  `<method-name>Result`, don't forget about file name
```suggestion
type EncryptResult = {
```

liza-veis (approved):

packages/shared/src/modules/users/libs/types/user-get-all-item-response-dto.type.ts
liza-veis: @what1s1ove this is an item type, so it should be like this

what1s1ove (approved):

PR #40, opened 2024-08-20
feat: add header component gf-11

TheDaidor:
### **Sumary**

This PR implements the _**header component**_

### **Added**

- Created a header component
- Added logo image
- Created a helper to get the first letter of the user
- Added header to the main page


apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,30 @@
+import { getAvatarInitials } from "~/libs/helpers/helpers.js";
+
+import styles from "./styles.module.css";
+
+const Header = (): JSX.Element => {
+	const name = "Daniel"; // This is developer data to test getAvatarInitials function
```
what1s1ove, line 6: take the user from the state instead

apps/frontend/src/libs/helpers/avatar-helper.ts
```diff
@@ -0,0 +1,11 @@
+const ZERO = 0;
+
+function getAvatarInitials(name: null | string | undefined): string {
```
what1s1ove, line 3: this helper is related to the `users` module so put it there

use an arrow function instead

what1s1ove (changes requested):

apps/frontend/public/assets/images/timer.svg
liza-veis, line 1: name it logo.svg, so it can be easily replaced

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,30 @@
...
+			<div className={styles["logo_container"]}>
+				<img
+					alt=""
+					className={styles["logo_img"]}
+					src="/assets/images/timer.svg"
+				/>
+				<span className={styles["logo_text"]}>Logo</span>
+			</div>
+
+			<div>
```
liza-veis, line 19: Let's create a separate component for Avatar

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -0,0 +1,42 @@
+.header {
+	display: flex;
+	align-items: center;
+	justify-content: space-between;
+	padding: 10px 20px;
+	color: var(--color-text-primary);
+	background-color: var(--color-background-primary);
+}
+
+.logo_container {
```
liza-veis, line 10: we use kebab case for class names
```suggestion
.logo-container {
```

apps/frontend/src/libs/helpers/avatar-helper.ts
liza-veis, line 3: we shouldn't check whether name is exist in this function, and we don't need to check if it's empty. Let's pass the user to the header component from the main page and render header only if user exists. This header will be shown on pages only when the user is present, also we will trim the name during sign-up. So the only action you need to do is to name[0], and I think we don't need a separate helper for this.

liza-veis (changes requested):

apps/frontend/src/libs/components/header/header.tsx
TheDaidor: I've taken the list of users from the store and used the first entry. I'll wait for the getAuthenticatedUser function to be created to be able to directly obtain the user.

apps/frontend/public/assets/images/timer.svg
TheDaidor: done

apps/frontend/src/libs/components/header/header.tsx
TheDaidor: created

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+const Avatar = (): JSX.Element => {
+	const dispatch = useAppDispatch();
+	const users = useAppSelector(({ users }) => users.users);
+	const dataStatus = useAppSelector(({ users }) => users.dataStatus);
+	const isLoading = dataStatus === "pending";
+	const ZERO = 0;
+
+	useEffect(() => {
+		void dispatch(userActions.loadAll());
+	}, [dispatch]);
```
liza-veis, line 20: This is a simple component, we shouldn't have this logic here, pass name as a parameter to Avatar component
```suggestion
```

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+	return (
+		<div className={styles["avatar"]}>
+			{isLoading ? (
+				<Loader />
+			) : (
```
liza-veis, line 26: We don't need to display loader here, this element won't be used in places where there is no user yet
```suggestion
```

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+				<span className={styles["avatar_letter"]}>
+					{users[ZERO]?.email[ZERO]}
```
liza-veis, line 28:
```suggestion
					{name[ARRAY_FIRST_INDEX]}
```

apps/frontend/src/libs/components/avatar/styles.module.css
```diff
@@ -0,0 +1,16 @@
+.avatar {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	width: 40px;
+	height: 40px;
+	color: white;
```
liza-veis, line 7: use common variables for colors

apps/frontend/public/assets/images/logo.svg
liza-veis, line 1: logo should be in frontend/src/assets/images. We use public folder only for content that is not imported into the react app directly

liza-veis (changes requested):

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,22 @@
...
+				<span className={styles["logo-text"]}>Logo</span>
+			</div>
+			<div>
+				<Avatar />
+			</div>
+		</header>
+	);
+};
+
+export { Header };
```
sandrvvu, line 22: maybe import the image source and then use it in your component

`import logoSrc from "/assets/images/logo.svg";`
...
`src={logoSrc}`

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,22 @@
...
+import styles from "./styles.module.css";
+
+const Header = (): JSX.Element => {
+	return (
+		<header className={styles["header"]}>
+			<div className={styles["logo-container"]}>
+				<img
+					alt=""
+					className={styles["logo-img"]}
+					src="/assets/images/logo.svg"
```
sandrvvu, line 11: maybe import the image source and then use it in your component

`import logoSrc from "/assets/images/logo.svg";`
...
`src={logoSrc}`

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,15 @@
+import styles from "./styles.module.css";
+
+interface Properties {
+	name?: string | undefined;
```
sandrvvu, line 4: `name?: string | undefined;` can be simplified to `name?: string;` since undefined is already implied by the optional modifier ?

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+	return (
+		<header className={styles["header"]}>
+			<div className={styles["logo-container"]}>
+				<img alt="GitFit logo" className={styles["logo-img"]} src={logoSrc} />
+				<span className={styles["logo-text"]}>Logo</span>
+			</div>
+			<div>
+				<Avatar name={ARRAY_FIRST_INDEX} />
+			</div>
```
sandrvvu, line 29: if this div doesn’t serve a purpose, you might consider removing it to simplify your code

what1s1ove (commented):
i see a lot of changed that you did not do. Merge the main branch into yours

GvoFor:
Main branch is already merged

liza-veis:
> i see a lot of changed that you did not do. Merge the main branch into yours

This is a github bug that sometimes appears, I've changed and restored base branch, now it's ok

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,28 @@
...
+const Avatar = ({ name }: Properties): JSX.Element => {
+	let firstLetter = null;
+
+	if (name && name.length > FIRST_LETTER_INDEX) {
+		firstLetter = name[FIRST_LETTER_INDEX];
+	}
+
+	const hasFirstLetter = Boolean(firstLetter);
+
+	// TODO: render avatar image instead of first letter (in case if image present)
```
liza-veis, line 18: We can remove this comment, it's not like something isn't working right, we can even not add the profile image feature, and then we will only have this simplified icon.
```suggestion
```

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,28 @@
...
+const FIRST_LETTER_INDEX = 0;
+
+const Avatar = ({ name }: Properties): JSX.Element => {
+	let firstLetter = null;
+
+	if (name && name.length > FIRST_LETTER_INDEX) {
+		firstLetter = name[FIRST_LETTER_INDEX];
+	}
+
+	const hasFirstLetter = Boolean(firstLetter);
```
liza-veis, line 16: Sign up validation will require us to have at least one letter, so we don't need this check here

apps/frontend/src/libs/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,28 @@
+import styles from "./styles.module.css";
+
+type Properties = {
+	name?: string;
```
liza-veis, line 4: We should only render header in a place where we have user
```suggestion
	name: string;
```

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,29 @@
...
+const Header = (): JSX.Element => {
+	const authenticatedUser = useAppSelector(
+		({ auth }) => auth.authenticatedUser,
+	);
+
+	if (!authenticatedUser) {
+		return <></>;
+	}
+
+	const userName = authenticatedUser.user.name;
```
liza-veis, line 16: Hm, authenticatedUser should already be a user with name property, isn't it?

liza-veis (changes requested):

apps/frontend/src/libs/components/header/header.tsx
GvoFor: Auth slice currently looks like this:
image

apps/frontend/src/libs/components/avatar/avatar.tsx
liza-veis: @sandrvvu with the new typescript versions, `name?` and `name: undefined | string` are different types. The first one is for case where you don't pass property at all and `name={undefined}` is not allowed in this case. The second one requires you to pass the property, but it can be undefined. So you can do this `name={undefined}`, but not to omit the name property

apps/frontend/src/libs/components/header/header.tsx
liza-veis: ok, we will change this in a separate ticket

liza-veis (approved):

what1s1ove (approved):

yevhenii-hladkov (approved):

PR #43, opened 2024-08-20
feat: add route protection gf-13

sandrvvu:
Added an authorization plugin to the Fastify server to protect endpoints, injected signed user information from authentication tokens into requests, and excluded sign-up and sign-in routes from protection as white-listed routes.

Some examples:
Screenshot 1: The /sign-up route bypasses authorization and proceeds without restriction.
image

Screenshot 2: A protected route without authorization results in an error.
image

Screenshot 3: A protected route with proper authorization successfully sends data.
image

Screenshot 4: Protected route with Authorization header but invalid token - authorization check fails.
image


apps/backend/src/libs/modules/plugins/auth-plugin.ts
```diff
@@ -0,0 +1,46 @@
...
+	try {
+		const decoded = await token.verifyToken(authToken);
+
+		if (typeof decoded === "object" && "userId" in decoded) {
+			request.user = decoded["userId"] as string;
+		}
+	} catch {
+		return await reply
+			.code(HTTPCode.UNAUTHORIZED)
+			.send({ error: "Invalid token" });
```
yevhenii-hladkov, line 42: Store error messages in enums

packages/shared/src/libs/enums/error-messages.ts
```diff
@@ -0,0 +1,6 @@
+const ErrorMessages = {
```
what1s1ove, line 1: the name should be more specific. ErrorMessages sounds too global 

B23

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -109,6 +110,8 @@ class BaseServerApplication implements ServerApplication {
 			"../../../../public",
 		);
 
+		this.app.addHook("preHandler", authPlugin);
```
what1s1ove, line 113: create a new method `initHooks` and use it in the `init` method

apps/backend/src/libs/modules/plugins/libs/types/custom-request.ts
```diff
@@ -0,0 +1,7 @@
+import { type FastifyRequest } from "fastify";
+
+interface CustomFastifyRequest extends FastifyRequest {
```
what1s1ove, line 3: We need to add a user to the request globally. Add it to the `d.ts` file

what1s1ove (changes requested):

apps/backend/src/libs/modules/plugins/libs/types/custom-request.ts
sandrvvu: Do you mean to do smth like that:
```
import { type FastifyRequest } from "fastify";

type OriginalFastifyRequest = FastifyRequest;

declare module "fastify" {
  interface FastifyRequest extends OriginalFastifyRequest {
    user?: string;
  }
}
```

or just to rename that file?

liza-veis: yes, like this in `backend/index.d.ts`

yevhenii-hladkov (approved):

apps/backend/src/libs/modules/plugins/auth-plugin.ts
```diff
@@ -0,0 +1,46 @@
+import { type FastifyReply, type FastifyRequest } from "fastify";
+
+import { AuthErrorMessages } from "~/libs/enums/enums.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+
+import { token } from "../token/token.js";
+
+const openRoutes: Set<RegExp> = new Set([
```
what1s1ove, line 8: move it to constants

apps/backend/src/index.d.ts
```diff
@@ -0,0 +1,9 @@
+import { type FastifyRequest } from "fastify";
+
+type OriginalFastifyRequest = FastifyRequest;
+
+declare module "fastify" {
+	interface FastifyRequest extends OriginalFastifyRequest {
+		user?: string;
```
what1s1ove, line 7: user should be an object

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -151,6 +156,8 @@ class BaseServerApplication implements ServerApplication {
 	public async init(): Promise<void> {
 		this.logger.info("Application initialization...");
 
+		this.initHooks();
```
what1s1ove, line 159: the order is important. should be after middlewares

packages/shared/src/libs/enums/auth-error-messages.ts
```diff
@@ -0,0 +1,6 @@
+const AuthErrorMessages = {
```
what1s1ove, line 1: B23

packages/shared/src/libs/enums/enums.ts
```diff
@@ -1,4 +1,5 @@
 export { APIPath } from "./api-path.enum.js";
 export { AppEnvironment } from "./app-environment.enum.js";
+export { AuthErrorMessages } from "./auth-error-messages.js";
```
what1s1ove, line 3:
```suggestion
export { AuthErrorMessages } from "./auth-error-messages.enum.js";
```
follow the style of the application

packages/shared/src/libs/enums/auth-error-messages.enum.ts
```diff
@@ -0,0 +1,6 @@
+const AUTHErrorMessages = {
```
what1s1ove, line 1: B23
(should be in a singular form)

AuthErrorMessage

apps/backend/src/libs/modules/plugins/libs/constants/open-routes.ts
```diff
@@ -0,0 +1,6 @@
+const openRoutes: Set<RegExp> = new Set([
```
what1s1ove, line 1: B22

packages/shared/src/libs/enums/auth-error-message.enum.ts
liza-veis, line 1: Let's use common ExceptionMessage enum, we already have it

apps/backend/src/libs/modules/plugins/auth-plugin.ts
```diff
@@ -0,0 +1,42 @@
+import { type FastifyReply, type FastifyRequest } from "fastify";
+
+import { AuthErrorMessage } from "~/libs/enums/enums.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+
+import { token } from "../token/token.js";
+import { OPEN_ROUTES } from "./libs/constants/constants.js";
+
+const authPlugin = async (
```
liza-veis, line 9: Use [Fastify Plugins](https://fastify.dev/docs/latest/Reference/Plugins/) to create a plugin, this is a separate lib. You need to register them then, not to use preHandlers

apps/backend/src/libs/modules/plugins/auth-plugin.ts
liza-veis, line 1: Let's keep it in `src/libs/plugins/authorization/authorization.plugin.ts` and the exported plugin should be named just `authorization`

apps/backend/src/libs/modules/plugins/libs/constants/open-routes.ts
```diff
@@ -0,0 +1,6 @@
+const OPEN_ROUTES: Set<RegExp> = new Set([
```
liza-veis, line 1: Can't we use `as const` and usual array instead of Set?
```suggestion
const OPEN_ROUTES: Set<RegExp> = new Set([
```

liza-veis, line 1: The common name for this is white routes
```suggestion
const WHITE_ROUTES: Set<RegExp> = new Set([
```

liza-veis (changes requested):

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,45 @@
...
+
+			if (!authToken) {
+				return await reply
+					.code(HTTPCode.UNAUTHORIZED)
+					.send({ error: ExceptionMessage.NO_TOKEN_PROVIDED });
+			}
+
+			try {
+				const decoded = await token.verifyToken(authToken);
+				request.user = extractUserFromToken(decoded);
```
what1s1ove, line 33: we won't save the full user payload to the token. we will save only the user ID.

then using this user ID you need to get the full user and add it to the request using userService.getById. Pass this service through plugin parameters

Also, to decode the token, you need to add a new method to the token service

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,45 @@
+import { type FastifyReply, type FastifyRequest } from "fastify";
+import fp from "fastify-plugin";
+
+import { ExceptionMessage } from "~/libs/enums/enums.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+import { token } from "~/libs/modules/token/token.js";
```
what1s1ove, line 6: you need to pass service through parameters to the plugin

apps/backend/src/index.d.ts
```diff
@@ -0,0 +1,11 @@
+import { type FastifyRequest } from "fastify";
+
+import { type UserAuthResponseDto } from "./modules/users/libs/types/types.js";
```
what1s1ove, line 3:
```suggestion
import { type UserAuthResponseDto } from "./modules/users/users.js";
```

apps/backend/src/libs/plugins/libs/constants/constants.ts
perekotypole, line 1: Shouldn't it be called `enums` according to the whole project?

apps/backend/src/libs/plugins/libs/helpers/token-user-extractor.ts
```diff
@@ -0,0 +1,28 @@
+import { type JWTPayload } from "jose";
+
+import { type UserAuthResponseDto } from "../../../../modules/users/libs/types/types.js";
```
perekotypole, line 3: Use absolute imports to take another module
```suggestion
import { type UserAuthResponseDto } from "~/modules/users/libs/types/types.js";
```

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
sandrvvu:
```
type AuthorizationPluginOptions = {
	token: Token;
	userService: UserService;
  }

const authorization = fp<AuthorizationPluginOptions>((fastify, options, done) => {
	const { token, userService } = options;
```

and then add options property to server??
`private options: Options;`
```
	private initPlugins(): void {
		const { token, userService } = this.options;

        this.app.register(authorization, { token, userService });
	}
```

apps/backend/src/index.d.ts
```diff
@@ -0,0 +1,11 @@
+import { type FastifyRequest } from "fastify";
+
+import { type UserAuthResponseDto } from "./modules/users/users.js";
+
+type OriginalFastifyRequest = FastifyRequest;
+
+declare module "fastify" {
+	interface FastifyRequest extends OriginalFastifyRequest {
+		user: null | UserAuthResponseDto;
+	}
```
what1s1ove, line 10:
```suggestion
declare module "fastify" {
	interface FastifyRequest {
		user: null | UserAuthResponseDto;
	}
```
you can declare the interface with the same name and the typescript will merge them automatically

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
what1s1ove: right, add it to the options

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,55 @@
...
+			const authToken = request.headers["authorization"];
+
+			if (!authToken) {
+				return await reply
+					.code(HTTPCode.UNAUTHORIZED)
+					.send({ error: ExceptionMessage.NO_TOKEN_PROVIDED });
+			}
+
+			try {
+				const { userId } = token.decodeToken(authToken);
```
what1s1ove, line 34:
```suggestion
				const { payload } = await token.verify(jwtToken);
				const { userId } = payload;
```
you need to verify it instead, to prevent invalid/expired tokens

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -28,9 +31,15 @@ type Constructor = {
...
 	logger: Logger;
+	options: Options;
 	title: string;
 };
 
+type Options = {
+	token: Token;
+	userService: UserService;
+};
+
```
liza-veis, line 42: Let's pass it this way
```suggestion
	token: Token;
	services: {
	  userService: UserService;
	},
	title: string;
};
```

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -27,6 +27,10 @@ class BaseToken implements Token {
 			.sign(this.secret);
 	}
 
+	public decodeToken(token: string): JWTPayload {
```
liza-veis, line 30: We don't need decode method, we need to use verifyToken

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,67 @@
...
+				for (const whiteRoute of WHITE_ROUTES) {
+					if (whiteRoute.test(url as string)) {
+						return;
+					}
+				}
+
+				const jwtToken = request.headers["authorization"];
+
+				if (!jwtToken) {
+					return await reply
```
liza-veis, line 37: Throw AuthError instead

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,67 @@
...
+				} catch (error) {
+					if (error instanceof UserError) {
+						return await reply
+							.code(HTTPCode.NOT_FOUND)
+							.send({ error: ExceptionMessage.USER_NOT_FOUND });
+					}
+
+					return await reply
+						.code(HTTPCode.UNAUTHORIZED)
+						.send({ error: ExceptionMessage.INVALID_TOKEN });
```
liza-veis, line 58: Throw AuthError, and we don't need to check if user was not found, but we need to add another error message if token expired 
```suggestion
					return await reply
						.code(HTTPCode.UNAUTHORIZED)
						.send({ error: ExceptionMessage.INVALID_TOKEN });
```

apps/backend/src/libs/plugins/libs/enums/white-routes.ts
```diff
@@ -0,0 +1,6 @@
+const WHITE_ROUTES = [
```
liza-veis, line 1: should be in constatns, not enums

packages/shared/src/libs/enums/exception-message.enum.ts
```diff
@@ -1,6 +1,8 @@
 const ExceptionMessage = {
 	EMAIL_USED: "Email address is already in use.",
 	INVALID_CREDENTIALS: "Invalid credentials.",
+	INVALID_TOKEN: "Invalid token",
```
liza-veis, line 4: End messages with a period

liza-veis (changes requested):

what1s1ove (commented):
Waiting for fixes provided by Liza

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -28,9 +31,17 @@ type Constructor = {
 	config: Config;
 	database: Database;
 	logger: Logger;
+	options: Options;
```
liza-veis, line 34: I meant that we don't need to pass them as separate options, let's pass them to the first level as other properties
```suggestion
	services: {
		userService: UserService;
	};
	token: Token;
```

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,69 @@
...
+import { ExceptionMessage } from "~/libs/enums/enums.js";
+import { HTTPCode } from "~/libs/modules/http/http.js";
+import { type Token } from "~/libs/modules/token/token.js";
+import { AuthError } from "~/modules/auth/auth.js";
+import { type UserService } from "~/modules/users/user.service.js";
+
+import { WHITE_ROUTES } from "../libs/constants/constants.js";
+import { isTokenExpiredError } from "../libs/helpers/helpers.js";
+
+type AuthorizationPluginOptions = {
```
liza-veis, line 13:
```suggestion
type Options = {
```

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,69 @@
...
+const authorization = fp<AuthorizationPluginOptions>(
+	(fastify, options, done) => {
+		const { token, userService } = options;
+
+		fastify.decorateRequest("user", null);
+
+		fastify.addHook("preHandler", async (request: FastifyRequest) => {
+			const { url } = request.raw;
+
+			for (const whiteRoute of WHITE_ROUTES) {
```
liza-veis, line 27: whiteRoutes should be passed as a parameter from the server-application. And we also should keep WHITE_ROUTES constant in `/src/libs/modules/server-application/libs/constants/white-routes.constant.ts`

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,69 @@
...
+
+			try {
+				const payload = await token.verifyToken(jwtToken);
+				const { userId } = payload;
+
+				const user = await userService.find(userId as number);
+
+				request.user = user;
+			} catch (error) {
+				if (error instanceof Error && isTokenExpiredError(error)) {
```
liza-veis, line 50: I think we can check it like this. But double-check that this works
```suggestion
                const isTokenExpiredError = error instanceof JWTExpired;
               
				if (isTokenExpiredError) {
```

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,69 @@
...
+					message: ExceptionMessage.NO_TOKEN_PROVIDED,
+					status: HTTPCode.UNAUTHORIZED,
+				});
+			}
+
+			try {
+				const payload = await token.verifyToken(jwtToken);
+				const { userId } = payload;
+
+				const user = await userService.find(userId as number);
```
liza-veis, line 46: we don't need to make the request if there is no user_id

liza-veis (changes requested):

what1s1ove (approved):

apps/backend/src/libs/modules/server-application/libs/constants/white-routes.ts
liza-veis, line 1: file name should be `white-routes.constant.ts`

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,69 @@
...
+				message: ExceptionMessage.NO_TOKEN_PROVIDED,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
+
+		try {
+			const payload = await token.verifyToken(jwtToken);
+			const { userId } = payload;
+
+			if (userId) {
```
liza-veis, line 44: We need to throw error if there is no userId

packages/shared/src/libs/enums/exception-message.enum.ts
```diff
@@ -1,6 +1,9 @@
...
```
liza-veis, line 4:
```suggestion
	INVALID_TOKEN: "Invalid token.",
```

liza-veis (changes requested):

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
sandrvvu: if there is no userId in payload, the result is already "Invalid token." error.
should it throw smth else?

liza-veis: Can't there be a case when the token is valid but contains any other info like project_id? if we have different tokens in the app for example? Or actually I'm thinking it still won't find user in this case 🤔 then we don't need this if condition to check if userId is present

liza-veis (approved):

PR #49, opened 2024-08-21
feat: add authenticated user flow gf-26

perekotypole:
- save the authorization token to the localStorage _(implemented for sign-up)_
- after refreshing the page, the authenticated user data should be loaded and saved in the app state _(in App component)_


apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -18,11 +18,24 @@ const initialState: State = {
...
+			state.dataStatus = DataStatus.REJECTED;
+		});
+
 		builder.addCase(signUp.pending, (state) => {
 			state.dataStatus = DataStatus.PENDING;
 		});
-		builder.addCase(signUp.fulfilled, (state) => {
+		builder.addCase(signUp.fulfilled, (state, action) => {
 			state.dataStatus = DataStatus.FULFILLED;
+			state.authenticatedUser = action.payload;
```
yevhenii-hladkov, line 38: `builder.addMatcher(
            isAnyOf(
               signUp.fulfilled,
               getAuthenticatedUser.fulfilled
            ),
            (state, action) => {
                state.dataStatus = DataStatus.FULFILLED;
			          state.authenticatedUser = action.payload;
            },
        )`

what1s1ove (approved):
lgtm

yevhenii-hladkov (approved):

apps/frontend/src/modules/auth/auth-api.ts
```diff
@@ -20,6 +21,18 @@ class AuthApi extends BaseHTTPApi {
...
+	public async getAuthenticatedUser(): Promise<null | UserAuthResponseDto> {
+		const response = await this.load(
+			this.getFullEndpoint(AuthApiPath.AUTHENTICATED_USER, {}),
+			{
+				hasAuth: true,
+				method: "GET",
+			},
+		);
+
+		return await response.json<null | UserAuthResponseDto>();
```
liza-veis, line 33: it will return either user or error
```suggestion
		return await response.json<UserAuthResponseDto>();
```

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -18,15 +18,29 @@ const initialState: State = {
...
 		});
+
+		builder.addMatcher(
+			isAnyOf(getAuthenticatedUser.pending, signUp.pending),
+			(state) => {
+				state.dataStatus = DataStatus.PENDING;
+			},
+		);
+
+		builder.addMatcher(
```
liza-veis, line 37: Let's avoid using matchers here and define these cases separately, this is more readable and convenient to work in case we need to change something (for example, use separate status properties for each request). Matchers are good, but only for really common logic where you don't need to consider what each of the individual actions do.

packages/shared/src/modules/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -1,4 +1,5 @@
 const AuthApiPath = {
+	AUTHENTICATED_USER: "/authenticated-user",
 	ROOT: "/",
```
liza-veis, line 3: Let's remove Root from here, I think it's not needed
```suggestion
```

liza-veis (changes requested):

apps/frontend/src/modules/auth/slices/auth.slice.ts
liza-veis: yes, do the same for fullfilled too please

apps/frontend/src/index.tsx
```diff
@@ -20,27 +21,34 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
...
-						element: (
-							<ProtectedRoute>
-								<App />
-							</ProtectedRoute>
-						),
+						children: [
+							{
+								element: (
+									<ProtectedRoute>
+										<DefaultPage />
```
liza-veis, line 28: Let's name it Projects
```suggestion
										<Projects />
```

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -18,16 +18,33 @@ const initialState: State = {
...
-			state.authenticatedUser = action.payload;
+		});
+		builder.addCase(signUp.pending, (state) => {
+			state.dataStatus = DataStatus.PENDING;
 		});
 		builder.addCase(signUp.rejected, (state) => {
 			state.dataStatus = DataStatus.REJECTED;
 		});
+
+		builder.addMatcher(
```
liza-veis, line 42: Add separate cases instead of matcher

liza-veis (changes requested):

apps/frontend/src/modules/auth/slices/auth.slice.ts
perekotypole: > Add separate cases instead of matcher

but what to do if @yevhenii-hladkov has already asked to add matches? https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/49#discussion_r1725067100

liza-veis: oh, didn't saw it, ok then let's leave it

liza-veis (approved):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -1,19 +1,24 @@
...
 
 const ProtectedRoute = ({ children }: Properties): JSX.Element => {
-	const authenticatedUser = useAppSelector(
-		(state) => state.auth.authenticatedUser,
-	);
+	const { authenticatedUser, dataStatus } = useAppSelector(({ auth }) => auth);
 
 	const hasAuthenticatedUser = Boolean(authenticatedUser);
+	const isLoading =
+		dataStatus === DataStatus.PENDING || dataStatus === DataStatus.IDLE;
```
yevhenii-hladkov, line 17:
```suggestion
	const isLoading = [DataStatus.PENDING, DataStatus.IDLE].includes(dataStatus);
```

what1s1ove (approved):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
perekotypole: There is an error `Argument of type 'ValueOf<{ readonly FULFILLED: "fulfilled"; readonly IDLE: "idle"; readonly PENDING: "pending"; readonly REJECTED: "rejected"; }>' is not assignable to parameter of type '"idle" | "pending"'.` which I don't know how to solve

yevhenii-hladkov: Okay then leave it as is )

liza-veis (approved):

PR #50, opened 2024-08-21
feat: create a select component gf-24

Vitaliistf:
Implemented custom select component. It supports multi-select and single-select. Basically its just a wrapper on react-select with custom styling. Also created generic Option type, used to put options to the select. Here what it looks like:
Screenshot 2024-08-21 at 14 40 01
Screenshot 2024-08-21 at 14 40 18
Screenshot 2024-08-21 at 16 11 35


apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -1,5 +1,9 @@
+import { useState } from "react";
+import { type MultiValue, type SingleValue } from "react-select";
+
 import reactLogo from "~/assets/images/react.svg";
 import {
+	CustomSelect,
```
what1s1ove, line 6:
```suggestion
	Select,
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -1,5 +1,9 @@
+import { useState } from "react";
+import { type MultiValue, type SingleValue } from "react-select";
```
what1s1ove, line 2:
```suggestion
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -1,5 +1,9 @@
+import { useState } from "react";
```
what1s1ove, line 1: see how we import hooks in other files

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -19,7 +23,29 @@ import {
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { type Option } from "../custom-select/libs/types/option.type.js";
+
 const App = (): JSX.Element => {
+	// Following code is for CustomSelect testing
+	const options = [
```
what1s1ove, line 30: pass an empty array for now

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -19,7 +23,29 @@ import {
...
+		{ label: "1", value: "1" },
+		{ label: "2", value: "2" },
+		{ label: "3", value: "3" },
+		{ label: "4", value: "4" },
+		{ label: "5", value: "5" },
+	];
+	const [singleValue, setSingleValue] = useState<Option>({
```
what1s1ove, line 37: you need to use `useForm` for working will all form elements

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -71,6 +97,12 @@ const App = (): JSX.Element => {
 					<Table<Person> columns={mockTableColumns} data={mockTableData} />
 				</>
 			)}
+			<CustomSelect
+				// eslint-disable-next-line react/jsx-no-bind
```
what1s1ove, line 101: you cannot disable linter just because

apps/frontend/src/libs/components/custom-select/custom-select.tsx
```diff
@@ -0,0 +1,80 @@
...
+		newValue: MultiValue<Option> | SingleValue<Option>,
+		actionMeta: ActionMeta<Option>,
+	) => void;
+	options: Option[];
+	placeholder?: string;
+	title?: string;
+	value: PropsValue<Option>;
+};
+
+const CustomSelect = ({
```
what1s1ove, line 23:
```suggestion
const Select = ({
```

apps/frontend/src/libs/components/custom-select/custom-select.tsx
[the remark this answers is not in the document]
what1s1ove: agreed. rename it to select

apps/frontend/src/libs/components/custom-select/custom-select.tsx
```diff
@@ -0,0 +1,80 @@
...
+	onChange,
+	options,
+	placeholder,
+	title,
+	value,
+}: Properties): JSX.Element => {
+	const getClassName = (
+		baseClass: string,
+		conditionClass?: string,
+		condition = false,
```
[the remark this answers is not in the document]
GvoFor: Look at theese conversetions:
- https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/37#discussion_r1723148841
- https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/37#discussion_r1724736770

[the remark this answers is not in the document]
Vitaliistf: I'll take a look on this, thanks

apps/frontend/src/libs/components/app/app.tsx
Vitaliistf: I need to use these types for select change function

Vitaliistf: Should I omit this somehow, or export types somewhere in `/libs` folder? Or is this suggestion outdated?

what1s1ove: no, you don't. only `select.tsx` should know about that we user react-select in our application

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -13,13 +16,40 @@ import {
...
+	// Following code is for Select testing
+	const options: Option<number>[] = [];
+
+	const { control } = useAppForm({
+		defaultValues: {
+			singleSelect: null,
+		},
+	});
+
+	const { field } = useFormController({
```
what1s1ove, line 40: this logic should be inside the select component

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -13,13 +16,40 @@ import {
...
 	useEffect,
+	useFormController,
 	useLocation,
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { type Option } from "../select/libs/types/option.type.js";
+
 const App = (): JSX.Element => {
+	// Following code is for Select testing
```
what1s1ove, line 31: remove comments, they are useless

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -13,13 +16,40 @@ import {
...
+	useAppForm,
 	useAppSelector,
+	useCallback,
 	useEffect,
+	useFormController,
 	useLocation,
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { type Option } from "../select/libs/types/option.type.js";
```
what1s1ove, line 28:
```suggestion
import { type Option } from "~/libs/components/select/select.js";
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -13,13 +16,40 @@ import {
...
+
 const App = (): JSX.Element => {
+	// Following code is for Select testing
+	const options: Option<number>[] = [];
```
what1s1ove, line 32:
```suggestion
```
pass en empty array directly in the props

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,81 @@
...
+		newValue: MultiValue<Option<T>> | SingleValue<Option<T>>,
+		actionMeta: ActionMeta<Option<T>>,
+	) => void;
+	options: Option<T>[];
+	placeholder?: string;
+	title?: string;
+	value: PropsValue<Option<T>>;
+};
+
+const Select = <T = string,>({
```
what1s1ove, line 24: you need to wrap it with react-form the same as our Input component.

Vitaliistf: Oh, understood, thanks.

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,81 @@
...
+	title,
+}: Properties<T>): JSX.Element => {
+	const { field } = useFormController({
+		control,
+		name,
+	});
+
+	return (
+		<div className={styles["container"]}>
+			{title && <span className={styles["title"]}>{title}</span>}
```
liza-veis, line 40: Label is required for selects, in case we need to hide it, we will still have it, but with visually-hidden class. But for now it's not needed
```suggestion
			<span className={styles["label"]}>{title}</span>
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,81 @@
...
+			<ReactSelect
+				classNames={{
+					control: (state) =>
+						getValidClassNames(
+							styles["control"],
+							state.isFocused && styles["control--focused"],
+						),
+					indicatorsContainer: () =>
+						getValidClassNames(styles["indicators-container"]),
```
liza-veis, line 49: We don't need to use getValidClassNames for one className, it's only for cases when we need to combine multiple classNames

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,81 @@
...
+	placeholder?: string;
+	title?: string;
+};
+
+const Select = <T extends FieldValues>({
+	control,
+	isMulti = false,
+	name,
+	options,
+	placeholder,
```
liza-veis, line 30: We won't need placeholder for selects, labels will be enough for now
```suggestion
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { getValidClassNames } from "~/libs/helpers/helpers.js";
+import { useFormController } from "~/libs/hooks/hooks.js";
+
+import styles from "./styles.module.css";
+
+type Properties<T extends FieldValues> = {
+	control: Control<T, null>;
+	isMulti?: boolean;
+	name: FieldPath<T>;
+	options: OptionsOrGroups<unknown, GroupBase<unknown>> | undefined;
```
liza-veis, line 20: Let's use our custom option type

apps/frontend/src/libs/components/select/styles.module.css
```diff
@@ -0,0 +1,96 @@
+.container {
+	display: flex;
+	flex-direction: column;
+	gap: 8px;
+}
+
+.title {
+	font-size: 16px;
```
liza-veis, line 8: what about line-height?

apps/frontend/src/libs/components/select/styles.module.css
```diff
@@ -0,0 +1,96 @@
...
+
+.option {
+	width: 100%;
+	padding: 8px 12px;
+	font-size: inherit;
+	color: var(--color-text-primary);
+	border-radius: 4px;
+}
+
+.control--focused {
```
liza-veis, line 75: We don't use bem here because of css modules. Just use kebab-case
```suggestion
.control-focused {
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -13,13 +14,26 @@ import {
...
+import { type Option } from "../select/select.js";
+
 const App = (): JSX.Element => {
+	type SelectData = {
+		options: Option<number>[];
+	};
+
+	const { control } = useAppForm<SelectData>({
+		defaultValues: {
+			options: [],
```
liza-veis, line 33: let's pass some options to test
```suggestion
			options: ['Option 1', 'Option 2'],
```

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -13,13 +14,26 @@ import {
...
 	useAppSelector,
 	useEffect,
 	useLocation,
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { type Option } from "../select/select.js";
+
 const App = (): JSX.Element => {
+	type SelectData = {
```
liza-veis, line 27: move type outside of component

apps/frontend/src/libs/components/select/select.tsx
Vitaliistf: If I do not use getValidClassNames then this I get this error 
Screenshot 2024-08-22 at 20 49 00
You're right, it's okay, when we style basic components, but this does not work with react-select. Previously I used `styles["..."] ?? ""`, and it looked quite messy, then I ended up with creating a function for this, which also combined classes on conditions, and then I changed it with a new one from other branch (getValidClassNames). Maybe there is another approach, I can use?

apps/frontend/src/libs/components/app/app.tsx
Vitaliistf: Vladyslav asked me to remove them in his previous review : ) but sure, I'll do it.

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -13,13 +14,31 @@ import {
...
 	useEffect,
 	useLocation,
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { type Option } from "../select/select.js";
+
+type SelectData = {
+	options: Option<number>[];
+};
```
GvoFor, line 28: I don't think that creating such a type everywhere we use `Select` is good idea. What is a purpose? Can't it be inferred inside component itself?

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -71,6 +90,13 @@ const App = (): JSX.Element => {
 					<Table<Person> columns={mockTableColumns} data={mockTableData} />
 				</>
 			)}
+			<Select<SelectData, number>
+				control={control}
+				isMulti
+				name="options"
+				options={options}
```
what1s1ove, line 97:
```suggestion
				options={[]}
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,81 @@
...
+	title?: string;
+};
+
+const Select = <TTFieldValues extends FieldValues, TValue>({
+	control,
+	isMulti = false,
+	name,
+	options,
+	title,
+}: Properties<TTFieldValues, TValue>): JSX.Element => {
```
GvoFor, line 33: In case you really need this first generic (`TTFieldValue`) it's better to rename it to `TFiedValues` (rid of extra T).
But I'd like to see that this type is not passed throuh props, but instead inferred (generated) inside (cause you have here `TValue` and `isMulti`, so I think it must be possible to do

GvoFor: P.s. I'm saying about `const Select <TTFiedValues ..., TValue>(` this first generic

Vitaliistf: Oh, I accidentally typed the second T, thanks.

apps/frontend/src/libs/components/app/app.tsx
Vitaliistf: Well, unfortunately it might be the best I can do. Controlled components are a bit confusing to me, and all of the typescript examples are made like this. also, the name that we pass to Select itself, should be a name of a field in the type that we pass as first generic (SelectData here). Currently I'm trying to do something about it. If you have more experience with controlled components, I'd be grateful to hear how this can be improved.

apps/frontend/src/libs/components/app/app.tsx
Vitaliistf: Liza asked me to leave some data to test, but okay.

apps/frontend/src/libs/components/app/app.tsx
GvoFor: I have no experience with controlled components to be honest, but I'd like to help. So I will clone your branch later and take some time trying to figure it out

apps/frontend/src/libs/components/select/select.tsx
liza-veis: let's use `styles["input"] as string`. It's ok to cast types when you are 100% it is present

Vitaliistf: Okay, thanks

Vitaliistf: Fixed

apps/frontend/src/libs/components/app/app.tsx
GvoFor: Not sure why, but I can't pull your last changes. Only til this commit:
image
Maybe the reason is that your branch after that commit have conflicts ? Can you merge main branch into yours and resolve the conflicts, and then I'll try again

apps/frontend/src/libs/components/select/styles.module.css
```diff
@@ -0,0 +1,104 @@
+.label {
+	display: flex;
+	flex-direction: column;
+	gap: 8px;
+}
+
+.label-text {
+	font-size: 16px;
+	font-weight: 500;
+	line-height: 19.2px;
```
liza-veis, line 10: line height should have relative value as in figma

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { type Option } from "./select.js";
+import styles from "./styles.module.css";
+
+type Properties<TFieldValues extends FieldValues, TValue> = {
+	control: Control<TFieldValues, null>;
+	isMulti?: boolean;
+	name: FieldPath<TFieldValues>;
+	options:
+		| OptionsOrGroups<Option<TValue>, GroupBase<Option<TValue>>>
+		| undefined;
```
liza-veis, line 23: This property can't be undefined, and we will only pass an array of options, we don't need groups for now
```suggestion
	options: Option<TValue>[];
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,78 @@
...
+	title?: string;
```
liza-veis, line 24: rename it to label and it should be required
```suggestion
	label: string;
```

liza-veis (changes requested):

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,78 @@
...
+} from "react-hook-form";
+import ReactSelect, {
+	type GroupBase,
+	type OptionsOrGroups,
+} from "react-select";
+
+import { getValidClassNames } from "~/libs/helpers/helpers.js";
+import { useFormController } from "~/libs/hooks/hooks.js";
+
+import { type Option } from "./select.js";
```
GvoFor, line 14: This looks not great. It's beyyer to import from "./libs/types/types.ts"

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,78 @@
...
+				options={options}
+				unstyled
+				value={field.value}
+			/>
+		</label>
+	);
+};
+
+export { Select };
+export { type Option } from "./libs/types/types.js";
```
GvoFor, line 78: You will import `type Option` above, and here export what was imported

apps/frontend/src/libs/components/app/app.tsx
```diff
@@ -13,13 +14,26 @@ import {
...
+	useAppForm,
 	useAppSelector,
 	useEffect,
 	useLocation,
 } from "~/libs/hooks/hooks.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { type Option } from "../select/select.js";
+
+type SelectData = {
```
what1s1ove, line 26: you do not need this type, remove it. write it directly in generic

GvoFor:
Oh, It's just came to my mind, that it's not always the case, that `Select` will be used as form element. For example, for `TablePagination` I'd like to you `Select` without any form (don't use `useForm` for `TablePagination`), i.e. it'll be more convinient for me to pass `options` and `onChange` props, and not `control`.

So how will I use this component for `TablePagination`? should I use `useForm` hook in `TablePagination` component? or `Select` component should be adjusted somehow?

@what1s1ove @liza-veis

apps/frontend/src/libs/components/select/select.tsx
liza-veis: agree, you don't need from here

liza-veis:
> Oh, It's just came to my mind, that it's not always the case, that Select will be used as form element. For example, for TablePagination I'd like to you Select without any form (don't use useForm for TablePagination), i.e. it'll be more convinient for me to pass options and onChange props, and not control.
>
> So how will I use this component for TablePagination? should I use useForm hook in TablePagination component? or Select component should be adjusted somehow?

Yes, we will need to use useForm for this in TablePagination components. All form controls should be inside a form.

apps/frontend/src/libs/components/select/libs/types/option.type.ts
```diff
@@ -0,0 +1,6 @@
+type Option<T = string> = {
```
liza-veis, line 1: This type will be used as a common type in different places in the application, so let's move it to `apps/frontend/src/libs/types/select-option.type.ts` and rename to SelectOption

liza-veis (changes requested):

apps/frontend/src/libs/components/select/select.tsx
Vitaliistf: > agree, you don't need from here

Linter would not allow me to do that, but anyway, I don't need that line at all now.

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -0,0 +1,74 @@
...
+	options,
+}: Properties<TFieldValues, TOptionValue>): JSX.Element => {
+	const { field } = useFormController({
+		control,
+		name,
+	});
+
+	return (
+		<label className={styles["label"]}>
+			<span className={styles["label-text"]}>{label}</span>
```
GvoFor, line 38: Not sure, if this [comment](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/50#discussion_r1727521295) was fully done. For example, I will need to hide label for select in pagination component, but for now your code does not provide the ability to hide the label.
So I suggest to add additional prop `isLabelVissible`, which should be `true` by default, and depending on value of this prop, you will choose class for this element: `className={isLabelVissible? styles["label-text"] : "visually-hidden"}`

Vitaliistf: "But for now it's not needed"

GvoFor: Okay, it makes sense. By the way, I'll be able to add this "feature", in my PR, I guess

what1s1ove (approved):

yevhenii-hladkov (approved):

liza-veis (approved):

PR #42, opened 2024-08-20
feat: create modal component gf-21

Fjortis:
This merge request implements a reusable modal component as per the requirements specified in the task gf-21. The modal component includes the following features:

Title and Content:

The modal accepts a title as a prop and uses children to populate the body of the modal.
Close Functionality:

The modal can be closed by clicking on the close ("×") icon.
The modal can also be closed by clicking outside the modal content area.
Accessibility:

The modal overlay is accessible via keyboard interactions with proper role and tabindex attributes.
Testing:
Added a button to the main page to open the modal for testing purposes.
Verified that the modal closes correctly when clicking the close icon, clicking outside the modal.
Ensured that the component is accessible and keyboard-navigable.

modal1 5

PR #51, opened 2024-08-21
feat: create table pagination component gf-23

GvoFor:
This PR adds `TablePagination` component as well as `useTablePagination` hook for mannaging state of component.

It looks like this:
screenshot

Before merge background color will be set to transparent

This PR also modifies `Icon`, `IconButton` and `Select` components adding useful and needed features to them


GvoFor:
In case we'll have `IconButton` component according to the [comment](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/42#discussion_r1725496624), I will replace my `ChangePageButton` component with it, as soon as `IconButton` component will be merged

apps/frontend/src/libs/hooks/use-table-pagination/use-table-pagination.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+
+type ReturnType = {
+	changeRowsPerPage: (value: number) => void;
+	openFirstPage: () => void;
+	openLastPage: () => void;
+	openNextPage: () => void;
+	openPreviousPage: () => void;
+	page: number;
+	rowsPerPage: number;
```
liza-veis, line 21:
```suggestion
	onPerPageChange: (perPage: number) => void;
	onPageChange: (page: number) => void;
	page: number;
	perPage: number;
```

apps/frontend/src/libs/hooks/use-table-pagination/use-table-pagination.ts
```diff
@@ -0,0 +1,84 @@
...
+
+	useEffect(() => {
+		if (page > totalPages) {
+			setPage(totalPages);
+		}
+
+		if (page < FIRST_PAGE) {
+			setPage(FIRST_PAGE);
+		}
+	}, [totalPages]);
```
liza-veis, line 48: we should prevent from setting invalid value to page, not change it after it was set
```suggestion
```

liza-veis (changes requested):

GvoFor:
I have issues with using of `IconButton`. Before it was introduced, I make my own `ChangePageButton` component (local to `TablePagination`), which looks like this:
image
there instead of rendering string I thinking of rendering `Icon` component passing `icon` as `iconName` to it.
And the styles looks like this:
image
But when `IconButton` was introduced, I thought I can replace my `ChangePageButton` with it (or at least use `ChangePageButton` as a wrapper for `IconButton`, to reduce props amount). The problem is that I can't style `IconButton` as I need, adding a border, size, disabled style and maybe some hover effect (do we need it?).

So what should I do? Can I redesign `IconButton` a bit?
And also have a question: do we really need to pass dimensions and color as a props? Isn't it better to set size and color using css?

apps/frontend/src/libs/hooks/use-table-pagination/use-table-pagination.ts
```diff
@@ -0,0 +1,84 @@
...
+	useCallback,
+	useEffect,
+	useMemo,
+	useState,
+} from "~/libs/hooks/hooks.js";
+
+type Parameters = {
+	startingPage: number;
+	startingRowsPerPage: number;
+	totalItems: number;
```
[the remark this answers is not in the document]
GvoFor: Maybe it's better to call `perPage` as `pageSize` everywhere?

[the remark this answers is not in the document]
GvoFor: And what if query does not contain mentioned params or params are not integers? Should I define some default values somewhere and use them?

apps/frontend/src/libs/hooks/use-table-pagination/use-table-pagination.ts
liza-veis: I think if we control page in this component, this shouldn't be needed now

liza-veis:
> So what should I do? Can I redesign IconButton a bit?

yes, you will need to update it a bit. On design, all icon buttons look the same. By default, they are transparent and when hovered they should have a slight background as on the top button. For pagination, buttons are the same, but with a border. 

Screenshot 2024-08-22 at 18 30 56

So to implement this, let's adjust an IconButton to have the default styles as in the image above (add hover effect too). And then add a variant property to it with two options: default and outlined. Outlined will be the variant with the border that you need.

> And also have a question: do we really need to pass dimensions and color as a props? Isn't it better to set size and color using css?

If your question is whether it's better to pass width/height as an html attribute or css style: for images and svg this could be beneficial to pass width and height so that the browser knows how much space to allocate for an image before styles are loaded. But the color should be passed to css in Icon component.

If you are asking why not pass properties like className instead of separate properties: className is too customizable, you can override a lot of things with it. When you will have a lot of components in your projects that override the default one as they need it's like not having a common component at all. Whenever you change something in the common component, you will need to check all the buttons in your project whether they didn't break because of their custom styles. You never know which variants your button can have without checking all the buttons in the project. Whereas when you pass only specific values to the common button component, you control all variants only from this common component and you can check this in any place in your project.

Also, I'm thinking that we probably don't need to pass width and height for IconButton. We should have those properties for Icon, but IconButtons usually will have the same size as usual buttons do. So let's remove width and height from IconButton component (until we need them), and pass default width and height for icons in IconButton.

liza-veis:
> Where it is better to use useSearchQuery hook: inside my usePagination hook, or inside the component, there TablePagination component will be used (as I did for now in the example of usage inside App component)?
>
>It seems to me, that it's better to use it inside my hook, so developer which will use my hook doesn't take care about search queries. (And also, each time we'll use usePagination hook, we'll need it to be synchronized with search queries, I think, so the code responsible for this will duplicate, if I'm not put useSearchQuery hook inside my hook)

Both options can be used depending on project. usePagination can either control pages only or fully control them with query params. I would prefer the second option to avoid a lot of repetitive logic with query params because we want to keep it the same for all places.

GvoFor:
Here you can see, how `IconButton` component looks like
image
1 and 2 have `primary` (default) variant
3 and 4 have `outlined` variant
1 is hovered
4 is disabled
Hover effect applies only on non-disabled buttons

apps/frontend/src/libs/maps/color-to-css.map.ts
```diff
@@ -0,0 +1,19 @@
+import { type Color } from "~/libs/types/types.js";
```
liza-veis, line 1: Instead of this, let's go with another approach. Let's add styles file to Icon component and set it's fill or color to currentColor. Then add need color to Icon wrapper

liza-veis (changes requested):

apps/frontend/src/libs/components/table-pagination/table-pagination.tsx
```diff
@@ -0,0 +1,133 @@
...
+	pageSize,
+	totalItems,
+}: Properties): JSX.Element => {
+	const totalPages = useMemo(
+		() => Math.ceil(totalItems / pageSize),
+		[totalItems, pageSize],
+	);
+
+	const hasNextPage = useMemo(() => page < totalPages, [page, totalPages]);
+	const hasPreviousPage = useMemo(() => page > FIRST_PAGE, [page]);
```
liza-veis, line 66: I think we don't need use memo here, too small check to allocate additional memory for this

apps/frontend/src/libs/hooks/use-pagination/use-pagination.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const usePagination = ({ totalItems }: Parameters): ReturnType => {
+	const [searchParameters, setSearchParameters] = useSearchParams();
+
+	// There is no check for page and pageSize to be in available range
+	const pageQueryParameter =
+		Number(searchParameters.get(QueryParameter.page)) || DEFAULT_VALUES.PAGE;
+	const pageSizeQueryParameter =
+		Number(searchParameters.get(QueryParameter.pageSize)) ||
+		DEFAULT_VALUES.PAGE_SIZE;
```
liza-veis, line 31: Set these properties as initial values in useEffect, and here you will be able to add a check whether the page is in range

apps/frontend/src/libs/components/table-pagination/table-pagination.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const TablePagination = ({
+	onPageChange,
+	onPageSizeChange,
+	page,
+	pageSize,
+	totalItems,
+}: Properties): JSX.Element => {
+	const totalPages = useMemo(
+		() => Math.ceil(totalItems / pageSize),
```
liza-veis, line 61: Can't you use calculateTotalPages here?

liza-veis (changes requested):

apps/frontend/src/libs/components/table-pagination/libs/components/change-page-button/change-page-button.tsx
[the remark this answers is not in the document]
GvoFor: Okay. I add it in order to avoid code repetition, such as setting `iconColor` and `variant`. But now I see, that this was not the best idea ;)

apps/frontend/src/libs/components/table-pagination/table-pagination.tsx
GvoFor: So you suggest to replace `useMemo` with `useCallback`? If yes, what benefits this bring? If no, one of the efficient 
solution may be to move this functions out of the component (with aproppriate parameters) and then use them:
```ts
const hasNextPage = (page: number, totalPages: number): boolean => page < totalPages;
const hasPreviousPage = ...;

const TablePagination = (...) => {
  ...
  isDisabled={!hasNextPage(page, totalPages)}
  ...
}
```

apps/frontend/src/libs/maps/color-to-css.map.ts
GvoFor: If I correctly understand, `IconButton` is an expample of such Icon wrapper, and I should set icon color inside `IconButton` styles file instead of passing it as a prop to `Icon` component. Right?

But then the question is how to control the color of `IconButton`'s icon? Sometimes it should have `text-primary` and sometimes `text-secondary`. 

I check figma design again, and come up with conclusion: all IconButtons without border have `text-secondary` color, and only with border have `text-primary`. So does it make sense to stick to this observation and for `primary` button variant use `text-secondary` color, and for `outlined` use `text-primary`? or there must be better solution?

apps/frontend/src/libs/hooks/use-pagination/use-pagination.ts
GvoFor: Not sure, what do you mean (

apps/frontend/src/libs/components/table-pagination/table-pagination.tsx
liza-veis: I suggest to write it without any hooks (if there is no restriction in linter) `const hasNextPage = page < totalPages`

apps/frontend/src/libs/maps/color-to-css.map.ts
liza-veis: > So does it make sense to stick to this observation and for primary button variant use text-secondary color, and for outlined use text-primary? or there must be better solution?

yes, let's stick to it. In case we need more colors we will add a color property, but it only will accept a predefined limited set of colors (not css variables) and then we will match them to appropriate styles. Just as example of what I mean:

```typescript
type Properties = {
  color: 'primary' | 'grey',
}

const classNames = getValidClassNames(
  color === "primary" && styles["color-primary"],
  color === "grey" && styles["color-gray"],
)
```

But for now implement it in simpler way as you described above (without color property)

apps/frontend/src/libs/hooks/use-pagination/use-pagination.ts
liza-veis: The point is that we should return only valid values from this hook. Here is an example of what I mean:

```typescript
// remove default values and parsing to number here
const pageQueryParameter = searchParameters.get(QueryParameter.page);
const pageSizeQueryParameter = searchParameters.get(QueryParameter.pageSize);

const [page, setPage] = useState(DEFAULT_PAGE);
const [pageSize, setPageSize] = useState(DEFAULT_PAGE_SIZE);

useEffect(() => {
 // parse query params values (turn into numbers)
 // check that values are in range, if not override query params with default values
 // set page and pageSize with valid values

}, [pageQueryParameter, pageSizeQueryParameter])

// return only valid page and pageSize from hook
```

apps/frontend/src/libs/components/table-pagination/table-pagination.tsx
GvoFor: Oh, but without useMemo hook our component won't be rerendered whenever `page` or `totalPages` change, if I'm not wrong

GvoFor: Leave it as it is, because it does not work without useMemo hook

apps/frontend/src/libs/hooks/use-pagination/use-pagination.ts
GvoFor: I tried to do this and as I was afraid it cause render problem, cause this useEffect changes page/pageSize when queryParams updates, and other useEffect listens for page/pageSize updates and change queryParams. So here is a cyclic problem. But I get the point, and will try to do something

GvoFor: I tried to do this way, but there was an issue with select component. Cause when I usePagination hook called, it initializes page/pageSize states with DEFAULT_..., and it not simultaniously execute useEffect. So inside  `TablePagination` component first passed those default values, and useAppForm uses this default values. And after useEffect executes, those values, I passed inside useAppForm are not being updated, and it rendered not correct number. Also cause of useEffect executes not simultaniously, there is some ripple effect in search params (after page refresh, first the values are equals to defaults, and then they change to previous, which were before refresh -- here I expect, that they just stay as was before refresh, and not change for a moment)

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -5,43 +5,76 @@ import {
...
 import { getValidClassNames } from "~/libs/helpers/helpers.js";
-import { useFormController } from "~/libs/hooks/hooks.js";
+import { useCallback, useFormController } from "~/libs/hooks/hooks.js";
 import { type SelectOption } from "~/libs/types/types.js";
 
 import styles from "./styles.module.css";
 
 type Properties<TFieldValues extends FieldValues, TOptionValue> = {
 	control: Control<TFieldValues, null>;
+	isLabelVissible?: boolean;
```
liza-veis, line 18: I would say it's better to call it isLabelHidden, because visible is preferred option
```suggestion
	isLabelHidden?: boolean;
```

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -37,6 +39,11 @@ const Projects = (): JSX.Element => {
 		},
 	});
 
+	const TOTAL_PAGINATION_ITEMS = 100;
```
liza-veis, line 42:
```suggestion
	const TABLE_TOTAL_ITEMS = 100;
```

apps/frontend/src/libs/hooks/use-pagination/use-pagination.ts
```diff
@@ -0,0 +1,79 @@
...
+
+const usePagination = ({ totalItems }: Parameters): ReturnType => {
+	const [searchParameters, setSearchParameters] = useSearchParams();
+
+	const pageQueryParameter = searchParameters.get(QueryParameter.page);
+	const pageSizeQueryParameter = searchParameters.get(QueryParameter.pageSize);
+
+	const initialPageSize = getInitialPageSize(pageSizeQueryParameter);
+	const initialTotalPages = calculateTotalPages(initialPageSize, totalItems);
+	const initialPage = getInitialPage(pageQueryParameter, initialTotalPages);
```
liza-veis, line 35: We don't want this to be calculated on each re-render. There as an ability to pass function to useState to compute initial state, can't we do it this way?

apps/frontend/src/libs/hooks/use-pagination/libs/helpers/is-number-in-range.helper.ts
liza-veis, line 1: Let's remove this helper, conditions usually vary too much so we want to define them where we need

apps/frontend/src/libs/components/select/styles.module.css
```diff
@@ -18,13 +18,24 @@
 	border-radius: 6px;
 }
 
+.control-small {
+	min-height: 32px;
+	padding: 4.5px 7px;
```
liza-veis, line 23: round the value

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -5,43 +5,76 @@ import {
...
 
 const Select = <TFieldValues extends FieldValues, TOptionValue>({
 	control,
+	isLabelVissible = true,
 	isMulti = false,
 	label,
 	name,
+	onChange = (): void => {},
 	options,
+	placeholder = "Select...",
```
liza-veis, line 36: Let's not use default placeholders
```suggestion
	placeholder
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -5,43 +5,76 @@ import {
...
 type Properties<TFieldValues extends FieldValues, TOptionValue> = {
 	control: Control<TFieldValues, null>;
+	isLabelVissible?: boolean;
 	isMulti?: boolean;
 	label: string;
 	name: FieldPath<TFieldValues>;
+	onChange?: (option: SelectOption<TOptionValue>) => void;
 	options: SelectOption<TOptionValue>[];
+	placeholder?: string;
+	variant?: "primary" | "small";
```
liza-veis, line 25:
```suggestion
	size?: "default" | "small";
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -5,43 +5,76 @@ import {
...
 };
 
 const Select = <TFieldValues extends FieldValues, TOptionValue>({
 	control,
+	isLabelVissible = true,
 	isMulti = false,
 	label,
 	name,
+	onChange = (): void => {},
```
liza-veis, line 34: remove this, onChange should be controlled by react-hook-form. If you need additional on onChange, use watch from react-hook-form
```suggestion
```

apps/frontend/src/libs/hooks/use-pagination/libs/helpers/get-initial-page.helper.ts
```diff
@@ -0,0 +1,20 @@
...
+const getInitialPage = (
+	pageQueryParameter: null | string,
+	totalPages: number,
+): number => {
+	const pageParameter = parseQueryParameterToNumber(pageQueryParameter);
+
+	if (
+		pageParameter !== null &&
+		isNumberInRange(pageParameter, { max: totalPages, min: FIRST_PAGE })
+	) {
```
liza-veis, line 13: I think you don't need parseQueryParameterToNumber, isNumberInRange, you can simplify code without them
```suggestion
	const page = Number(pageQueryParameter);

	if (
		Number.isInteger(page) &&
		page <= totalPages &&
		page >= FIRST_PAGE 
	) {
```

apps/frontend/src/libs/hooks/use-pagination/libs/enums/query-parameter.enum.ts
```diff
@@ -0,0 +1,6 @@
+const QueryParameter = {
```
liza-veis, line 1: let's call it `QueryParameterName` and move to `apps/frontend/src/libs/enums/query-parameter-name.enum.ts`
```suggestion
const QueryParameterName = {
```

apps/frontend/src/libs/hooks/use-pagination/libs/constants/default-values.constant.ts
```diff
@@ -0,0 +1,6 @@
+const DEFAULT_VALUES = {
+	PAGE: 1,
+	PAGE_SIZE: 10,
+};
```
liza-veis, line 4: we will use FIRST_PAGE as the default page, so we don't need a separate constant for it
```suggestion
const DEFAULT_PAGE_SIZE = 10;
```
Don't forget about file name

apps/frontend/src/libs/components/table-pagination/styles.module.css
```diff
@@ -0,0 +1,38 @@
...
+}
+
+.change-page-buttons {
+	column-gap: 7px;
+}
+
+.total-items-text,
+.rows-per-page-text,
+.page-text {
+	padding: 5.5px 0;
```
liza-veis, line 36: round values and maybe instead we need to add padding to container? or make it display flex and just center everything

apps/frontend/src/libs/components/table-pagination/styles.module.css
```diff
@@ -0,0 +1,38 @@
+.container,
```
liza-veis, line 1: define these styles separately, and add them only where they are really needed. The same for styles below

apps/frontend/src/libs/components/table-pagination/libs/types/page-size-option.type.ts
liza-veis, line 1: I think we don't need separate type for this

apps/frontend/src/libs/components/table-pagination/libs/constants/page-size-options.constant.ts
```diff
@@ -0,0 +1,9 @@
+import { type PageSizeOption } from "../types/types.js";
+
+const options: PageSizeOption[] = [
```
liza-veis, line 3:
```suggestion
const PAGE_SIZE_OPTIONS: PageSizeOption[] = [
```

liza-veis (changes requested):

apps/frontend/src/libs/components/select/styles.module.css
GvoFor: But this will cause component be 33px height instead of 32px. Or this is not a big deal? or should I use `height: 32px` here?

apps/frontend/src/libs/hooks/use-pagination/use-pagination.ts
GvoFor: Sure. Don't know about this ability ^__^

apps/frontend/src/libs/hooks/use-pagination/libs/helpers/get-initial-page.helper.ts
GvoFor: Yeah, thats definitely better

apps/frontend/src/libs/components/table-pagination/libs/types/page-size-option.type.ts
GvoFor: Oh, It's just used in lots of places:
image
so I decide to separate this to a type (and also it acts as an alias)

But ok, this I'll remove this type. Or maybe it's better just to move it inside `table-pagination.tsx` ? but then I will need to import it in corresponding `.constant` file, which is imported here, and I don't like the fact we'll have cyclic dependencies. So I'd better remove it at all.

And what about `FormData` type? does it make sense to remove it too? or move it to `table-pagination.tsx` ?

apps/frontend/src/libs/components/icon-button/libs/constants/icon-size.constant.ts
```diff
@@ -0,0 +1,6 @@
+const ICON_SIZE = {
```
what1s1ove, line 1: it seems like an enum

GvoFor: Hm. So you suggest to move it into `apps/frontend/src/libs/enums/icon-size.enum.ts`? And I see it like:
```ts
const IconSize = {
  DEFAULT: 20;
} as const;
```

apps/frontend/src/libs/components/select/styles.module.css
GvoFor: I suppose it's not a big deal so I round it for now

apps/frontend/src/libs/components/icon-button/libs/constants/icon-size.constant.ts
GvoFor: Do this the way I described, but if you mean something different, please explain your suggestion

apps/frontend/src/libs/components/table-pagination/libs/types/page-size-option.type.ts
GvoFor: I moved these types into `table-pagination.tsx` for now

apps/frontend/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -1,38 +1,44 @@
 import { Icon } from "~/libs/components/components.js";
+import { IconSize } from "~/libs/enums/icon-size.enum.js";
```
Vitaliistf, line 2: Should be imported from` ~/libs/enums/enums.js`, I guess

GvoFor: Done

apps/frontend/src/libs/components/table-pagination/styles.module.css
```diff
@@ -0,0 +1,30 @@
...
+	justify-content: space-between;
+}
+
+.container {
+	height: 32px;
+	font-size: 14px;
+	font-weight: 400;
+	color: var(--color-text-primary);
+
+	/* TODO: remove the folowing bg-color style after visual testing */
```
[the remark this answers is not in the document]
what1s1ove: it not resolved

liza-veis (changes requested):

apps/frontend/src/libs/components/table-pagination/styles.module.css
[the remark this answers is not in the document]
GvoFor: done

GvoFor:
There is still an issue with select background-color. It's always `background-secondary` but in my case it should be kind of transparent:
image
image
Now I'm trying to figure out how to do this

GvoFor:
I came up with such a solution: I add additional prop `background: "primary" | "secondary"` to the `Select` and `TablePagination` components with default some values, and according to this prop I add corresponding styles.

To be honest, I'm not really happy with this solution (I wanted `TablePagination` to be transparent, so it's background mathes page background), but `Select` component is a bit tricky, and this is the only solution I came up.

GvoFor:
Forget to attach some screenshots

Here is how `TablePagination` component looks like (with default `background = "primary"`)
image

And here is how `Select` component looks like (with `background = "primary"`)
image

And here is how  `Select` component looks like (with default `background = "secondary"`)
image

apps/frontend/src/libs/components/table-pagination/table-pagination.tsx
```diff
@@ -0,0 +1,157 @@
...
+} from "./libs/constants/constants.js";
+import styles from "./styles.module.css";
+
+type Properties = {
+	background?: "primary" | "secondary";
+	onPageChange: (page: number) => void;
+	onPageSizeChange: (pageSize: number) => void;
+	page: number;
+	pageSize: number;
+	totalItems: number;
```
what1s1ove, line 26:
```suggestion
	totalItemsCount: number;
```

what1s1ove (approved):

apps/frontend/src/libs/components/table-pagination/table-pagination.tsx
GvoFor: Renamed everywhere

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -40,6 +42,11 @@ const Projects = (): JSX.Element => {
 		},
 	});
 
+	const TABLE_TOTAL_ITEMS_COUNT = 100;
```
what1s1ove, line 45: B28

apps/frontend/src/libs/hooks/use-pagination/use-pagination.ts
what1s1ove, line 1: apps/frontend/src/libs/hooks/use-pagination/use-pagination.hook.ts

apps/frontend/src/libs/hooks/use-pagination/use-pagination.ts
```diff
@@ -0,0 +1,83 @@
...
+type Parameters = {
+	totalItemsCount: number;
+};
+
+type ReturnType = {
+	onPageChange: (page: number) => void;
+	onPageSizeChange: (pageSize: number) => void;
+	page: number;
+	pageSize: number;
+};
```
what1s1ove, line 25:
```ts
type UsePagination = (params: {
totalItemsCount: number;
}) => {
	onPageChange: (page: number) => void;
	onPageSizeChange: (pageSize: number) => void;
	page: number;
	pageSize: number;
}
```

liza-veis:
Looks good to me, fix @what1s1ove comments

liza-veis (approved):

yevhenii-hladkov (approved):

PR #48, opened 2024-08-21
feat: add authenticated user flow gf-25

s1rserg:
1. Added `/auth/authenticated-user` endpoint to the controller
2. Updated find methods to find user by id. Need `gf-13` to extract injected userId from request

PR #52, opened 2024-08-22
feat: add authenticated user flow gf-25

s1rserg:
1. Added `/auth/authenticated-user` endpoint to the controller.

This PR replaces the previous [PR](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/48) due to an incorrect branch name. The branch has been renamed from `25-add-authenticated-user-flow` to `25-feat-add-authenticated-user-flow` to match the required naming conventions.


packages/shared/src/modules/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -1,4 +1,5 @@
 const AuthApiPath = {
+	AUTHENTICATED_USER: "/authenticated-user",
 	ROOT: "/",
```
perekotypole, line 3: We have connected tasks and Yelyzaveta asked me to remove ROOT. https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/49#discussion_r1725518554
Can you do this as well, so that later it does not accidentally return?

s1rserg: I’m keeping an eye on your PR too, so ROOT definitely won’t return)

liza-veis (approved):

yevhenii-hladkov (approved):

s1rserg:
Not sure If I can change white routes, because it's not my ticket, but it's necessary to get token to test the functionality of this PR. I just added `v1` to the white routes so it would work. If it's not needed I can get rid of this change.

liza-veis (approved):

liza-veis:
> Not sure If I can change white routes, because it's not my ticket, but it's necessary to get token to test the functionality of this PR. I just added v1 to the white routes so sign routes would work. If it's not needed I can get rid of this change.

yes, you can if your functionality requires this

Issue #53, opened 2024-08-22
feat: Add an ability to create a group


liza-veis:
### What feature?

- [ ] Add `permissions` table (columns: id, key, name, created_at, updated_at). Limit the name to 100 chars.
- [ ] Add `user_groups` table (columns: id, key, name, created_at, updated_at). Limit the name to 100 chars.
- [ ] Add `users_to_user_groups` table (columns: id, user_group_id, user_id, created_at, updated_at).
- [ ] Add `user_groups_to_permissions` table (columns: id, user_group_id, permission_id, created_at, updated_at).
- [ ] Add `manage_user_access` permission (name: "Manage User Access") through migration.
- [ ] Add needed models, and services on the backend.
- [ ] Add a controller for groups with `POST /groups` route
- [ ] The payload for group creation should contain name, user_ids, and permission_ids. Using this payload you will need to create an item in `user_groups`, then items in `user_groups_to_permissions` and then in `users_to_user_groups`. 
- [ ] The key should be automatically generated in the service by turning name into snake_case.
- [ ] The key should be unique for permissions and user_groups table (restriction on db level and check on backend as well)
- [ ] Update db schema in readme accordingly (rename groups table to user_groups, columns etc.)

liza-veis:
A validation on request payload should be added too.

VeronikaHordus:
Everythink works as ordered 👍 Thank you !

PR #56, opened 2024-08-22
feat: add an ability to create a group gf-53

kolibri753:
feat: Add an ability to create a group #53

1.  Add permissions table (columns: id, key, name, created_at, updated_at). Limit the name to 100 chars.
2.  Add user_groups table (columns: id, key, name, created_at, updated_at). Limit the name to 100 chars.
3.  Add users_to_user_groups table (columns: id, user_group_id, user_id, created_at, updated_at).
4.  Add user_groups_to_permissions table (columns: id, user_group_id, permission_id, created_at, updated_at).
5.  Add manage_user_access permission (name: "Manage User Access") through migration.
6.  Add needed models, and services on the backend.
7.  Add a controller for groups with POST /groups route
8.  The payload for group creation should contain name, user_ids, and permission_ids. Using this payload you will need to create an item in user_groups, then items in user_groups_to_permissions and then in users_to_user_groups.
9.  The key should be automatically generated in the service by turning name into snake_case.
10.  The key should be unique for permissions and user_groups table (restriction on db level and check on backend as well)
11.  Update db schema in readme accordingly (rename groups table to user_groups, columns etc.)

permissions
image

user_groups
image

user_groups_to_permissions
image

users_to_user_groups
image

POST
image

Success
image

Same key
image


apps/backend/src/modules/groups/group.controller.ts
```diff
@@ -0,0 +1,77 @@
...
+
+		this.groupService = groupService;
+
+		this.addRoute({
+			handler: (options) =>
+				this.createGroup(
+					options as APIHandlerOptions<{ body: GroupCreateRequestDto }>,
+				),
+			method: "POST",
+			path: "/create",
```
yevhenii-hladkov, line 28: Store path in enum

apps/backend/src/libs/helpers/transform-to-snake-case/transform-to-snake-case.helper.ts
```diff
@@ -0,0 +1,7 @@
+import snakeCase from "lodash/snakeCase.js";
+
+const transformToSnakeCase = (name: string): string => {
+	return snakeCase(name);
```
yevhenii-hladkov, line 4: I don't think it's a good idea to install lodash for this. You can create your own snakeCase

image

apps/backend/src/modules/groups/group.controller.ts
```diff
@@ -0,0 +1,77 @@
...
+	 *         description: Failed to create group
+	 */
+
+	private async createGroup(
+		options: APIHandlerOptions<{ body: GroupCreateRequestDto }>,
+	): Promise<APIHandlerResponse> {
+		await this.groupService.create(options.body);
+
+		return {
+			payload: { message: "Group created successfully" },
```
yevhenii-hladkov, line 71: Same

apps/backend/package.json
```diff
@@ -26,13 +26,15 @@
...
 		"@git-fit/shared": "*",
+		"@types/lodash": "4.17.7",
 		"bcrypt": "5.1.1",
 		"convict": "6.2.4",
 		"dotenv": "16.4.5",
 		"fastify": "4.28.1",
 		"fastify-plugin": "4.5.1",
 		"jose": "5.6.3",
 		"knex": "3.1.0",
+		"lodash": "4.17.21",
```
what1s1ove, line 37: we will not use "lodash" in our project

apps/backend/src/db/migrations/20240822140145_add_permissions_table.ts
```diff
@@ -0,0 +1,36 @@
+import { UserValidationRule } from "@git-fit/shared/src/modules/users/libs/enums/enums.js";
```
what1s1ove, line 1: you can not import anything in migrations

apps/backend/src/db/migrations/20240822141237_add_user_groups_table.ts
```diff
@@ -0,0 +1,36 @@
+import { UserValidationRule } from "@git-fit/shared/src/modules/users/libs/enums/enums.js";
```
what1s1ove, line 1:
```suggestion
```

apps/backend/src/libs/helpers/transform-to-snake-case/transform-to-snake-case.helper.ts
```diff
@@ -0,0 +1,7 @@
+import snakeCase from "lodash/snakeCase.js";
+
+const transformToSnakeCase = (name: string): string => {
```
what1s1ove, line 3: use https://www.npmjs.com/package/change-case this library instead. install it in the shared folder and create a wrapper around this function

apps/backend/src/modules/groups/group.controller.ts
what1s1ove, line 28:
```suggestion
			path: "/",
```
should be just a root

apps/backend/src/modules/groups/group.service.ts
```diff
@@ -0,0 +1,63 @@
...
+import { UserGroupModel } from "./user-group.model.js";
+import { UserGroupsToPermissionsModel } from "./user-groups-to-permissions.model.js";
+import { UsersToUserGroupModel } from "./users-to-user-group.model.js";
+
+class GroupService implements Service {
+	public async create(payload: GroupCreateRequestDto): Promise<void> {
+		const { name, permissionIds, userIds } = payload;
+		const key = transformToSnakeCase(name);
+
+		const existingGroup = await UserGroupModel.query().findOne({ key });
```
perekotypole, line 16: Services shouldn't work with database. You have to create a repository for it

packages/shared/src/modules/groups/libs/validation-schemas/create-group.validation-schema.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { GroupValidationMessage, GroupValidationRule } from "../enums/enums.js";
+import { type GroupCreateRequestDto } from "../types/types.js";
+
+const groupCreate: z.ZodType<GroupCreateRequestDto> = z.object({
+	name: z.string().max(GroupValidationRule.NAME_MAXIMUM_LENGTH, {
+		message: GroupValidationMessage.NAME_TOO_LONG,
+	}),
+	permissionIds: z.array(z.number().int().positive()),
+	userIds: z.array(z.number().int().positive()),
```
perekotypole, line 11: Will there be no error messages?

apps/backend/src/modules/groups/user-groups-to-permissions.model.ts
```diff
@@ -0,0 +1,20 @@
...
+	DatabaseTableName,
+} from "~/libs/modules/database/database.js";
+
+class UserGroupsToPermissionsModel extends AbstractModel {
+	public permissionId!: number;
+
+	public userGroupId!: number;
+
+	public static override get idColumn(): string {
+		return "id";
```
perekotypole, line 12: I can't get why you return "id" string here, could you explain?

packages/shared/src/modules/groups/libs/validation-schemas/create-group.validation-schema.ts
kolibri753: Oh... I was planning to add them later, but I forgot

apps/backend/src/modules/groups/group.service.ts
kolibri753: That's a good point, but the issue only mentioned adding services and models, so I initially did it this way.

apps/backend/src/modules/groups/user-groups-to-permissions.model.ts
kolibri753: Maybe it's not needed. I just thought of adding this to tables that represent a many-to-many relationship.

apps/backend/src/db/migrations/20240822143738_add_manage_user_access_to_permission.ts
```diff
@@ -0,0 +1,24 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "permissions";
+
+const PermissionName = {
+	MANAGE_USER_ACCESS: "Manage User Access",
+};
```
liza-veis, line 7: Use permission key instead

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -0,0 +1,84 @@
...
+class GroupRepository implements Repository {
+	private groupModel: typeof UserGroupModel;
+
+	public constructor(groupModel: typeof UserGroupModel) {
+		this.groupModel = groupModel;
+	}
+
+	public async create(entity: GroupEntity): Promise<GroupEntity> {
+		const { name, permissionIds, userIds } = entity.toNewObject();
+		const key = changeToSnakeCase(name);
```
liza-veis, line 18:
```suggestion
		const key = changeCase(name, "snakeCase");
```

packages/shared/src/modules/groups/libs/enums/group-validation-message.enum.ts
```diff
@@ -0,0 +1,10 @@
+import { GroupValidationRule } from "./group-validation-rule.enum.js";
+
+const GroupValidationMessage = {
+	NAME_REQUIRED: "Project name is required.",
+	NAME_TOO_LONG: `Name is too long (> ${String(GroupValidationRule.NAME_MAXIMUM_LENGTH)} symbols).`,
+	PERMISSION_IDS_REQUIRED: "At least one permission ID is required.",
```
liza-veis, line 6: We should allow to create groups without permissions

liza-veis (changes requested):

packages/shared/src/modules/groups/libs/validation-schemas/create-group.validation-schema.ts
GvoFor, line 1: Rename to `group-create.validation-schema.ts`

packages/shared/src/modules/groups/libs/enums/group-validation-message.enum.ts
kolibri753: Screenshot from 2024-08-27 13-02-30

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -0,0 +1,88 @@
+import { changeCase } from "~/libs/helpers/helpers.js";
+import { type Repository } from "~/libs/types/types.js";
+
+import { GroupEntity } from "./group.entity.js";
+import { type UserGroupModel } from "./user-group.model.js";
+import { UserGroupsToPermissionsModel } from "./user-groups-to-permissions.model.js";
+import { UsersToUserGroupModel } from "./users-to-user-group.model.js";
+
+const MINIMUM_PERMISSIONS = 0;
```
kolibri753, line 9: Should I move `MINIMUM_PERMISSIONS` to `~/libs/constants/constants.js`?

what1s1ove (approved):

packages/shared/src/modules/groups/libs/validation-schemas/group-create.validation-schema.ts
```diff
@@ -0,0 +1,22 @@
...
+	name: z
+		.string()
+		.trim()
+		.min(GroupValidationRule.NAME_MINIMUM_LENGTH, {
+			message: GroupValidationMessage.NAME_REQUIRED,
+		})
+		.max(GroupValidationRule.NAME_MAXIMUM_LENGTH, {
+			message: GroupValidationMessage.NAME_TOO_LONG,
+		}),
+	permissionIds: z.array(z.number().int().positive()).optional(),
```
liza-veis, line 16: I would say it's not optional, it's required in payload, but the array itself can be empty

packages/shared/src/modules/groups/libs/types/group-create-request-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+type GroupCreateRequestDto = {
+	name: string;
+	permissionIds?: number[] | undefined;
```
liza-veis, line 3:
```suggestion
	permissionIds: number[];
```

packages/shared/src/libs/helpers/change-case/types/case-type.type.ts
liza-veis, line 1: you missed libs folder in  file path, should be `change-case/libs/types/case-type.type.ts`

apps/backend/src/db/migrations/20240822143738_add_manage_user_access_to_permission.ts
liza-veis: Still not fixed

apps/backend/src/modules/groups/user-groups-to-permissions.model.ts
```diff
@@ -0,0 +1,20 @@
+import {
+	AbstractModel,
+	DatabaseTableName,
+} from "~/libs/modules/database/database.js";
+
+class UserGroupsToPermissionsModel extends AbstractModel {
```
liza-veis, line 6: I believe we can get rid of these "through" models, like something to something. We can define relationMappings in Groups model. Like:

```javascript
	permissions: {
				join: {
					from: '',
					through: { // groups to permissions table relation here
						from: ''
						to: '',
					to: '',
				},
				modelClass: PermissionModel,
				relation: Model.ManyToManyRelation,
			},
```

And then we can also simplify the logic in repositories and services to have something like this:

```javascript
// on service level payload will look like this
const group = {
  // ...props to create a group
  permissions: permissionIds,
  users: userIds,
}

// on repository level
const createdGroup = await this.groupModel
			.query()
			.insertGraph(group.toNewObject(), { relate: true })
			.returning("*")
			.withGraphJoined(RelationName.PERMISSIONS)
             .withGraphJoined(RelationName.USERS)
			.execute();
```
ps. the above is just an example, it may look different

liza-veis (changes requested):

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -0,0 +1,75 @@
...
+			name,
+			permissions: permissionIds.map((id) => ({ id })),
+			users: userIds.map((id) => ({ id })),
+		};
+
+		const group = await this.groupModel
+			.query()
+			.insertGraph(groupData, { relate: true })
+			.returning("*")
+			.withGraphJoined("[permissions, users]");
```
kolibri753, line 29: Should I create RelationName enum for this?

apps/backend/src/db/migrations/20240822143738_add_manage_user_access_to_permission.ts
kolibri753: This is why I am confused... I've already done it

```
function down(knex: Knex): Promise<void> {
	return knex(TABLE_NAME)
		.where({
			key: PermissionKey.MANAGE_USER_ACCESS,
		})
		.del();
}
```

yevhenii-hladkov (approved):

apps/backend/src/modules/groups/group.model.ts
```diff
@@ -0,0 +1,53 @@
...
+				join: {
+					from: `${DatabaseTableName.USER_GROUPS}.id`,
+					through: {
+						from: `${DatabaseTableName.USER_GROUPS_TO_PERMISSIONS}.userGroupId`,
+						to: `${DatabaseTableName.USER_GROUPS_TO_PERMISSIONS}.permissionId`,
+					},
+					to: `${DatabaseTableName.PERMISSIONS}.id`,
+				},
+				modelClass: PermissionModel,
+				relation: Model.ManyToManyRelation,
```
kolibri753, line 31: When I have my electricity back, I'll check if I can put AbstractModel here (instead of Model)

apps/backend/src/modules/groups/group.model.ts
```diff
@@ -0,0 +1,53 @@
+import { Model, type RelationMappings } from "objection";
```
kolibri753, line 1: Instead of `Model`, I'll use `AbstractModel`, but what about `type RelationMappings`. Shoul I re-export it from other place, or is it okay to import it here from "objection"?

kolibri753:
I assume that I should also check that users and permissions exist before creating a group

apps/backend/src/modules/groups/group.model.ts
```diff
@@ -0,0 +1,53 @@
...
+import { UserModel } from "~/modules/users/user.model.js";
+
+class GroupModel extends AbstractModel {
+	public key!: string;
+
+	public name!: string;
+
+	public permissions!: { id: number }[];
+
+	public users!: { id: number }[];
```
GvoFor, line 17: Why can't we use just `number[]` instead of `{ id: number}[]` for `permissions` and `users`?
I just wanna know, cause belive you do it correct

kolibri753: I think that Objection.js requires an object with an id field to properly handle many-to-many relationships.

apps/backend/src/db/migrations/20240822143738_add_manage_user_access_to_permission.ts
liza-veis: oh, didn't see it, then ok

packages/shared/src/libs/helpers/change-case/libs/maps/case-type-to-function.map.ts
```diff
@@ -0,0 +1,10 @@
+import { camelCase, snakeCase } from "change-case";
+
+import { type CaseType } from "../types/types.js";
+
+const caseTypeToFunction: Record<CaseType, (input: string) => string> = {
```
liza-veis, line 5: If this a map, it should be just key/value object. So either move it to helpers and name it appropriately or use object instead of function here (I would prefer the second option)

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+		const { name, permissionIds, userIds } = entity.toNewObject();
+		const key = changeCase(name, "snakeCase");
+
+		const trx = await transaction.start(this.groupModel.knex());
+
+		const groupData = {
+			key,
+			name,
+			permissions: permissionIds.map((id) => ({ id })),
+			users: userIds.map((id) => ({ id })),
```
liza-veis, line 26: This mapping should be on service level

packages/shared/src/libs/helpers/change-case/libs/maps/case-type-to-function.map.ts
kolibri753: Do you mean like this?
```
const caseTypeFunctions = {
	camelCase: (input: string) => camelCase(input),
	snakeCase: (input: string) => snakeCase(input),
} as Record<CaseType, (input: string) => string>;
```

liza-veis:
```typescript
const caseTypeToFunction = {
	camelCase,
	snakeCase,
}
```

kolibri753: Oh, now I see :cat2:

apps/backend/src/modules/groups/group.entity.ts
```diff
@@ -0,0 +1,86 @@
...
+			name,
+			permissionIds,
+			userIds,
+		});
+	}
+
+	public toNewObject(): {
+		name: string;
+		permissionIds: { id: number }[];
+		userIds: { id: number }[];
```
yevhenii-hladkov, line 67: ` { id: number } ` is repeated 12 times. I would create a separate type for this. At least to make it look more attractive

liza-veis: I think we just don't need these props in the model, we need permissions and users that will have type Pick<OurCommonUserType, 'id'>

kolibri753: Hope I understand correctly
```
permissions: Pick<PermissionModel, "id">[];
users: Pick<UserModel, "id">[];
```

kolibri753: and also delete in the group.model.ts

```
public permissions!: { id: number }[];
public users!: { id: number }[];
```

what1s1ove (approved):

apps/backend/src/modules/groups/group.entity.ts
liza-veis: yes, rename it to permissions and users instead of permissionIds, we may pass additional fields when creating items.

liza-veis: let's use this way if model allows so
> permissions: Pick<PermissionModel, "id">[];

apps/backend/src/modules/groups/group.entity.ts
```diff
@@ -0,0 +1,88 @@
...
+		};
+	}
+
+	public toObject(): GroupCreateResponseDto {
+		return {
+			id: this.id as number,
+			name: this.name,
+			permissionIds: this.permissions.map((perm) => perm.id),
+			userIds: this.users.map((user) => user.id),
+		};
```
liza-veis, line 84: we don't want to have this logic in the entity and I think we will need to return it in object format
```suggestion
			permissions: this.permissions,
			users: this.users,
		};
```

liza-veis (changes requested):

apps/backend/src/modules/groups/group.entity.ts
kolibri753: to remove it I can refactor GroupCreateResponseDto like this
```
type GroupCreateResponseDto = {
	id: number;
	name: string;
	permissions: { id: number }[];
	users: { id: number }[];
};
```

liza-veis: yes, this will be ok

liza-veis (approved):

PR #59, opened 2024-08-22
feat: add users table gf-41

AnutkaGn:
### Summary
 
Add access management item to the sidebar
 Add common page layout.
 Add users table to the page. It have only "Name" and "Created At" columns.
 Data taken from the current users table on the backend.
 
image


AnutkaGn:
Now the page looks like this. I'm waiting for the sidebar
image

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -0,0 +1,40 @@
...
+		}
+	}, [isAccessManagement, dispatch]);
+
+	const transformedData: User[] = transformUserData(users);
+
+	return (
+		<div className={styles["content"]}>
+			<p className={styles["content__title"]}>Access Management</p>
+			<p className={styles["content__sub-title"]}>Users</p>
+			<div className={styles["content__table"]}>
```
kolibri753, line 33: We agreed on writing styles in kebab-case, like `.logo-wrapper`

apps/frontend/src/pages/common-layout/common-layout.tsx
```diff
@@ -0,0 +1,27 @@
...
+	children: ReactNode;
+};
+
+const CommonLayout = ({ children }: CommonLayoutProperties): JSX.Element => {
+	return (
+		<div className={styles["container"]}>
+			<Header />
+			<div className={styles["container__sidebar-content"]}>
+				<div
+					style={{ backgroundColor: "white", height: "100vh", width: "280px" }}
```
kolibri753, line 17: try not to use vh unit if possible (better % or flex)

AnutkaGn: These styles will then be removed when the header and sidebar are added

apps/frontend/src/pages/access-management/libs/helpers/format-user-date/format-user-data.helper.ts
GvoFor, line 1: Don't think we need this extra nesting for helpers. Put `format-use-data.helper.ts` right into the `helpers` folder (ridding of one extra level of nesting)

apps/frontend/src/pages/access-management/libs/helpers/transform-user-data/transform-user-data.helper.ts
GvoFor, line 1: Same as for `format-use-data.helper.ts`

apps/frontend/src/pages/access-management/styles.module.css
GvoFor, line 1: Don't forget about kebab as @kolibri753 mentioned )

apps/frontend/src/pages/access-management/libs/enums/name-column-table.enum.ts
GvoFor, line 1: Not sure, if this is an enum

apps/frontend/src/pages/access-management/libs/enums/name-column-table.enum.ts
```diff
@@ -0,0 +1,10 @@
+import { type ColumnDef } from "@tanstack/react-table";
+
+import { type User } from "../types/user.type.js";
+
+const columns: ColumnDef<User>[] = [
+	{ accessorKey: "name", header: "Name" },
+	{ accessorKey: "createdAt", header: "Created At" },
+];
```
sandrvvu, line 8: I think this fits better as constants rather than enums. If so, then B22 too, but maybe hold off for now

apps/frontend/src/pages/access-management/libs/enums/name-column-table.enum.ts
GvoFor: and maybe better to change `name-column` to `column-name` or `column-header` everywhere it appears (so for example instead of `nameOfColumns` you will use `columnName` or `columnHeader`)

apps/frontend/src/pages/access-management/libs/helpers/format-user-date/format-user-data.helper.ts
```diff
@@ -0,0 +1,14 @@
+const formatUserDate = (createdAt: string): string => {
```
sandrvvu, line 1: maybe it would be better to generalize formatUserDate into a formatDate function. this way, it can be used in other contexts as well

apps/frontend/src/libs/helpers/format-data/format-data.helper.ts
```diff
@@ -0,0 +1,14 @@
+const formatDate = (createdAt: string): string => {
```
what1s1ove, line 1: it should be in the shared module. let's use `date-fns` library for it

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { columnHeader } from "./libs/constants/constants.js";
+import { transformUserData } from "./libs/helpers/helpers.js";
+import { type User } from "./libs/types/types.js";
+import styles from "./styles.module.css";
+
+const AccessManagement = (): JSX.Element => {
+	const dispatch = useAppDispatch();
+	const { pathname } = useLocation();
+	const users = useAppSelector(({ users }) => users.users);
+	const isAccessManagement = pathname === AppRoute.ACCESS_MANAGEMENT;
```
what1s1ove, line 20:
```suggestion
```
the page should be accessible directly by that route

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -0,0 +1,40 @@
...
+	}, [isAccessManagement, dispatch]);
+
+	const transformedData: User[] = transformUserData(users);
+
+	return (
+		<div className={styles["content"]}>
+			<p className={styles["title"]}>Access Management</p>
+			<p className={styles["sub-title"]}>Users</p>
+			<div className={styles["table-container"]}>
+				<Table<User> columns={columnHeader} data={transformedData} />
```
what1s1ove, line 34:
```suggestion
     const headers = getHeaders()
     const columns = getColumns()

				<Table<User> columns={headers} data={columns} />
```
let's do it this way

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -67,6 +75,7 @@ class UserEntity implements Entity {
 		passwordSalt: string;
 	}): UserEntity {
 		return new UserEntity({
+			createdAt: "",
```
GvoFor, line 78: Not sure, but maybe it's better to bring back `createdAt` initializing, as it was in constructor:
```suggestion
			createdAt: new Date().toISOString(),
```
so when we call this method, this field won't be empty string, but contain real created at date

apps/frontend/src/pages/access-management/libs/constants/column-header.constant.ts
```diff
@@ -0,0 +1,10 @@
+import { type ColumnDef } from "@tanstack/react-table";
+
+import { type User } from "../types/user.type.js";
```
GvoFor, line 3:
```suggestion
import { type User } from "../types/types.js";
```

apps/frontend/src/pages/access-management/styles.module.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.sub-title {
+	font-size: 24px;
+	font-weight: 600;
+	line-height: 120%;
+}
+
+.table-container {
+	width: 70%;
```
GvoFor, line 21: Where this number comes from? You'd better remove this at all, it should have fill all available space (so `width: 100%`, which is defaults). Instead of this `.content` will have appropriate `padding`

apps/frontend/src/pages/access-management/styles.module.css
```diff
@@ -0,0 +1,22 @@
+.content {
+	display: flex;
+	flex-direction: column;
+	width: 100%;
+	padding-left: 10%;
```
GvoFor, line 5: According to Figma design
```suggestion
	padding: 32px 80px;
```

GvoFor: Or maybe even it would be better to move this padding to the `CommonLayout` component, as it is the same for all pages

apps/frontend/src/pages/common-layout/styles.module.css
```diff
@@ -0,0 +1,19 @@
...
+	display: flex;
+	flex-direction: row;
+	width: 100%;
+}
+
+.content-container {
+	display: flex;
+	flex-direction: column;
+	width: 100%;
+}
```
GvoFor, line 19: As I mentioned above, maybe it would be better to add `padding: 32px 80px` here, instead of adding its for every component like `AccessManagement`

apps/backend/src/modules/users/user.entity.ts
AnutkaGn: I don't think it needs to be done this way because there is a function in the abstract.model.ts file that adds build time when adding an object to the database

GvoFor: Oh, then it's okay as it is

liza-veis:
Now we have sidebar, so you can add it.

packages/shared/src/libs/helpers/format-date/format-date.helper.ts
```diff
@@ -0,0 +1,7 @@
+import { format } from "date-fns";
+
+const formatDate = (createdAt: string): string => {
```
liza-veis, line 3: If this is a common helper it should accept any date and allowed formats
```suggestion
type DateFormat = 'd MMM yyyy HH:mm'

const formatDate = (date: Date, format: DateFormat): string => {
```

package.json
```diff
@@ -68,5 +68,8 @@
 	"simple-git-hooks": {
 		"pre-commit": "npm run lint:type && npm run git:hook:precommit",
 		"commit-msg": "npm run git:hook:commit"
+	},
+	"dependencies": {
+		"date-fns": "3.6.0"
```
liza-veis, line 73: should be in shared dependencies

apps/frontend/src/pages/common-layout/styles.module.css
```diff
@@ -0,0 +1,20 @@
...
+	display: flex;
+	flex-direction: row;
+	width: 100%;
+}
+
+.content-container {
+	display: flex;
+	flex: 1;
+	flex-direction: column;
+	padding: 0 80px 35px;
```
liza-veis, line 19: As I remember on design all dimensions, padding and margins should be multiples of 8. This is a common pattern for layouts.

apps/frontend/src/pages/common-layout/common-layout.tsx
```diff
@@ -0,0 +1,22 @@
+import { Header, Sidebar } from "~/libs/components/components.js";
+import { SIDEBAR_ITEMS } from "~/libs/constants/navigation-items.constant.js";
+
+import styles from "./styles.module.css";
+
+type CommonLayoutProperties = {
+	children: React.ReactNode;
+};
+
+const CommonLayout = ({ children }: CommonLayoutProperties): JSX.Element => {
```
liza-veis, line 10:
```suggestion
const PageLayout = ({ children }: CommonLayoutProperties): JSX.Element => {
```

apps/frontend/src/pages/common-layout/common-layout.tsx
```diff
@@ -0,0 +1,22 @@
+import { Header, Sidebar } from "~/libs/components/components.js";
+import { SIDEBAR_ITEMS } from "~/libs/constants/navigation-items.constant.js";
+
+import styles from "./styles.module.css";
+
+type CommonLayoutProperties = {
```
liza-veis, line 6:
```suggestion
type Properties = {
```

apps/frontend/src/pages/access-management/libs/types/user.type.ts
liza-veis, line 1: What is this for? we should use the type returned from the backend

apps/frontend/src/pages/access-management/libs/helpers/transform-user-data.helper.ts
```diff
@@ -0,0 +1,12 @@
+import { formatDate } from "~/libs/helpers/helpers.js";
+
+import { type User } from "../types/types.js";
+
+const transformUserData = (users: User[]): User[] => {
```
liza-veis, line 5: Let's remove this and pass the data as it's returned. This data is only used for this table, so if we wanted to re-map it on the frontend we should've done it on API level, not inside the components. But in our case we don't remap data returned from the backend and because this data is only supposed to be used for this table, we can restrict properties on the backend side if needed or pass all the fields from usual entity. So what I mean here, that in this case I would just remove this mapper and pass data directly

apps/frontend/src/pages/access-management/libs/constants/column-header.constant.ts
```diff
@@ -0,0 +1,10 @@
+import { type ColumnDef } from "@tanstack/react-table";
+
+import { type User } from "../types/types.js";
+
+const columnHeader: ColumnDef<User>[] = [
```
liza-veis, line 5:
```suggestion
const TABLE_COLUMNS: ColumnDef<User>[] = [
```
don't forget about file name

apps/backend/src/modules/users/user.entity.ts
liza-veis, line 78: Should have the same approach as with id here, pass null instead

liza-veis (changes requested):

apps/frontend/src/pages/access-management/libs/types/user.type.ts
AnutkaGn: This type is also used in the table like this:
<Table<User> columns={headers} data={data} />

apps/frontend/src/pages/access-management/libs/helpers/transform-user-data.helper.ts
AnutkaGn: I'm not sure if I understood correctly, but I tried to correct it

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -0,0 +1,37 @@
...
+	const dispatch = useAppDispatch();
+	const users = useAppSelector(({ users }) => users.users);
+
+	useEffect(() => {
+		void dispatch(userActions.loadAll());
+	}, [dispatch]);
+
+	const columns: UserAuthResponseDto[] = users.map((user) => ({
+		...user,
+		createdAt: formatDate(new Date(user.createdAt), "d MMM yyyy HH:mm"),
```
liza-veis, line 25: we should do this not in data, but in a place where we define columns. We should be able to pass a function to the accessor field in columns. Also instead of having TABLE_COLUMNS constant, let's have getUserColumns helper (in case we need to change columns them dynamically based on params)

apps/frontend/src/pages/access-management/libs/types/user.type.ts
liza-veis, line 1: We don't need this type, remove it

liza-veis (changes requested):

apps/frontend/src/pages/access-management/libs/types/user.type.ts
AnutkaGn: This type is also used in the table like this:
`<Table<User> columns={headers} data={data} />`
I don't understand why it should be removed, since it is used to define columns in the table, maybe it should be named differently

apps/frontend/src/pages/access-management/access-management.tsx
AnutkaGn: I don't understand, since an array of complete user objects is returned from the backend, but for this user table we only need the fields "Name", "Group" and "Created At" (for my task only "Name" and "Created At"). Why then can't we store column names in constants if they are not variables?

Do you think it should look something like this?
```
import { formatDate, type UserAuthResponseDto } from "@git-fit/shared";

export const getUserColumns = () => {
return [
 {
 Header: "Name",
 accessor: "name",
 },
 {
 Header: "Email",
 accessor: "email",
 },
 {
 Header: "Created At",
 accessor: (user: UserAuthResponseDto) =>
 formatDate(new Date(user.createdAt), "d MMM yyyy HH:mm"),
 },
 ];
};
```

apps/frontend/src/pages/access-management/libs/types/user.type.ts
liza-veis: After you move the logic with formatting data to the columns (comment above), you will no longer need a separate type for user data

apps/frontend/src/pages/page-layout/page-layout.tsx
```diff
@@ -0,0 +1,22 @@
+import { Header, Sidebar } from "~/libs/components/components.js";
+import { SIDEBAR_ITEMS } from "~/libs/constants/navigation-items.constant.js";
```
Vitaliistf, line 2: Should be imported from `~/libs/constants/constants.js` i think.

apps/frontend/src/libs/helpers/get-user-columns/get-user-columns.ts
```diff
@@ -0,0 +1,20 @@
+import { type ColumnDef } from "@tanstack/react-table";
```
liza-veis, line 1: Re-export this type as TableColumn in common types, then use it in Table component too. We don't want to use library import everywhere

liza-veis (changes requested):

apps/frontend/src/libs/helpers/get-user-columns/get-user-columns.ts
what1s1ove, line 1: add `.helper` to the filename

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -0,0 +1,34 @@
...
+const AccessManagement = (): JSX.Element => {
+	const dispatch = useAppDispatch();
+	const users = useAppSelector(({ users }) => users.users);
+
+	useEffect(() => {
+		void dispatch(userActions.loadAll());
+	}, [dispatch]);
+
+	const columns = getUserColumns();
+	const data: UserAuthResponseDto[] = users;
```
what1s1ove, line 23:
```suggestion
	const data: UserAuthResponseDto[] = getUserRows();
```
you need always create a helper for it even if you can just use an object keys

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -1,21 +1,21 @@
 import {
-	type ColumnDef,
 	flexRender,
 	getCoreRowModel,
+	type TableColumn,
 	useReactTable,
-} from "@tanstack/react-table";
+} from "~/libs/types/types.js";
```
liza-veis, line 6: We want to keep only TableColumn in common types, all others are not supposed to be used elsewhere except the Table component itself.

so imports should look like this:
```
import {
	flexRender,
	getCoreRowModel,
	useReactTable,
} from "@tanstack/react-table";
import { type TableColumn } from "~/libs/types/types.js";
```

liza-veis (changes requested):

apps/frontend/src/libs/components/table/table.tsx
AnutkaGn: Ok, I'll fix it now

apps/frontend/src/pages/access-management/access-management.tsx
AnutkaGn: I don't understand what the function is for, what exactly is it supposed to do?

apps/frontend/src/libs/components/page-layout/styles.module.css
```diff
@@ -0,0 +1,41 @@
+:root {
+	--header-height: 73px;
+}
```
GvoFor: Look at [this comment](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/105#discussion_r1734270942)

apps/frontend/src/pages/access-management/access-management.tsx
GvoFor: I think he means that it should look like this:
```ts
const getUserRows = (users: UserGetAllItemResponseDto[]): UserGetAllItemResponseDto[] =>
  users.map(({ name, ... }) => { name, ... });
```
(not sure if I write types correct)
So the point is that this function will decide which fields of users to display in rows (same as `getUserColumns` decide which columns to dispaly)

what1s1ove: right, you need only to take the keys that you need

AnutkaGn: Good, but then it turns out that we return to the fact that the User type will be needed, which @liza-veis told me to remove
```
import { type UserAuthResponseDto } from "~/modules/users/users.js";

type User = {
    name: string,
    createdAt: string,
}


const getUserRows = (users: UserAuthResponseDto[]): User[] =>
    users.map((user) => ({
      name: user.name,
      createdAt: user.createdAt,
}));

export { getUserRows };
```

AnutkaGn: Should I add this type again and add a mapper?

what1s1ove: yeah, you need to create a new return type for it

what1s1ove (approved):

apps/frontend/src/libs/helpers/get-user-rows/get-user-rows.helpers.ts
```diff
@@ -0,0 +1,10 @@
+import { type TableUser } from "~/libs/types/types.js";
```
liza-veis, line 1: this shouldn't be in common helpers, should be in page helpers the same for TableUser (let's rename it to UserRow)

liza-veis (changes requested):

apps/frontend/src/libs/helpers/get-user-rows/get-user-rows.helpers.ts
AnutkaGn: Maybe move the get-user-columns.helper.ts file there too?

liza-veis: yes

apps/frontend/src/libs/helpers/get-user-columns/get-user-columns.helper.ts
```diff
@@ -0,0 +1,19 @@
+import { formatDate } from "~/libs/helpers/helpers.js";
```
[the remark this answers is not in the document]
AnutkaGn: The formatDate function is located in the shared folder. In my opinion, it is a universal function, and there is no need to move it there.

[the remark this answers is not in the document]
AnutkaGn: Should only the import be moved there, and not the function?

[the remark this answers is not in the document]
liza-veis: I mean not format get-user-columns helper

liza-veis (approved):

PR #60, opened 2024-08-23
feat: add sign-in flow gf-4

s1rserg:
1. Added `/sign-in` route.
2. Added sign-in form and sign-in action flow.
3. Saving an authenticated user to the redux store.
4. Added form validation (shared).
5. Implemented design.
6. "Create new" link navigates to the sign-up page.
image
image


s1rserg:
I have a question regarding [this requested change](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/37#discussion_r1723150813) on `5-feat-add-sign-up-flow`. 

The issue I'm facing is that without using async/await or .then() to wait for the auth/sign-up action to be fulfilled, the user ends up on the sign-in page instead of being redirected to the root. After the form submission, the user should be redirected to the root page, but without waiting for the completion of the auth/sign-up action, the ProtectedRoute can't perform the redirection because there is no authenticatedUser in the Redux store yet.

Could you please advise if using await here is the right approach, or if there is another solution to ensure proper redirection?

Thank you!

liza-veis:
Hi! The logic here is that after sign-up/sign-in we should set the authenticated user that we receive with those requests to the store. Then in the auth page,  check if you there is a user in the store and if there is, then redirect the user to the main page

apps/frontend/src/pages/auth/components/sign-in-form/sign-in-form.tsx
perekotypole, line 1: I'm not sure, but isn't it better to create the layout for the sign-in and sign-up forms rather than creating two identical pages?

liza-veis: Yes, agree. Let's move the common part to the auth, and have only form inside sign-up/sign-in form components

apps/frontend/src/pages/auth/components/sign-in-form/sign-in-form.tsx
```diff
@@ -1,16 +1,89 @@
...
+	}, []);
+
+	return (
+		<section className={styles["auth-container"]}>
+			<div className={styles["left-side"]}>
+				{/* TODO: replace logo */}
+				<img alt="logo" className={styles["logo-wrapper"]} src="" />
+			</div>
+			<div className={styles["right-side"]}>
+				<h3 className={styles["form-title"]}>Welcome back</h3>
```
what1s1ove, line 49:
```suggestion
```
These styles should be on the auth page. You should reuse them for sign-in/sign-up forms

what1s1ove (approved):

yevhenii-hladkov (approved):

liza-veis (approved):

liza-veis (approved):

PR #74, opened 2024-08-23
fix: correct white routes constant and token extraction gf-13

sandrvvu:
This PR fixes [https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/43](url)
- Adjust white routes constant
- Refactor token extraction to remove 'Bearer ' prefix

### DIFF
--- a/apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
+++ b/apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
@@ -28,21 +28,17 @@ const authorization = fp<Options>((fastify, options, done) => {
 			}
 		}
 
+		const BEARER_PREFIX = "Bearer ";
 		const authHeader = request.headers["authorization"];
 
-		if (!authHeader) {
+		if (!authHeader || !authHeader.startsWith(BEARER_PREFIX)) {
 			throw new AuthError({
 				message: ExceptionMessage.NO_TOKEN_PROVIDED,
 				status: HTTPCode.UNAUTHORIZED,
 			});
 		}
 
-		const BEARER_PREFIX = "Bearer ";
-		const BEARER_PREFIX_LENGTH = BEARER_PREFIX.length;
-
-		const jwtToken = authHeader.startsWith(BEARER_PREFIX)
-			? authHeader.slice(BEARER_PREFIX_LENGTH)
-			: authHeader;
+		const jwtToken = authHeader.slice(BEARER_PREFIX.length);
 
 		try {
 			const payload = await token.verifyToken(jwtToken);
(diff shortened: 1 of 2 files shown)


apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -28,15 +28,22 @@ const authorization = fp<Options>((fastify, options, done) => {
...
 				status: HTTPCode.UNAUTHORIZED,
 			});
 		}
 
+		const BEARER_PREFIX = "Bearer ";
+		const BEARER_PREFIX_LENGTH = BEARER_PREFIX.length;
+
+		const jwtToken = authHeader.startsWith(BEARER_PREFIX)
+			? authHeader.slice(BEARER_PREFIX_LENGTH)
+			: authHeader;
```
liza-veis, line 45: We can do this easier
```suggestion
		const [, jwtToken] = authHeader.split(BEARER_PREFIX);
```

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -28,15 +28,22 @@ const authorization = fp<Options>((fastify, options, done) => {
 			}
 		}
 
-		const jwtToken = request.headers["authorization"];
+		const authHeader = request.headers["authorization"];
 
-		if (!jwtToken) {
+		if (!authHeader) {
```
liza-veis, line 33: let's move this check here
```suggestion
		if (!authHeader || !authHeader.startsWith(BEARER_PREFIX)) {
```

liza-veis (changes requested):

apps/backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -28,15 +28,22 @@ const authorization = fp<Options>((fastify, options, done) => {
...
 			throw new AuthError({
 				message: ExceptionMessage.NO_TOKEN_PROVIDED,
 				status: HTTPCode.UNAUTHORIZED,
 			});
 		}
 
+		const BEARER_PREFIX = "Bearer ";
+		const BEARER_PREFIX_LENGTH = BEARER_PREFIX.length;
+
+		const jwtToken = authHeader.startsWith(BEARER_PREFIX)
```
GvoFor, line 43: I may be wrong, but in our case authorization header always starts with "Bearer ", so there is no need in this check

liza-veis: We don't know whether this is a request we send from our frontend, so we should check. Also, backend doesn't know anything about how it's implemented on the frontend.

yevhenii-hladkov (approved):

liza-veis (approved):

PR #76, opened 2024-08-23
docs: add data view information and about select project dropdown gf-67

AnnaRomaniuta:


docs/specification.md
```diff
@@ -232,7 +232,11 @@ The "Analytics" page includes:
 - "Search” field: Allows the user to search for contributors by name.
 - “Project” select.
 - “Date” field: Allows the user to select a period.
+  The period for viewing data is limited to a maximum of 1 month. Users can select specific months rather than a custom date range.
+  The application does not allow selecting future dates for viewing data.
 - “Analytics” table: Displays selected dates as columns and contributors as rows. Each cell shows the number of commits made by the contributor on that date. If there are no commits, a minus sign is displayed.
+  The "Select project" dropdown list displays all projects with a scroll function if there are many. It should also allow users to search by typing the project name.
```
liza-veis, line 238: I think this should be moved above, to the `“Project” select` item and I believe there is no need to specify a scroll function (should be determined by design, not specification) or search during typing (sounds like a part of UX experience, not business logic).
```suggestion
  - “Project” select. Displays list of all projects.
```

liza-veis (changes requested):

what1s1ove (approved):

docs/specification.md
AnnaRomaniuta: I fixed in [commit](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/76/commits/26b5778783fdd39893870e97f2b8ece2d602dc66)

liza-veis (approved):

PR #77, opened 2024-08-23
feat: add ability to create project gf-57

perekotypole:
- added projects table
- implemented project model, repository, service etc.
- implemented the controller with POST /projects route
- added validation


what1s1ove (approved):
lgtm

yevhenii-hladkov (approved):

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		return item.toObject();
+	}
+
+	public findAll(): ReturnType<Service["findAll"]> {
+		return Promise.resolve({ items: [] });
+	}
+
+	public async getByName(name: string): Promise<ProjectCreateResponseDto> {
```
liza-veis, line 64: Looks like this method is not needed

liza-veis (changes requested):

liza-veis (approved):

PR #91, opened 2024-08-26
docs: requirements update 2.2 sidebar gf-62

VeronikaHordus:


docs/specification.md
```diff
@@ -25,11 +25,11 @@ The application is designed to collect and analyze developer activity in Git-bas
```
liza-veis, line 25: The only thing I would leave in this PR is redirecting to the main page on logo click. Other changes I would return because they are not related to the business logic and are not important for the application.

liza-veis (changes requested):

PR #95, opened 2024-08-26
feat: add user popup gf-58

Vitaliistf:
Added a user popup, displaying an authenticated user name and email in it.
On logout click, the authenticated user (store) and token (localStorage) are removed, and the user is redirected to the /sign-in page. When a user clicks on the avatar icon, a popup is shown. Popup is closed on click outside.
Besides main features, I created a pop-up element, which can be reused for notification pop-up, and added a hook for outside click. While reviewing PR's noticed that in #42 was created hook with the same functionality, so if it will be merged first, I'll remove mine.
Screenshot 2024-08-26 at 19 33 56


apps/frontend/src/libs/hooks/use-on-click-outside/use-on-click-outside.hook.ts
```diff
@@ -0,0 +1,24 @@
+import { type RefObject, useLayoutEffect } from "react";
```
yevhenii-hladkov, line 1:
```suggestion
import { useLayoutEffect } from "~/libs/hooks/hooks.js";
```

apps/frontend/src/libs/hooks/use-on-click-outside/use-on-click-outside.hook.ts
```diff
@@ -0,0 +1,24 @@
...
+
+type Parameters = {
+	callback: () => void;
+	ref: RefObject<HTMLElement>;
```
yevhenii-hladkov, line 5:
```suggestion
	ref: React.RefObject<HTMLElement>;
```

liza-veis:
Please attach a screenshot

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -1,27 +1,27 @@
...
 
-import { Avatar } from "../components.js";
+import { Avatar, Popup, UserPopup } from "../components.js";
 import styles from "./styles.module.css";
 
 const Header = (): JSX.Element => {
-	const authenticatedUser = useAppSelector(
-		({ auth }) => auth.authenticatedUser,
-	);
+	let authenticatedUser = useAppSelector(({ auth }) => auth.authenticatedUser);
```
yevhenii-hladkov, line 8:
```suggestion
	const authenticatedUser = useAppSelector(({ auth }) => auth.authenticatedUser);
```

perekotypole:
I think you should add more space between the popup and the child component.

apps/frontend/src/libs/components/user-popup/user-popup.tsx
```diff
@@ -0,0 +1,34 @@
...
+		<div className={styles["user-popup"]}>
+			<div className={styles["user-info"]}>
+				<p className={styles["user-name"]}>{name}</p>
+				<p className={styles["user-email"]}>{email}</p>
+			</div>
+			<div className={styles["buttons"]}>
+				<button className={styles["button"]}>Profile</button>
+				<button className={styles["button"]} onClick={handleLogout}>
+					Log out
+				</button>
```
perekotypole, line 28: Shouldn't we use the Button component?

Vitaliistf: The button here needs different styling and onClick prop. So Button component is not very suitable for that.

apps/frontend/src/libs/components/popup/popup.tsx
```diff
@@ -0,0 +1,60 @@
...
+import styles from "./styles.module.css";
+
+const ESCAPE_KEY = "Escape";
+
+type Properties = {
+	children: React.ReactNode;
+	content: React.ReactNode;
+};
+
+const Popup = ({ children, content }: Properties): JSX.Element => {
```
what1s1ove, line 17: we have a modal component for it

Vitaliistf: But they are not the same, modal is positioned in center of the screen, while pop-ups will be positioned relative to buttons that trigger their button. Also, modal has a backdrop and blocks interactions with everything, while pop-up does not. And as for modal used dialog element from html, we will not be able to use it as pop-up. I think it's the best to use separate components for modal and pop-up. However, I can reuse outside click hook from that PR, I'm just waiting for it to be merged.
Screenshot 2024-08-27 at 14 34 00
Screenshot 2024-08-27 at 14 41 29

liza-veis (changes requested):

apps/frontend/src/modules/auth/slices/actions.ts
```diff
@@ -54,4 +54,14 @@ const signUp = createAsyncThunk<
...
 });
 
-export { getAuthenticatedUser, signIn, signUp };
+const logout = createAsyncThunk<null, undefined, AsyncThunkConfig>(
+	`${sliceName}/logout`,
+	async (_payload, { extra }) => {
+		const { storage } = extra;
+		await storage.drop(StorageKey.TOKEN);
+
+		return null;
```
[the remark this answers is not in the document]
Vitaliistf: Yeah, I wanted to ask about that. I think I either have to return null here, or give that async arrow function return type of `Promise<undefined>`, because if I put void as first generic argument for `createAsyncThunk`, eslint throws the error: `void is only valid as a return type or generic type argument` which is strange, because here void is actually a generic type argument. I guess I should stick with second approach then?

apps/frontend/src/modules/auth/slices/auth.ts
```diff
@@ -1,9 +1,10 @@
-import { getAuthenticatedUser, signIn, signUp } from "./actions.js";
+import { getAuthenticatedUser, logout, signIn, signUp } from "./actions.js";
 import { actions } from "./auth.slice.js";
 
 const allActions = {
 	...actions,
 	getAuthenticatedUser,
+	logout,
```
AnnKasian, line 7: just a stylistic remark but if "Log out" contains 2 words, so maybe everywhere should be logOut, like in signIn and signUp?

Vitaliistf: Yeah I thought about it too, but almost always logout is written using a single word (you can google it). I think if we'd use "sign out", then that would be the case. It's not that hard to change it though, if we'll need to do it, I'll do it, but for now this way of naming looks better and more convenient to me.🙌

apps/frontend/src/libs/components/popover/popover.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const ESCAPE_KEY = "Escape";
+
+type Properties = {
+	children: React.ReactNode;
+	content: React.ReactNode;
+};
+
+const Popover = ({ children, content }: Properties): JSX.Element => {
+	const [isOpen, setIsOpen] = useState<boolean>(false);
```
GvoFor, line 18:
```suggestion
	const [isOpened, setIsOpened] = useState<boolean>(false);
```

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -13,15 +13,17 @@ const Header = (): JSX.Element => {
...
 	return (
 		<header className={styles["header"]}>
 			<div className={styles["logo-container"]}>
 				<img alt="GitFit logo" className={styles["logo-img"]} src={logoSrc} />
 				<span className={styles["logo-text"]}>Logo</span>
 			</div>
-			<Avatar name={userName} />
+			<Popover content={<UserPopover email={email} name={name} />}>
+				<Avatar name={name} />
+			</Popover>
```
GvoFor, line 26: This looks a bit ugly (`content` prop) and confusing(`UserPopover` as a content). 
Reading `UserPopover` I expected it to be `Popover` wrapper. So I expect something like:
```suggestion
			<UserPopover email={email} name={name}>
				<Avatar name={name} />
			</UserPopover >
```
where `UserPopover` will use `Popover` behind the scene

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -13,15 +13,17 @@ const Header = (): JSX.Element => {
...
+	const { email, name } = authenticatedUser;
 
 	return (
 		<header className={styles["header"]}>
 			<div className={styles["logo-container"]}>
 				<img alt="GitFit logo" className={styles["logo-img"]} src={logoSrc} />
 				<span className={styles["logo-text"]}>Logo</span>
 			</div>
-			<Avatar name={userName} />
+			<Popover content={<UserPopover email={email} name={name} />}>
```
what1s1ove, line 24: I am not sure if we need a separate component for the `UserPopover` or at least we do not need to put it in the `libs` since this is not a common component

let's remove this component or put it near the header component at least

apps/frontend/src/libs/components/popover/popover.tsx
Vitaliistf: I think both variants are ok, even `isOpen` might be better because "open" there is an adjective describing a state. "Opened" is more like a passive construction, and it describes the action/process of opening up. (ex. you say 'the shop is open right now', not 'the shop is opened right now'). But I see that in modal it was named like `isOpened`, so for consistency I'll do it.

apps/frontend/src/libs/components/header/header.tsx
Vitaliistf: Agree, I'll try to improve it.

apps/frontend/src/libs/components/header/header.tsx
Vitaliistf: Agree, I think I'll put it near the header, so we have styles and tsx of UserPopover in a separated from header files to not litter the header's code.

Vitaliistf: Done

apps/frontend/src/libs/components/header/libs/components/user-popover/user-popover.tsx
```diff
@@ -0,0 +1,42 @@
...
+	const dispatch = useAppDispatch();
+
+	const handleLogout = useCallback((): void => {
+		void dispatch(authActions.logout());
+	}, [dispatch]);
+
+	return (
+		<Popover
+			content={
+				<div className={styles["user-popup"]}>
```
GvoFor, line 23: It's not a big deal, but
```suggestion
				<div className={styles["user-popover"]}>
```

Vitaliistf: Oh, yeah, thanks, forgot about it)

yevhenii-hladkov (approved):

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -1,7 +1,8 @@
 import logoSrc from "~/assets/images/logo.svg";
+import { Avatar } from "~/libs/components/components.js";
 import { useAppSelector } from "~/libs/hooks/hooks.js";
 
-import { Avatar } from "../components.js";
+import { UserPopover } from "./libs/components/user-popover/user-popover.js";
```
what1s1ove, line 5:
```suggestion
import { UserPopover } from "./libs/components/components.js";
```

apps/frontend/src/libs/components/popover/popover.tsx
```diff
@@ -0,0 +1,57 @@
+import {
+	useCallback,
+	useHandleClickOutside,
+	useRef,
+	useState,
+} from "~/libs/hooks/hooks.js";
+
+import styles from "./styles.module.css";
+
+const ESCAPE_KEY = "Escape";
```
what1s1ove, line 10: create an enum for it instead

Vitaliistf: Made suggested changes

GvoFor (commented):
Looks good

what1s1ove (approved):

apps/frontend/src/libs/components/header/libs/components/user-popover/styles.module.css
```diff
@@ -0,0 +1,55 @@
...
+	cursor: pointer;
+	background-color: transparent;
+	border: none;
+	border-radius: 4px;
+}
+
+.button:hover {
+	background-color: color-mix(
+		in srgb,
+		var(--color-brand-primary-hover),
```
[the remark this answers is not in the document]
Vitaliistf: This is for hover effect. I didn't find anything about hover effect of this elements in design, however I thought that it would be good to have one, so made the same as in select component.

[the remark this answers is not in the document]
liza-veis: Let's use the same effect as for icon buttons / options menu on hover. In figma it's called background hover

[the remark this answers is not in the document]
Vitaliistf: Okay

perekotypole:
Could we hide the opened popover when clicking on avatar? It's not such a big problem, but it's not convenient either.

Vitaliistf:
> Could we hide the opened popover when clicking on avatar? It's not such a big problem, but it's not convenient either.

Oh, yeah, thanks, in ticket was stated that hiding popover should be triggered by click outside of it and I totally forgot that the avatar will be a child of the popover. I'll fix it.

liza-veis (approved):

PR #105, opened 2024-08-27
feat: Add projects page gf-72

AnnKasian:
Done: 
- Added a common page layout.
- Cleaned and updated projects page.
- Added projects get all api route.
- Added development ui page.

Preview:
image


apps/frontend/src/libs/helpers/get-development-routes/get-development-routes.helper.ts
```diff
@@ -0,0 +1,12 @@
+import { type RouteObject } from "react-router-dom";
```
liza-veis, line 1: I think this is unnecessary, a good approach overall, but we will remove this page as soon as possible, so I believe this helper is not needed. Just leave a todo comment on this item that it should be removed after components usage

liza-veis (changes requested):

apps/frontend/src/libs/components/page-layout/page-layout.tsx
```diff
@@ -0,0 +1,27 @@
+import { type ReactNode } from "react";
```
yevhenii-hladkov, line 1:
```suggestion
```

apps/frontend/src/libs/helpers/get-development-routes/get-development-routes.helper.ts
AnnKasian: Done

apps/frontend/src/libs/components/page-layout/styles.module.css
```diff
@@ -0,0 +1,44 @@
...
+	padding-left: var(--sidebar-width);
+	color: var(--color-text-primary);
+	background-color: var(--color-background-primary);
+}
+
+/* @media (--large-screen) {
+	.page-sidebar {
+		width: calc(var(--sidebar-width) - 80px);
+	}
+} */
```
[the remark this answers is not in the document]
AnnKasian: Done

apps/frontend/src/libs/components/page-layout/styles.module.css
```diff
@@ -0,0 +1,44 @@
+:root {
+	--header-height: 73px;
+	--sidebar-width: max(280px, 20%);
+}
+
+.page-sidebar {
+	position: fixed;
```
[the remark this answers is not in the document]
AnnKasian: I made both header and sidebar sticky. Also, I communicate about styles and page layout with @AnutkaGn. We resolving this unmatch.

apps/frontend/src/libs/components/page-layout/styles.module.css
```diff
@@ -0,0 +1,41 @@
+:root {
+	--header-height: 73px;
```
perekotypole, line 2: If --header-height is used only for the page and related components, it's better to keep it in .page scope, not in root

apps/frontend/src/pages/projects/components/project-card/project-card.tsx
perekotypole, line 1: Something has happened to the number of tab spaces. I'm not sure if it matters, but pay attention to it

AnnKasian: What do you mean?

perekotypole: Maybe I messed up, but in the second example there are obvious more tab space:
`apps/frontend/src/libs/components/page-layout/page-layout.tsx`
image
`apps/frontend/src/pages/projects/components/project-card/project-card.tsx`
image

apps/frontend/src/libs/components/page-layout/styles.module.css
AnnKasian: oh, right, changed it

apps/frontend/src/pages/projects/components/project-card/project-card.tsx
AnnKasian: image
image
maybe it's github show it incorrect? Locally i have everything right

packages/shared/src/modules/projects/libs/types/project-get-all-item-response-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+type ProjectGetAllItemResponseDto = {
```
GvoFor, line 1: I think better to use plural `-Items-` instead of singular `-Item-`
```suggestion
type ProjectGetAllItemsResponseDto = {
```
Do not forget about file name and other places it's used

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -1,118 +1,45 @@
...
-
-	const { control } = useAppForm({
-		defaultValues: {
-			options: [],
-		},
-	});
+	const { dataStatus, projects } = useAppSelector(({ projects }) => ({
+		dataStatus: projects.dataStatus,
+		projects: projects.projects,
+	}));
```
GvoFor, line 19: If I'm not wrong this can be simplified to 
```suggestion
	const { dataStatus, projects } = useAppSelector(({ projects }) => projects);
```
with no trade-off thanks to destructuring syntax. I mean, that if there will be more fields in `projects` slice, and some of them (except destructured ones) change, this component won't rerender, as it is in your implementation (if I'm not wrong)

packages/shared/src/modules/projects/libs/types/project-get-all-item-response-dto.type.ts
AnnKasian: 'getAll' gets many items. And in this dto we have 1 item from all items that we got. 
Literally "one item from method 'getAll' ".
Hope i explained clear

liza-veis: yes, we use Item here because it's just 1 item

apps/frontend/src/pages/projects/projects.tsx
liza-veis: agree

AnnKasian: simplified, thanks

apps/frontend/src/libs/components/page-layout/styles.module.css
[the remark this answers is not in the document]
AnnKasian: We decide to use my approach, she changed her code

packages/shared/src/modules/projects/libs/types/project-get-all-item-response-dto.type.ts
GvoFor: Oh, I didn't notice that. Then sure

apps/frontend/src/libs/components/page-layout/page-layout.tsx
```diff
@@ -0,0 +1,27 @@
+import { SIDEBAR_ITEMS } from "~/libs/constants/constants.js";
+
+import { Header } from "../header/header.js";
+import { Sidebar } from "../sidebar/sidebar.js";
```
Vitaliistf, line 4: I think we might need to prefer absolute imports here. 
```suggestion
import { Header, Sidebar } from "~/libs/components/components.js";
```

Vitaliistf: Yes it is, but I don't think it is a big deal here. Usually we use relative imports for cases if the file is in the same, or sub directory (so, strictly dependent). Common components should be standalone (kind of) and if they depend on other common ones we import them using absolute imports. (It's reliable, improves component maintainability, and is useful for example if we move component somewhere).

what1s1ove (approved):

apps/frontend/src/libs/components/page-layout/page-layout.tsx
liza-veis: yes, let's use absolute import here

AnnKasian: changed

liza-veis (approved):

Issue #87, opened 2024-08-25
feat: Add an ability to create a project


liza-veis:
### What feature?

- [x] Add a "Create New" button near the projects list.
- [x] On button click, the modal should be opened to create a new project.
- [x] After form submission, the new project should be added at the top of the project list.
- [x] Validation is the same as for the backend #57.

The backend part is added here #77.

### Screenshots

![Image](https://github.com/user-attachments/assets/471fc69f-b2c6-47cf-8047-ceab47bded62)

GvoFor:
A few questions:
1. Should I create new common component `TextArea` for description? or should I adjust `Input` component somehow?
2. Maximum description length is 1000 symbols (for now), and this means that there will be "text overflow". What is expected behaviour: should description height grow? should scrollbar appear on the right edge?

GvoFor:
Also I will need button `width` to be `150px` so I will adjust `Button` component. I'd like to do the following:

Add prop `variant` to `Button` which type will be `type ButtonVariant = "danger" | "default" | "normal" | "outlined";`
- "default" -- is the default value. It represent button, which width is flexible, i.e. its width fits min-content OR stretchs to 100% of container
![image](https://github.com/user-attachments/assets/8f6ef33f-8e45-4ca5-a900-37b314e567ce)
↑ stretched
![image](https://github.com/user-attachments/assets/7d10e817-ca07-4b58-a0e0-9b383154e668)
↑ min-content
- "normal" -- have the same styles as a default, but it's width is fixed to 150px
![image](https://github.com/user-attachments/assets/07d5607c-da6a-4338-ae78-ec9f4ffe2198)
![image](https://github.com/user-attachments/assets/3af9eb01-919f-48a8-b7d0-cc3818d21b94)
- "outlined" -- have fixed width (150px) and looks like this
![image](https://github.com/user-attachments/assets/483eae26-080c-4182-8220-1558633b2673)
- "danger" -- have fixed width (150px) and looks like this
![image](https://github.com/user-attachments/assets/bd5a537c-445f-4bf1-80c3-b13c67fcc46e)

Can I do it this way?

liza-veis:
> Should I create new common component TextArea for description? or should I adjust Input component somehow?

let's add rows property to input and in case we pass this property, then render textarea instead

> Maximum description length is 1000 symbols (for now), and this means that there will be "text overflow". What is expected behaviour: should description height grow? should scrollbar appear on the right edge?

[...]

PR #107, opened 2024-08-28
feat: add an ability to edit user name gf-89

perekotypole:
- added profile page with user info edit form
- disabled button and input components
- implemented ability to update user data in backend
- authenticated user data refreshes on updating
- implemented notification on successful data update

image


apps/frontend/src/libs/components/popup/popup.tsx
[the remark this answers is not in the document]
perekotypole: Yeah, I know... I used the old code, and after was waiting for something more stable

apps/frontend/src/libs/helpers/compare-objects/compare-objects.helper.ts
```diff
@@ -0,0 +1,37 @@
+const compareObjects = <T extends Record<string, unknown>>(
```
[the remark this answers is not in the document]
perekotypole: Thanks! I couldn't think of anything more appropriate than `compare`

apps/frontend/src/libs/helpers/compare-objects/compare-objects.helper.ts
```diff
@@ -0,0 +1,37 @@
...
+			return false;
+		}
+	}
+
+	return true;
+};
+
+const isPlainObject = (value: unknown): value is Record<string, unknown> => {
+	return typeof value === "object" && value !== null && !Array.isArray(value);
+};
```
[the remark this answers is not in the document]
perekotypole: You are right! I didn't find it in quality criteria, but it makes sense

apps/frontend/src/modules/users/slices/users.ts
```diff
@@ -1,9 +1,10 @@
-import { loadAll } from "./actions.js";
+import { loadAll, updateUser } from "./actions.js";
 import { actions } from "./users.slice.js";
 
 const allActions = {
 	...actions,
 	loadAll,
+	updateUser,
```
GvoFor, line 7: `updateUser` action will be used like `usersActions.updateUser(...)`, so it's better to rename it to `update`, so we do not violate A5 of QC

apps/frontend/src/pages/profile/components/edit-user-form/edit-user-form.tsx
```diff
@@ -0,0 +1,87 @@
...
+		useAppForm<UserInfoResponseDto>({
+			defaultValues,
+		});
+
+	const [isChanged, setIsChanged] = useState<boolean>(false);
+
+	const watchedValues = watch();
+
+	const handleFormSubmit = useCallback(
+		(event_: React.BaseSyntheticEvent): void => {
```
GvoFor, line 36: Why there is underscore `_` ?
```suggestion
		(event: React.BaseSyntheticEvent): void => {
```

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -0,0 +1,39 @@
...
+		<PageLayout>
+			<div className={styles["profile-content"]}>
+				<h1 className={styles["label"]}>Profile</h1>
+
+				{isLoading ? (
+					<div className={styles["profile-loader"]}>
+						<Loader />
+					</div>
+				) : (
+					authenticatedUser && (
```
GvoFor, line 24: It doesn't make sense to render the page if authenticatedUser is undefined/null. It's better to add
```ts
if (!authenticatedUser) {
  return <></>;
}
```
right after useAppSelector

BTW `ProtectedRoute` doesn't even allow non-authenticated user to visit this page

packages/shared/src/modules/users/libs/types/user-info-request-dto.type.ts
```diff
@@ -0,0 +1,5 @@
+import { type UserInfoResponseDto } from "./user-info-response-dto.type.js";
+
+type UserInfoRequestDto = UserInfoResponseDto;
```
GvoFor, line 3: I think it's better to duplicate type declaration and avoid this assignment in case types might change in the future (and from semantic perspective too)

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,24 +1,36 @@
...
+	className?: string;
+	disabled?: boolean;
 	href?: string | undefined;
 	label: string;
 	type?: "button" | "submit";
 };
 
-const Button = ({ href, label, type = "button" }: Properties): JSX.Element => {
+const Button = ({
+	className = "",
```
GvoFor, line 16: Adding `className` prop to common components is not good idea. In case you want to control the size of the button, you can do this with wrapper classes. Check this [comment](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/issues/87#issuecomment-2316856589). Also you can check my [PR](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/115) where I also modify this component and how I use it in `ProjectCreateForm`

apps/frontend/src/modules/users/slices/users.ts
perekotypole: Fair enough

apps/frontend/src/pages/profile/components/edit-user-form/edit-user-form.tsx
perekotypole: I took the example from sign in/up forms

packages/shared/src/modules/users/libs/types/user-info-request-dto.type.ts
perekotypole: I assumed that it would contain absolutely all user data (without passwords and id) and would be reused fot it wherever it was needed.

liza-veis: For request we only need name, so it should be different from response

apps/frontend/src/libs/components/button/button.tsx
perekotypole: got it

packages/shared/src/modules/users/libs/types/user-info-request-dto.type.ts
liza-veis, line 3: Rename to UserPatchRequestDto
```suggestion
type UserPatchRequestDto = {
  name: string;
};
```

packages/shared/src/modules/users/libs/types/user-info-response-dto.type.ts
```diff
@@ -0,0 +1,6 @@
+type UserInfoResponseDto = {
+	email: string;
```
liza-veis, line 2: we need to return id for entities
```suggestion
type UserPatchResponseDto = {
    id: number;
	email: string;
```

packages/shared/src/modules/users/libs/types/user-auth-response-dto.type.ts
```diff
@@ -1,8 +1,8 @@
+import { type UserInfoResponseDto } from "./user-info-response-dto.type.js";
+
 type UserAuthResponseDto = {
 	createdAt: string;
-	email: string;
 	id: number;
-	name: string;
-};
+} & UserInfoResponseDto;
```
liza-veis, line 6: We need to either pick all these properties from UserGetAllItemResponseDto, including createdAt and id or leave it as it was, without combining types

apps/frontend/src/pages/profile/components/edit-user-form/edit-user-form.tsx
```diff
@@ -0,0 +1,87 @@
...
+				void dispatch(authActions.getAuthenticatedUser());
+			})(event_);
+		},
+		[dispatch, handleSubmit, userId],
+	);
+
+	useEffect(() => {
+		const isEqual = areObjectsEqual(watchedValues, defaultValues);
+		setIsChanged(!isEqual);
+	}, [watchedValues, defaultValues]);
```
liza-veis, line 54: Aren't there properties like isTouched or isDirty or isFormChanged in react-hook-form?

apps/frontend/src/modules/users/users-api.ts
```diff
@@ -29,6 +33,23 @@ class UserApi extends BaseHTTPApi {
 
 		return await response.json<UserGetAllResponseDto>();
 	}
+
+	public async update(
```
liza-veis, line 37: we will use update for put
```suggestion
	public async patch(
```

liza-veis (changes requested):

apps/frontend/src/pages/profile/components/edit-user-form/edit-user-form.tsx
liza-veis: yep, we can't use event because we have an `event` global variable

apps/frontend/src/pages/profile/components/edit-user-form/edit-user-form.tsx
perekotypole: Oh, really! I completely missed it

packages/shared/src/modules/users/libs/types/user-auth-response-dto.type.ts
perekotypole: Okay, I get it. I'd rather leave it the way it was so as not to make a mess

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -13,6 +13,7 @@ import styles from "./styles.module.css";
 type Properties<T extends FieldValues> = {
 	autoComplete?: string;
 	control: Control<T, null>;
+	disabled?: boolean;
```
what1s1ove, line 16: b19

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -41,6 +46,18 @@ class UserController extends BaseController {
 			method: "GET",
 			path: UsersApiPath.ROOT,
 		});
+
+		this.addRoute({
+			handler: (options) =>
+				this.patch(
+					options as APIHandlerOptions<{
+						body: UserPatchResponseDto;
```
GvoFor, line 54:
```suggestion
						body: UserPatchRequestDto;
```

apps/frontend/src/modules/users/slices/actions.ts
```diff
@@ -15,4 +19,17 @@ const loadAll = createAsyncThunk<
 	return userApi.getAll();
 });
 
-export { loadAll };
+const updateProfile = createAsyncThunk<
+	UserPatchResponseDto,
+	{ id: number; userPayload: UserPatchResponseDto },
```
GvoFor, line 24:
```suggestion
	{ id: number; payload: UserPatchRequestDto },
```

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -0,0 +1,35 @@
...
+		return <></>;
+	}
+
+	const isLoading =
+		dataStatus === DataStatus.PENDING || dataStatus === DataStatus.IDLE;
+
+	return (
+		<PageLayout>
+			<div className={styles["profile-content"]}>
+				<h1 className={styles["label"]}>Profile</h1>
```
liza-veis, line 21: should be title, it will be renamed in other places in another PR
```suggestion
				<h1 className={styles["title"]}>Profile</h1>
```

liza-veis (changes requested):

what1s1ove (approved):

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -13,6 +13,11 @@
 	transition: background-color 0.3s ease;
 }
 
-.button:hover {
+.button:disabled {
+	cursor: unset;
```
liza-veis, line 17:
```suggestion
	cursor: not-allowed;
```

liza-veis (changes requested):

liza-veis (approved):

PR #108, opened 2024-08-28
feat: add an ability to create a project group gf-71

sandrvvu:
Add project_permissions table (columns: id, key, name, created_at, updated_at). Limit the name to 100 chars.

Add project_groups table (columns: id, key, name, created_at, updated_at). Limit the name to 100 chars.

Add projects_to_project_groups table (columns: id, project_group_id, project_id, created_at, updated_at).

Add users_to_project_groups table (columns: id, project_group_id, user_id, created_at, updated_at).

Add project_groups_to_project_permissions table (columns: id, project_group_id, project_permission_id, created_at, updated_at).

Add view_project permission (name: "View Project") through migration.

Add needed models, and services on the backend.

Add a controller for project_groups with POST /project-groups route.

The payload for group creation should contain name, project_id, user_ids, and permission_ids. Using this payload you will need to create an item in project_groups, then items in project_groups_to_permissions, and then in projects_to_project_groups.

The key should be automatically generated in the service by turning the name into snake_case.

The key should be unique for permissions and user_groups table (restriction on db level and check on the backend as well).

Add validation for request payload and path parameters.

Add new tables to the db schema in readme.

Example:
image

users_to_project_groups
image

projects_to_project_groups
image

project_groups_to_project_permissions
image

project_groups
image

image


what1s1ove (approved):

apps/backend/src/modules/project-groups/project-group.entity.ts
```diff
@@ -0,0 +1,99 @@
+import { type Entity } from "~/libs/types/types.js";
+
+import { type ProjectGroupCreateResponseDto } from "./libs/types/types.js";
+
+class ProjectGroupEntity implements Entity {
+	private id: null | number;
+	private name!: string;
+	private permissionIds!: { id: number }[];
+	private projectId!: { id: number };
+	private userIds!: { id: number }[];
```
kolibri753, line 10: You are using the similar approach as I was, I refactored it a little, maybe take a look at this [comment](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/56#discussion_r1735003814)

sandrvvu: I'll take a look, thanks

apps/backend/src/modules/project-groups/project-group.repository.ts
```diff
@@ -0,0 +1,70 @@
...
+			users: userIds,
+		};
+
+		const createdProjectGroup = await this.projectGroupModel
+			.query(trx)
+			.insertGraph(projectGroupData, { relate: true })
+			.returning("*")
+			.withGraphJoined("[permissions, projects, users]");
+
+		await trx.commit();
```
liza-veis, line 36: Hm, I think this syntax requires us to wrap everything into a try/catch and rollback the migration in case of error, doesn't it?

readme.md
```diff
@@ -34,7 +34,7 @@ erDiagram
       int avatar_file_id FK
    }
 
-   user_groups {
+ user_groups {
```
liza-veis, line 37: I think there shouldn't be a change in indentation

readme.md
```diff
@@ -66,6 +66,46 @@ erDiagram
       int permission_id FK
    }
 
+   user_groups_to_projects {
```
liza-veis, line 69: We don't need these newly added tables, they previously were removed 🤔 Instead we need to have a similar logic to user_groups, but instead user_groups we have project_groups and instead permissions we have project_permissions. We need to describe tables added in this pr

liza-veis (changes requested):

apps/backend/src/modules/project-groups/project-group.repository.ts
```diff
@@ -0,0 +1,82 @@
...
+
+	public constructor(projectGroupModel: typeof ProjectGroupModel) {
+		this.projectGroupModel = projectGroupModel;
+	}
+
+	public async create(entity: ProjectGroupEntity): Promise<ProjectGroupEntity> {
+		const { name, permissions, projectId, users } = entity.toNewObject();
+		const key = changeCase(name, "snakeCase");
+
+		const trx = await transaction.start(this.projectGroupModel.knex());
```
liza-veis, line 23: It looks like we have only one query here, do we need this transaction at all in this case 🤔

liza-veis (changes requested):

apps/backend/src/modules/project-groups/project-group.repository.ts
sandrvvu: But if an error occurs, for example if the user id does not exist, trx helps to return the database to the state it was in before the transaction started, because otherwise it would add the record with the error to the database

liza-veis: When we have only one query to db if it doesn't pass, it won't change anything in it. Transactions are only needed when we have multiple queries, for example, we create groups and then in another query we update user_groups. In this case these are two different queries, so we need to pass transaction to rollback the first query in case the second one fails. For single queries transactions do nothing. In our case I see we have only one query, don't we?

sandrvvu: okay,
how would you recommend to resolve the problem when the keys from payload are not present in another tables (like projectId in projects etc.)?

liza-veis: Nothing additional is needed if this is done in a single query (single unit). But I've just checked objection documentation and it says that insertGraph makes a separate query, so the transaction is recommended, so let's leave it.

liza-veis (approved):

PR #114, opened 2024-08-29
feat: Add search to the project list gf-85

AnnKasian:
Done:

- Added search to the project list.
- Updated projects get all module.

Preview:
image
image
image


apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -14,7 +14,7 @@ type Properties<T extends FieldValues> = {
 	autoComplete?: string;
 	control: Control<T, null>;
 	errors: FieldErrors<T>;
-	label: string;
+	label?: string;
```
GvoFor, line 17: Label must be required

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -47,7 +47,7 @@ const Input = <T extends FieldValues>({
 
 	return (
 		<label className={styles["input-label"]}>
-			<span className={styles["input-label-text"]}>{label}</span>
+			{label && <span className={styles["input-label-text"]}>{label}</span>}
```
GvoFor, line 50: As I mentioned above label must be present on the page anyway -- look A2 of quality criteria. Look how `Select` component implemented and use the same aproach

apps/frontend/src/libs/components/search/search.tsx
```diff
@@ -0,0 +1,40 @@
+import { Icon, Input } from "~/libs/components/components.js";
+import { debounce } from "~/libs/helpers/helpers.js";
+import { useAppForm, useEffect } from "~/libs/hooks/hooks.js";
+
+type Properties = {
+	onValueChange: (search: string) => void;
```
GvoFor, line 6: Not sure, but maybe it will be okay, if we just call this `onChange`

package.json
GvoFor, line 1: This dependency should be on frontend

packages/shared/src/libs/helpers/debounce/debounce.helper.ts
GvoFor, line 1: Do we need `debounce` on backend?

apps/frontend/src/libs/components/input/input.tsx
AnnKasian: I don't need label in my search field. And if I use an empty string the empty space is appeared and styles brokes) I didn't find a way to resolve it except for this one.

apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -35,8 +35,11 @@ class ProjectRepository implements Repository {
 		return item ? ProjectEntity.initialize(item) : null;
 	}
 
-	public async findAll(): Promise<ProjectEntity[]> {
-		const projects = await this.projectModel.query().execute();
+	public async findAll(name: string = ""): Promise<ProjectEntity[]> {
```
liza-veis, line 38: Let's rename it to findAllByName, also name should be required
```suggestion
	public async findAllByName(name: string): Promise<ProjectEntity[]> {
```

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -58,8 +59,10 @@ class ProjectService implements Service {
 		return item.toObject();
 	}
 
-	public async findAll(): Promise<ProjectGetAllResponseDto> {
-		const items = await this.projectRepository.findAll();
+	public async findAll(
```
liza-veis, line 62:
```suggestion
	public async findAllByName(
```

apps/frontend/src/libs/components/input/input.tsx
liza-veis: Agree, take a look at the approach with IconButton

apps/frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
```diff
@@ -22,6 +23,7 @@ type ReturnValue<T extends FieldValues = FieldValues> = {
 	control: Control<T, null>;
 	errors: FieldErrors<T>;
 	handleSubmit: UseFormHandleSubmit<T>;
+	watch: UseFormWatch<T>;
```
liza-veis, line 26: Hm, I saw in another place we used useWatch hook, let's keep only one approach

liza-veis (changes requested):

apps/frontend/src/libs/components/search/search.tsx
AnnKasian: renamed

apps/backend/src/modules/projects/project.repository.ts
AnnKasian: renamed

package.json
AnnKasian: moved to frontend

apps/frontend/src/libs/helpers/debounce/libs/constants.ts
```diff
@@ -0,0 +1,3 @@
+const DEBOUNCE_TIMEOUT = 300;
```
what1s1ove, line 1: we do not need it globally. it should be more specific and put it near the place where you will use it

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -1,30 +1,41 @@
...
+			void dispatch(projectActions.loadAll(value));
+		},
+		[dispatch],
+	);
+
 	const isLoading =
-		dataStatus === DataStatus.PENDING || dataStatus === DataStatus.IDLE;
+		dataStatus === DataStatus.IDLE ||
+		(dataStatus === DataStatus.PENDING &&
+			projects.length === EMPTY_ARRAY_LENGTH);
```
what1s1ove, line 33:
```suggestion
		(dataStatus === DataStatus.PENDING &&
			hasProject);
```
decompose such variables

apps/frontend/src/libs/components/input/input.tsx
AnnKasian: changed

apps/frontend/src/libs/constants/constants.ts
```diff
@@ -1 +1,3 @@
+export { DEBOUNCE_TIMEOUT } from "./debounce-timeout.js";
 export { SIDEBAR_ITEMS } from "./navigation-items.constant.js";
+export { EMPTY_ARRAY_LENGTH } from "@git-fit/shared";
```
GvoFor, line 3: Maybe `EMPTY_LENGTH` sounds more general? I used this name in my PR. However, it was not approved yet, so I'm not sure, how it should be called. Check the [comment](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/115#discussion_r1738431638)

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -14,18 +14,20 @@ type Properties<T extends FieldValues> = {
...
 	rightIcon?: JSX.Element;
-	type?: "email" | "password" | "text";
+	type?: "email" | "password" | "search" | "text";
 };
 
 const Input = <T extends FieldValues>({
 	autoComplete,
 	control,
 	errors,
+	isLabelVisible = true,
```
GvoFor, line 30: https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/51#discussion_r1732365570

packages/shared/package.json
GvoFor, line 1: Maybe I don't understand something, but it seems to me, that `debounce` will be used only for `Search` component (or in other place, but there user can interact with something). So that means that it'll be used only on frontend. So I don't think we need this library at shared package.

@liza-veis @what1s1ove what do you think?

AnnKasian: They said me to add it to shared

liza-veis: @GvoFor thank you for the question, we have discussed these with coaches and decided on this approach. The reason we decided so is that we decided on an approach where we use shared for all common helpers (even if they are currently used only for one app). So all the non-app-specific stuff like helpers for dates, validation, strings, arrays, etc. will be stored in shared.

apps/frontend/src/libs/constants/debounce-timeout.ts
liza-veis, line 1: file name should be debounce-timeout.constant.ts

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -1,30 +1,42 @@
...
+		},
+		[dispatch],
+	);
+
+	const hasProject = projects.length === EMPTY_ARRAY_LENGTH;
+
 	const isLoading =
-		dataStatus === DataStatus.IDLE || dataStatus === DataStatus.PENDING;
+		dataStatus === DataStatus.IDLE ||
+		(dataStatus === DataStatus.PENDING && hasProject);
```
liza-veis, line 34: Hm, why do we need to add hasProject here 🤔

packages/shared/src/libs/constants/constants.ts
liza-veis, line 1: should be in `libs/constants/empty-array-length.constant.ts` as in other places

liza-veis (changes requested):

apps/frontend/src/libs/constants/constants.ts
what1s1ove: agreed

apps/frontend/src/libs/constants/debounce-timeout.ts
what1s1ove: it isn't good to have such a constant. it's too generic. move it to near the place where you will use it and use a more specific name

apps/frontend/src/pages/projects/components/project-search/project-search.tsx
```diff
@@ -0,0 +1,26 @@
...
+	onChange: (search: string) => void;
+};
+
+const ProjectsSearch = ({ onChange }: Properties): JSX.Element => {
+	const handleSubmit = useCallback((event: React.FormEvent) => {
+		event.preventDefault();
+	}, []);
+
+	return (
+		<form className={styles["form-wrapper"]} onSubmit={handleSubmit}>
```
what1s1ove, line 16:
```suggestion
		<form className={styles["form-wrapper"]} onChange={handleSubmit}>
```

apps/frontend/src/pages/projects/components/project-search/project-search.tsx
```diff
@@ -0,0 +1,26 @@
...
+			<Search
+				isLabelVisible={false}
+				label="Projects search"
+				onChange={onChange}
```
what1s1ove, line 20:
```suggestion
```
the only way of working with form - use the useForm hook

apps/frontend/src/pages/projects/projects.tsx
AnnKasian: With fast Internet connection after user searched projects the loader appears instead of projects and creates "blink" effect. However with slow connection there's no loader at all for a couple of seconds and it doesn't feel as an user-friendly experience too. Should I always keep loader visible during pending state?

liza-veis: I'm just thinking about initial loading, when we have status pending but no projects yet

AnnKasian: Loader will be on initial load both with hasProjects and without it

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -44,10 +46,22 @@ const Input = <T extends FieldValues>({
...
+		hasRightIcon && styles["with-right-icon"],
+	);
 
 	return (
 		<label className={styles["input-label"]}>
-			<span className={styles["input-label-text"]}>{label}</span>
+			<span
+				className={
+					isLabelHidden ? "visually-hidden" : styles["input-label-text"]
+				}
```
liza-veis, line 61: "input-label-text" is a default style which should be always present
```suggestion
				className={
				  getValidClassNames(
				    styles["input-label-text"],
				    isLabelHidden && "visually-hidden")
				}
```

AnnKasian: Oh, yeah, changed

liza-veis (approved):

what1s1ove (approved):

Issue #57, opened 2024-08-22
feat: Add an ability to create a project


liza-veis:
### What feature?

- [x] Add `projects` table (columns: id, name, description created_at, updated_at). Limit the name to 50 chars and description to 1000 chars.
- [x] Add model, repository, and service on the backend.
- [x] Add a controller for projects with POST /projects route.
- [x] Project name should be unique (both in db and there should be a check for this on the backend).
- [x] Add validation for the request body.

AnnaRomaniuta:
Good job! Thanks!

PR #109, opened 2024-08-28
feat: Add project page gf-86

HoroshkoMykhailo:
Summary:

- Added a new route to handle navigation to the Project page.
- Developed a Project page component that displays the project title and description. 
- The project item in the project list was made to be a link to the project page.

image

image

PR #115, opened 2024-08-29
feat: add an ability to create a project gf-87

GvoFor:
## Main feature
This PR adds an ability to create new project in such a way:
1. There is `Create New` button on the `Projects` page.
image

2. When you press `Create New` button modal with creation form appears
image

3. Create button is disabled while `name` field is empty
4. When you fill-in name and description (and they are valid according to #57 ), and press `Create` button, new project appears on the top of the projects list if submission success, and toast notification apears on fail.

## Additional features
This PR also adjusts `Input` component, adding `rows` prop. When `rows` is present, and it's greater than 1 `Input` renders as `textarea` (instead of `input`), with exact `rows` lines in height.

Here is how it looks like with 4 rows (enouth space for text)
image

And the same but with overflowed text (not enouth space)
image

This PR also changes `Button` component, adding useful props and styles
image


apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -19,9 +19,12 @@ type Properties<T extends FieldValues> = {
 	name: FieldPath<T>;
 	placeholder?: string;
 	rightIcon?: JSX.Element;
+	rows?: number;
 	type?: "email" | "password" | "text";
 };
 
+const MIN_ROWS_TO_BE_TEXTAREA = 2;
```
AnnKasian, line 26: maybe it should be in separate file? like constants

GvoFor: I just didn't want to create that folders structure `libs/constants/...` just for this little constant... But your suggestion sounds right

GvoFor:
I have a few comments to clarify
1. When project successfully created it appears on top of the list. But when the page is refreshed (so the data refetched), the data is sorted in insertion order (the most recent are at the bottom). So should I modify `getAll` on backend so database query return sorted by `created_at` column (in order issue states) ?
2. When form data is filled correctly and `Create` button pressed, the modal closes, no matter creation operation success or fail (for example in case when project name is occupied). Should I close it only on fulfilled creation? (so if name is occupied, the modal didn't close and user have an ability to continue modifying the form)

liza-veis:
> When project successfully created it appears on top of the list. But when the page is refreshed (so the data refetched), the data is sorted in insertion order (the most recent are at the bottom). So should I modify getAll on backend so database query return sorted by created_at column (in order issue states) ?

yes, the most recent ones should be at the top

> When form data is filled correctly and Create button pressed, the modal closes, no matter creation operation success or fail (for example in case when project name is occupied). Should I close it only on fulfilled creation? (so if name is occupied, the modal didn't close and user have an ability to continue modifying the form)

If it's easy to implement, then the option with closing only on success sounds better to me. But if it's hard to implement, then the option with closing in any case is ok

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,24 +1,48 @@
 import { NavLink } from "react-router-dom";
 
+import { getValidClassNames } from "~/libs/helpers/helpers.js";
+
 import styles from "./styles.module.css";
 
+type ButtonVariant = "danger" | "default" | "outlined";
+
 type Properties = {
+	disabled?: boolean;
```
liza-veis, line 10:
```suggestion
	isDisabled?: boolean;
```

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -13,6 +13,28 @@
...
+}
+
+.button-outlined:hover:not(:disabled) {
+	background-color: var(--color-background-hover);
+}
+
+:disabled {
+	cursor: default;
+	opacity: 0.5;
+}
```
liza-veis, line 40: not sure if this selector matches anything 🤔 also let's use not allowed cursor instead of default

apps/frontend/src/libs/components/input/input.tsx
liza-veis: I think we don't need this constant at all. In case we pass any value to rows, we will render textarea, even if it's 1

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -60,15 +60,33 @@ const Input = <T extends FieldValues>({
...
+							styles["input-textarea"],
+						)}
+						name={field.name}
+						onChange={field.onChange}
+						placeholder={placeholder}
+						style={
+							{
+								"--rows": rows,
+							} as React.CSSProperties
+						}
```
liza-veis, line 76: Can't we just pass rows to textarea?
```suggestion
rows={rows}
```

liza-veis (changes requested):

apps/frontend/src/libs/components/button/styles.module.css
GvoFor: Oops, missed `.button` (however that works somehow)

apps/frontend/src/libs/components/input/input.tsx
GvoFor: Sure

GvoFor:
>If it's easy to implement, then the option with closing only on success sounds better to me. But if it's hard to implement, then the option with closing in any case is ok

One of the solution may be (I tested it, it's working)
```ts
dispatch(projectActions.create(payload))
  .then(({ meta }) => {
    if (meta.requestStatus === "fulfilled") {
      onModalClose();
    }
  }).catch(() => {});
```
But I'm not sure if using returned by dispatch promise (and then/catch) is okay. Is it okay?

Another may be to add additional "data status" field into State of the projects slice, representing creating operation status

What will be better?

---
I also noticed an issue. Currently projects reducer for `create` action looks like:
image
But here `dataStatus` is the same as for `getAll` action. And when I dispatch `create` action `dataStatus` updates, which leads to rerendering of whole projects list (even if submission fails, so projects list doesn't change).
I have a suggestion how to solve it, but not sure, if it's fine:
I need to rid of `dataStatus` changing, cause it's corresponds to `getAll` action and shouldn't be changed, so among those 3 cases (for `create` action), I have to left only fullfiled one:
```ts
builder.addCase(create.fulfilled, (state, action) => {
  state.projects = [action.payload, ...state.projects];
});
```
Is this solution fine?

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,24 +1,48 @@
...
+type ButtonVariant = "danger" | "default" | "outlined";
+
 type Properties = {
 	href?: string | undefined;
+	isDisabled?: boolean;
 	label: string;
+	onClick?: () => void;
 	type?: "button" | "submit";
+	variant?: ButtonVariant;
 };
```
kolibri753, line 16: What do you think about adding a width property to the Button component? For example, in my task, I need a modal with two buttons of the same width. Or is there a better way to achieve this?

GvoFor: Check this [comment](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/issues/87#issuecomment-2316084322) and its replies

GvoFor:
I guess I figure out these [questions](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/115#issuecomment-2318774270). Adding `projectCreateStatus` field to reducer state solved them and kept code consistent (`create` action has all 3 cases in reducer, as other actions do)

apps/frontend/src/libs/enums/success-message.enum.ts
```diff
@@ -0,0 +1,5 @@
+const SuccessMessage = {
```
liza-veis, line 1: Let's rename it to NotificationMessage. We will use it for toast notifications, whereas ExceptionMessage is for Error class notifications only
```suggestion
const NotificationMessage = {
```

apps/frontend/src/modules/projects/slices/project.slice.ts
```diff
@@ -4,30 +4,42 @@ import { DataStatus } from "~/libs/enums/enums.js";
...
+	projectCreateStatus: DataStatus.IDLE,
 	projects: [],
+	projectsStatus: DataStatus.IDLE,
 };
 
 const { actions, name, reducer } = createSlice({
 	extraReducers(builder) {
 		builder.addCase(loadAll.pending, (state) => {
-			state.dataStatus = DataStatus.PENDING;
+			state.projectsStatus = DataStatus.PENDING;
```
liza-veis, line 24: for get all let's leave it as dataStatus for now, to keep it the same as in other places
```suggestion
			state.dataStatus = DataStatus.PENDING;
```

apps/frontend/src/pages/projects/styles.module.css
```diff
@@ -1,7 +1,13 @@
-.label {
+.title {
 	margin: 0;
 }
 
+.projects-header {
+	display: flex;
+	justify-content: space-between;
+	min-height: 44px;
```
liza-veis, line 8: Do we need this? can't it just use it's children size?
```suggestion
```

liza-veis (changes requested):

apps/frontend/src/pages/projects/styles.module.css
GvoFor: I put it cause in figma height is exact 44px, but I have a bit less than 44px. Now I noticed, that the issue is with button component. It's height 42px, but according to Figma it should has `min-height` 44px.
So, sure, I'll rid of this `min-height` and add this style to button.

apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -36,7 +36,10 @@ class ProjectRepository implements Repository {
 	}
 
 	public async findAll(): Promise<ProjectEntity[]> {
-		const projects = await this.projectModel.query().execute();
+		const projects = await this.projectModel
+			.query()
+			.orderBy("created_at", "desc")
```
what1s1ove, line 41: add an enum `SortType`

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -19,6 +19,7 @@ type Properties<T extends FieldValues> = {
 	name: FieldPath<T>;
 	placeholder?: string;
 	rightIcon?: JSX.Element;
+	rows?: number;
```
what1s1ove, line 22:
```suggestion
	rowsCount?: number;
```

apps/frontend/src/pages/projects/components/project-create-form/libs/constants/description-rows-count.constant.ts
```diff
@@ -0,0 +1,3 @@
+const DESCRIPTION_ROWS_COUNT = 4;
```
what1s1ove, line 1: you can avoid this constant. just use `4` in the props

apps/frontend/src/pages/projects/components/project-create-form/libs/constants/empty-string-length.constant.ts
```diff
@@ -0,0 +1,3 @@
+const EMPTY_STRING_LENGTH = 0;
```
what1s1ove, line 1: It looks like a common constant. Let's use more generic name and move it to the shared package 

```suggestion
const EMPTY_LENGTH_COUNT = 0;
```

GvoFor: I'll call it `EMPTY_LENGTH`. Count sounds redundant

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
```diff
@@ -0,0 +1,68 @@
...
+		[enteredName],
+	);
+
+	const handleFormSubmit = useCallback(
+		(event: React.BaseSyntheticEvent): void => {
+			void handleSubmit((formData: ProjectCreateRequestDto) => {
+				onSubmit(formData);
+			})(event);
+		},
+		[handleSubmit, onSubmit],
```
GvoFor, line 42: Should I rename `event` to `event_`?

liza-veis: yes, better to keep it consistent

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
```diff
@@ -0,0 +1,68 @@
...
+			validationSchema: projectCreateValidationSchema,
+		},
+	);
+
+	const { name: enteredName } = useFormWatch({ control });
+
+	const isButtonDisabled = useMemo(
+		() => enteredName?.length === EMPTY_LENGTH,
+		[enteredName],
+	);
```
liza-veis, line 34: I think we don't need this, we just use usual validation approach for all forms. First time validation triggers on submit (the button is not disabled) and then it won't submit until validation is passed. This is one of the common approaches - not to trigger validation until user tries to submit the form. And sometimes user doesn't know which fields are required or what is wrong with the values and why the button is disabled. 
```suggestion
```

liza-veis (changes requested):

GvoFor:
I have noticed, that our `projects` table migration has a mistake: desctiption limit should be 1000, but it is 50.
image
I'm not sure, how correctly fix migrations. Let me tell my _assumption_, and say whether it's correct, or not.
**My assumption**: I should create new migration file, which will be the same, except the correct constant value (1000 instead of 50), and the name of the file will be the same except the first word (which will be the date, when I create the file). And it I think I also must remove old migration file (not sure).

liza-veis:
> I'm not sure, how correctly fix migrations. Let me tell my assumption, and say whether it's correct, or not.
My assumption: I should create new migration file, which will be the same, except the correct constant value (1000 instead of 50), and the name of the file will be the same except the first word (which will be the date, when I create the file). And it I think I also must remove old migration file (not sure).

We already fixed this in another PR, take a look at it https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/109/files#diff-d929b4557bf6f144110f92cc6bbccfdc470bb1ef222047ae496f034a740f6a5a

apps/frontend/src/pages/projects/components/project-create-form/styles.module.css
```diff
@@ -0,0 +1,12 @@
...
+	display: flex;
+	flex-direction: column;
+	gap: 16px;
+	align-items: flex-end;
+}
+
+.button-wrapper {
+	display: flex;
+	flex-direction: column;
+	min-width: 150px;
```
liza-veis, line 11: sounds like max-width will fit better here
```suggestion
	max-width: 150px;
```

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+	const handleFormSubmit = useCallback(
+		(event_: React.BaseSyntheticEvent): void => {
+			void handleSubmit((formData: ProjectCreateRequestDto) => {
+				onSubmit(formData);
+			})(event_);
+		},
+		[handleSubmit, onSubmit],
+	);
+
+	const isButtonDisabled = Boolean(errors.name) || Boolean(errors.description);
```
liza-veis, line 32: I meant that we don't need to disable button because of validation errors at all. User should be able to click on it, so it will trigger the validation again and appropriate messages will be displayed
```suggestion
```

liza-veis (changes requested):

apps/frontend/src/pages/projects/components/project-create-form/styles.module.css
GvoFor: Button wrapper is placed inside flex container, that forces shrinking to min-content. So this should be `min-width` in order to prevent flex from shrinking the container to have width less than 150px (so flex shrinks it exact to 150px). If I use `max-width` instead, the width of the button (as the width of the wraper itself) will be less than 150px;
image
↑ with `max-width: 150px`
image
↑ with `min-width: 150px`

GvoFor: So I left `min-width`

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
GvoFor: Oh, get it.
I didn't get that first time ;) The thing that confused me is that in Figma the button was disabled
image

Should I remove `isDisabled` prop of `Button` component? Not sure, if we will need this prop, if we have field errors instead.
P.s. I will remove it for now

apps/backend/src/libs/enums/sort-type.enum.ts
```diff
@@ -0,0 +1,6 @@
+const SortType = {
```
what1s1ove, line 1: it's a generic think. let's move to the shared module

apps/frontend/src/libs/components/button/button.tsx
```diff
@@ -1,24 +1,41 @@
...
 import styles from "./styles.module.css";
 
+type ButtonVariant = "danger" | "default" | "outlined";
+
 type Properties = {
 	href?: string | undefined;
 	label: string;
+	onClick?: () => void;
 	type?: "button" | "submit";
+	variant?: ButtonVariant;
```
what1s1ove, line 14: inline values here

what1s1ove (approved):

liza-veis (approved):

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
liza-veis: We will need disabled state, for example, for profile form, we have different logic. And there we will disable button until nothing changed. But it will be added here #107

apps/backend/src/libs/enums/enums.ts
```diff
@@ -1,4 +1,4 @@
-export { SortType } from "./sort-type.enum.js";
+export { SortType } from "@git-fit/shared";
```
liza-veis, line 1: Move it to the export below

liza-veis (approved):

PR #116, opened 2024-08-29
fix: refactor user groups to permissions relation gf-78

s1rserg:
Created new migration to change FK user_groups_to_permissions.permission_id to reference permissions.id


apps/backend/src/db/migrations/20240829181900_fix_user_groups_to_permissions.ts
```diff
@@ -0,0 +1,44 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "user_groups_to_permissions";
+
+const ColumnName = {
```
liza-veis, line 5: Leave only needed columns in enum here

liza-veis (changes requested):

liza-veis (approved):

PR #122, opened 2024-08-29
feat: add an ability to view groups gf-78

s1rserg:
* Added groups table on Access Management page that displays Name, Permissions, and Created At columns.
* Added backend route for getting all groups.

image


apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -4,29 +4,39 @@ import {
...
+	const userColumns = getUserColumns();
+	const userData: UserRow[] = getUserRows(users);
+
+	const groupColumns = getGroupColumns();
+	const groupData: GroupRow[] = getGroupRows(groups);
 
 	return (
 		<PageLayout>
 			<p className={styles["title"]}>Access Management</p>
 			<p className={styles["sub-title"]}>Users</p>
```
liza-veis, line 36: Haven't noticed this previously, let's use appropriate tags and wrap each section with `section`
```suggestion
			<h1 className={styles["title"]}>Access Management</p>
			<section>
			  <h2 className={styles["section-title"]}>Users</p>
			  //...table
			<section>
```

packages/shared/src/modules/groups/libs/types/group-get-all-item-response-dto.type.ts
```diff
@@ -0,0 +1,9 @@
+type GroupGetAllItemResponseDto = {
+	createdAt: string;
+	id: number;
+	name: string;
+	permissions: { id: number; name?: string }[];
```
liza-veis, line 5: Why name is optional 🤔 
```suggestion
	permissions: { id: number; name: string }[];
```

liza-veis (changes requested):

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -4,29 +4,39 @@ import {
 	useAppSelector,
 	useEffect,
 } from "~/libs/hooks/hooks.js";
+import { actions as groupActions } from "~/modules/groups/groups.js";
 import { actions as userActions } from "~/modules/users/users.js";
 
+import { getGroupColumns } from "./libs/helpers/get-group-columns.helper.js";
+import { getGroupRows } from "./libs/helpers/get-group-rows.helper.js";
 import { getUserColumns } from "./libs/helpers/get-user-columns.helper.js";
 import { getUserRows } from "./libs/helpers/helpers.js";
```
AnutkaGn, line 13: Import all helpers into a file helpers.ts and import everything from one place
```suggestion
import { getUserRows, getUserColumns, getGroupColumns, getGroupRows } from "./libs/helpers/helpers.js";
```

what1s1ove (approved):
Looking good after Lisa's PR changes

yevhenii-hladkov (approved):

apps/backend/src/modules/groups/group.entity.ts
```diff
@@ -1,49 +1,52 @@
...
+	type GroupCreateResponseDto,
+	type GroupGetAllItemResponseDto,
+} from "./libs/types/types.js";
 
 class GroupEntity implements Entity {
+	private createdAt: null | string;
 	private id: null | number;
 	private name!: string;
-	private permissions!: Pick<PermissionModel, "id">[];
+	private permissions!: Array<{ id: number; name?: string }>;
```
liza-veis, line 13: Why did you change it this way instead of `Pick<PermissionModel, "id" | "name">[]`?

apps/backend/src/modules/groups/group.entity.ts
```diff
@@ -52,34 +55,49 @@ class GroupEntity implements Entity {
...
 		return new GroupEntity({
+			createdAt: null,
 			id: null,
 			name,
 			permissions,
 			users,
 		});
 	}
 
+	public toGetAllItemObject(): GroupGetAllItemResponseDto {
```
liza-veis, line 70: We need to modify toObject instead, in our case we will need to have the same entity fields for get all as for create, because after creation we are going to insert returned item to the store
```suggestion
```

liza-veis (changes requested):

apps/backend/src/modules/groups/group.entity.ts
s1rserg: I needed to somehow state that name can be optional, since we don't receive it in payload and I can't provide those names to `initializeNew` in `group.service`.  I am now using `Array<Partial<Pick<PermissionModel, "name">> & Pick<PermissionModel, "id">>`.  Can't really find more elegant solution for this.

liza-veis (approved):

PR #123, opened 2024-08-30
fix: Make sidebar item to indicate the active page gf-118

HoroshkoMykhailo:
Summary

- The active page is highlighted on the sidebar when the page is opened.
- When hovering over sidebar item the background changes to gray.

image

image


apps/frontend/src/libs/components/sidebar/sidebar-item/sidebar-item.tsx
```diff
@@ -1,19 +1,29 @@
...
-			<label className={styles["navigation-item-label"]}>{label}</label>
-		</NavLink>
-	</li>
-);
+const SidebarItem = ({ href, icon, label }: NavigationItem): JSX.Element => {
+	const location = useLocation();
+	const currentPath = location.pathname;
+
+	const isActive = currentPath === href;
+
```
liza-veis, line 13: As I remember NavLink already has isActive flag when adding styles
```suggestion
```

liza-veis (changes requested):

apps/frontend/src/libs/components/sidebar/sidebar-item/helper/get-class-name-nav.helper.ts
```diff
@@ -0,0 +1,7 @@
+import styles from "../styles.module.css";
+
+const getClassName = ({ isActive }: { isActive: boolean }): string => {
```
liza-veis, line 3: let's keep this helper inside SidebarItem component, it most likely will depend on other props later, and usually we keep className related logic inside components

liza-veis (changes requested):

apps/frontend/src/libs/components/sidebar/sidebar-item/sidebar-item.tsx
```diff
@@ -5,15 +5,21 @@ import { type NavigationItem } from "~/libs/types/types.js";
...
-	</li>
-);
+const getClassName = ({ isActive }: { isActive: boolean }): string => {
+	return `${styles["navigation-link"] ?? ""} ${isActive ? (styles["active"] ?? "") : ""}`;
+};
+
+const SidebarItem = ({ href, icon, label }: NavigationItem): JSX.Element => {
+	return (
+		<li className={styles["sidebar-item"]}>
+			<NavLink className={getClassName} to={href}>
```
what1s1ove, line 15: Remove all `import { NavLink } from "react-router-dom";` from the project
we need to use our `Link` component wrapper. You can use `NavLink` from `react-router-dom` in our Link component

HoroshkoMykhailo: But how then change style of Link component? It was said that we cannot pass classname to Link component

apps/frontend/src/libs/components/sidebar/sidebar-item/sidebar-item.tsx
```diff
@@ -5,15 +5,21 @@ import { type NavigationItem } from "~/libs/types/types.js";
...
-			<span className={styles["navigation-icon"]}>
-				<Icon height={20} name={icon} width={20} />
-			</span>
-			<label className={styles["navigation-item-label"]}>{label}</label>
-		</NavLink>
-	</li>
-);
+const getClassName = ({ isActive }: { isActive: boolean }): string => {
+	return `${styles["navigation-link"] ?? ""} ${isActive ? (styles["active"] ?? "") : ""}`;
+};
```
perekotypole, line 10: Might it better to use `get-valid-classes` helper?

apps/frontend/src/libs/components/sidebar/sidebar-item/sidebar-item.tsx
liza-veis: Let's leave it as it is in this PR, will create a separate ticket when we decide on approach

liza-veis (changes requested):

liza-veis (approved):

yevhenii-hladkov (approved):

PR #126, opened 2024-08-30
fix: Inconsistent layout on the Projects and Access Managment page gf-120

AnutkaGn:
Changed the styles on the Access Management page
Changed styles in the PageLayout component
Changed the styles on the Project page for the title of this page

Projects page
image

Access Management page
image

### DIFF
--- a/apps/frontend/src/pages/projects/styles.module.css
+++ b/apps/frontend/src/pages/projects/styles.module.css
@@ -1,5 +1,8 @@
 .label {
 	margin: 0;
+	font-size: 30px;
+	font-weight: 600;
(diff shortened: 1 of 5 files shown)


apps/frontend/src/pages/projects/styles.module.css
```diff
@@ -1,5 +1,8 @@
 .label {
```
liza-veis, line 1: class names should also be consistent
```suggestion
.title {
```

liza-veis (changes requested):

liza-veis (approved):

PR #127, opened 2024-08-30
fix: Loader is missing when switching between pages gf-119

GvoFor:
## Changes
- fix access management route location, so there is no white screen, when switching pages
- fix loader styles so it's displayed correctly
- add loader to access management page

## Screenshots
image
image


apps/frontend/src/libs/helpers/is-data-loading.helper.ts
```diff
@@ -0,0 +1,7 @@
+import { DataStatus } from "~/libs/enums/enums.js";
+import { type ValueOf } from "~/libs/types/types.js";
+
+const isDataLoading = (dataStatus: ValueOf<typeof DataStatus>): boolean =>
```
liza-veis, line 4: I would prefer not to add this helper, because sometimes we only need to check for idle or pending state, so this may vary too much

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -22,11 +23,19 @@ const AccessManagement = (): JSX.Element => {
...
 	const data: UserRow[] = getUserRows(users);
 
+	const isLoading = isDataLoading(dataStatus);
+
 	return (
 		<PageLayout>
 			<p className={styles["title"]}>Access Management</p>
-			<p className={styles["sub-title"]}>Users</p>
-			<Table<UserRow> columns={columns} data={data} />
+			{isLoading ? (
```
liza-veis, line 31: Let's move loader inside PageLayout and pass isLoading as parameter

liza-veis (changes requested):

apps/frontend/src/pages/access-management/access-management.tsx
GvoFor: Make sense

liza-veis (approved):

what1s1ove (approved):

yevhenii-hladkov (approved):

PR #128, opened 2024-08-30
feat: add an ability to delete a group gf-80

kolibri753:
1. Options Icon Addition:
image
Added an options icon as a new column at the end of each row in the Groups table.

2.  Options Group Menu:
image
Implemented an options group menu that appears on clicking the icon. Added a new popover hook, usePosition, to calculate where the menu should appear. This hook is optional and ensures that the menu’s appearance remains consistent, even in projects with absolute positioning.

3. Confirmation Modal:
image
Added a confirmation modal that appears when the delete option is selected. The modal's styling was adjusted to match the Figma design specifications.

4. Group Deletion:
image
image
image

Implemented backend functionality to delete a group upon confirmation. Updated the table and adjusted pagination to reflect changes when a group is deleted.


apps/backend/src/modules/groups/group.controller.ts
```diff
@@ -115,6 +124,36 @@ class GroupController extends BaseController {
...
+	 *         description: Group deleted successfully
+	 *       404:
+	 *         description: Group not found
+	 */
+
+	private async deleteGroup(
+		options: APIHandlerOptions<{ params: { id: number } }>,
+	): Promise<APIHandlerResponse> {
+		const { id } = options.params;
+		await this.groupService.delete(Number(id));
```
yevhenii-hladkov, line 150: Since `id` is already typed as a number, there's no need to convert it to a Number.

apps/backend/src/modules/groups/group.controller.ts
```diff
@@ -115,6 +124,36 @@ class GroupController extends BaseController {
...
+
+		return {
+			payload: null,
```
yevhenii-hladkov, line 153:
```suggestion
			payload: await this.groupService.delete(id),
```

yevhenii-hladkov: Actually, you can leave it as is. I missed something

kolibri753: It's just a draft, I haven't thought much into it yet, thanks for advices :cat2:

apps/backend/src/modules/groups/group.controller.ts
```diff
@@ -61,6 +61,15 @@ class GroupController extends BaseController {
 			},
 		});
 
+		this.addRoute({
+			handler: (options) =>
+				this.deleteGroup(
+					options as APIHandlerOptions<{ params: { id: number } }>,
+				),
+			method: "POST",
```
GvoFor, line 69: The method should be "DELETE"

kolibri753: thanks, I've already fixed it, just haven't pushed yet (electricity)

apps/backend/src/modules/groups/group.service.ts
```diff
@@ -46,8 +46,17 @@ class GroupService implements Service {
...
-		return Promise.resolve(true);
+	public async delete(id: number): Promise<boolean> {
+		const group = await this.groupRepository.find(id);
+
+		if (!group) {
+			throw new GroupError({
+				message: ExceptionMessage.GROUP_NOT_FOUND,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
```
GvoFor, line 57: Do we really need this check? Maybe lets just return `true` if group was deleted and `false` otherwise?

GvoFor: At least, you can avoid `find` request, and call `delete` right away and then check if result is `false` (item was not deleted), that means that there is no item with this id and we can throw error

apps/frontend/src/pages/access-management/libs/components/group-popover/group-popover.tsx
```diff
@@ -0,0 +1,67 @@
...
+									Edit
+								</button>
+								<button
+									className={styles["button"]}
+									onClick={handleDeleteClick}
+								>
+									<Icon height={17} name="trash" width={17} />
+									Delete
+								</button>
+							</div>
```
kolibri753, line 47: Menu component?

apps/frontend/src/assets/images/icons/pen.svg
GvoFor, line 1: Let's call this icon `pencil`

apps/frontend/src/assets/images/icons/trash.svg
GvoFor, line 1: Let's call this icon `trash-can`

apps/frontend/src/assets/images/icons/pen.svg
kolibri753: not sure about this, I see that some of us used "edit"

kolibri753: I was thinking of waiting until the first one is merged, and then redoing it to match. Or would it be better if we discussed this first?

apps/backend/src/modules/groups/group.service.ts
```diff
@@ -46,8 +46,17 @@ class GroupService implements Service {
...
+		const isDeleted = await this.groupRepository.delete(id);
+
+		if (!isDeleted) {
+			throw new GroupError({
+				message: ExceptionMessage.GROUP_NOT_FOUND,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		return isDeleted;
```
kolibri753, line 59: Maybe I should just return await this.groupRepository.delete(id); instead of this.

apps/frontend/src/modules/groups/slices/actions.ts
```diff
@@ -15,4 +16,16 @@ const loadAll = createAsyncThunk<
 	return await groupApi.getAll();
 });
 
-export { loadAll };
+const deleteById = createAsyncThunk<number, number, AsyncThunkConfig>(
```
AnnKasian, line 19: Maybe just `delete`? We  don't have another version of deleting, just by Id

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -37,12 +37,17 @@ class GroupRepository implements Repository {
 		return GroupEntity.initialize(group);
 	}
 
-	public delete(): ReturnType<Repository["delete"]> {
-		return Promise.resolve(true);
+	public async delete(id: number): Promise<boolean> {
+		const ROWS_DELETED = 0;
```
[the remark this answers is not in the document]
kolibri753: Yeah, I thought about it and even checked other PRs later, and found that they [exported](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/168#discussion_r1742173332) it. So I'll change it.

apps/frontend/src/modules/groups/slices/actions.ts
kolibri753: If shortly, it's a reserved keyword

apps/frontend/src/assets/images/icons/pen.svg
GvoFor: >I see that some of us used "edit"

I asked her to rename it to `pencil` too (she already done this)
I also asked her to rename `options` to `ellipsis`

The point (in my opinion) is that icon name should represent its content precisely. For example, from `edit` name can you imagine, how the icon looks like? or from `options` name?

apps/backend/src/modules/groups/group.service.ts
GvoFor: I clarified this and you should keep this as it is. Check [this](https://bsa2024.slack.com/archives/C07HWM8616U/p1725378501043329).

apps/frontend/src/assets/images/icons/pen.svg
kolibri753: It can be the opposite too. Sometimes we know what the icon represents (like edit or delete), but the design isn’t final. Using a generic name allows us to swap the icon without renaming everything.

GvoFor: Hm. That's a good point. Then lets @liza-veis decide

GvoFor: By the way, our design is final (at least this right for icons)

kolibri753: I just wanted to tell the opposite opinion 😅

kolibri753:
Working on fixing that

1. with position relative on options container
The popup is displayed correctly, even with a scrollbar
image

however, the last popup gets rendered inside the table
image

2. without position relative on options container
The popup is displayed correctly when there is no scrollbar

But, when there are many elements and the table overflows, the popup does not adjust properly.
image

Fjortis:
> Working on fixing that
> 
> 1. with position relative on options container
>    The popup is displayed correctly, even with a scrollbar
>    image
> 
> however, the last popup gets rendered inside the table image
> 
> 2. without position relative on options container
>    The popup is displayed correctly when there is no scrollbar
> 
> But, when there are many elements and the table overflows, the popup does not adjust properly. image
what i think of is to try to use position: absolute with a higher level container, or use position absolutefor popup and dynmically adjust positioning.

apps/frontend/src/libs/components/confirmation-modal/confirmation-modal.tsx
```diff
@@ -0,0 +1,45 @@
...
+const ConfirmationModal = ({
+	cancelLabel = "Cancel",
+	confirmationText,
+	confirmLabel = "Confirm",
+	isModalOpened,
+	onConfirm,
+	onModalClose,
+	title,
+}: ConfirmationModalProperties): JSX.Element => {
+	const handleConfirmClick = useCallback(() => {
```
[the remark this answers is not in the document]
perekotypole: We can't use simple functions for events

apps/backend/src/modules/groups/group.controller.ts
```diff
@@ -122,6 +131,36 @@ class GroupController extends BaseController {
...
+	private async deleteGroup(
+		options: APIHandlerOptions<{ params: { id: string } }>,
+	): Promise<APIHandlerResponse> {
+		const { id } = options.params;
+		const isDeleted = await this.groupService.delete(Number(id));
+
+		return {
+			payload: isDeleted,
+			status: HTTPCode.NO_CONTENT,
+		};
```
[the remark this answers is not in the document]
kolibri753: *status: HTTPCode.OK,

kolibri753:
Should we freeze the table from scrolling when the popover menu is opened?

apps/frontend/src/libs/components/menu/menu.tsx
```diff
@@ -7,13 +7,15 @@ type Properties = {
 	isOpened: boolean;
 	onClose: () => void;
 	onOpen: () => void;
+	usePositioning?: boolean;
```
what1s1ove, line 10:
```suggestion
	usePositioning?: boolean;
```
`use` is a better prefix for functions. use `has` instead

apps/frontend/src/libs/components/popover/popover.tsx
```diff
@@ -7,23 +11,46 @@ type Properties = {
 	content: React.ReactNode;
 	isOpened: boolean;
 	onClose: () => void;
+	usePositioning?: boolean;
```
what1s1ove, line 14: same

what1s1ove (approved):

apps/frontend/src/libs/components/menu/menu.tsx
kolibri753: Maybe `enablePositioning` is better than `hasPositioning`? Has sounds like it's already active, while enable makes it clear we’re deciding to turn it on or not.

apps/frontend/src/pages/access-management/libs/components/groups-table/groups-table.tsx
```diff
@@ -0,0 +1,112 @@
...
+	onPageChange,
+	onPageSizeChange,
+	page,
+	pageSize,
+	totalItemsCount,
+}: Properties): JSX.Element => {
+	const dispatch = useAppDispatch();
+
+	const { isOpened, onClose, onOpen } = useModal();
+	const [groupToDelete, setGroupToDelete] =
```
liza-veis, line 44: It looks like we only need id, not the whole group

apps/frontend/src/libs/hooks/use-popover-position/use-popover-position.hook.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type Properties = {
+	left?: number;
+	top?: number;
+};
+
+const usePopoverPosition = <T extends HTMLElement>(
+	reference: React.RefObject<T>,
+	isActive: boolean,
+	usePositioning: boolean = false,
```
liza-veis, line 11: do we need this flag at all? can't we use the same logic for all popups?
```suggestion
```

apps/frontend/src/libs/hooks/use-popover-position/use-popover-position.hook.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from "~/libs/hooks/hooks.js";
+
+type Properties = {
+	left?: number;
+	top?: number;
+};
+
+const usePopoverPosition = <T extends HTMLElement>(
+	reference: React.RefObject<T>,
+	isActive: boolean,
```
liza-veis, line 10: What is isActive? if it means if popup is open, then let's name it isOpen

apps/frontend/src/libs/hooks/use-popover-position/use-popover-position.hook.ts
```diff
@@ -0,0 +1,34 @@
...
+
+	useEffect(() => {
+		if (!usePositioning || !isActive || !reference.current) {
+			setPosition({});
+
+			return;
+		}
+
+		const { left, top } = reference.current.getBoundingClientRect();
+		const scrollOffset = window.scrollY || document.documentElement.scrollTop;
```
liza-veis, line 23: Can't we use just one of these values?

apps/frontend/src/modules/groups/slices/group.slice.ts
```diff
@@ -32,6 +32,13 @@ const { actions, name, reducer } = createSlice({
 			state.groups = [];
 			state.dataStatus = DataStatus.REJECTED;
 		});
+		builder.addCase(deleteById.fulfilled, (state, action) => {
+			const deletedGroupId = action.meta.arg;
+			state.groups = state.groups.filter(
+				(group) => group.id !== deletedGroupId,
+			);
+			state.groupsTotalCount -= 1;
```
liza-veis, line 40:
```suggestion
			state.groupsTotalCount -= GROUPS_DELETE_COUNT;
```

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -48,8 +55,19 @@ const Table = <T extends object>({
 						table.getRowModel().rows.map((row) => (
 							<tr className={styles["table-row"]} key={row.id}>
 								{row.getVisibleCells().map((cell) => (
-									<td className={styles["table-data"]} key={cell.id}>
-										{flexRender(cell.column.columnDef.cell, cell.getContext())}
+									<td
+										className={getValidClassNames(
+											styles["table-data"],
+											cell.column.id === "options" && styles["options-column"],
```
liza-veis, line 61: no, we need to do this differently. The table component should now nothing about options and what if we will have more than just one options column? One of the possible solutions to pass width as an optional parameter to column config
```suggestion
```

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -48,8 +55,19 @@ const Table = <T extends object>({
...
+										)}
+										key={cell.id}
+									>
+										{typeof cell.getValue() === "object"
+											? (cell.getValue() as React.ReactNode)
+											: flexRender(
+													cell.column.columnDef.cell,
+													cell.getContext(),
```
liza-veis, line 69: What is this for?

liza-veis (changes requested):

apps/frontend/src/libs/components/menu/menu.tsx
what1s1ove: prefer `has`/`is`

apps/frontend/src/libs/components/table/table.tsx
perekotypole: `flexRender` doesn't display react components, so we have to check what the content of cell

apps/frontend/src/libs/components/icon-button/styles.module.css
```diff
@@ -1,4 +1,5 @@
 .icon-button {
+	z-index: 1;
```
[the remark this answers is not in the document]
kolibri753: without it we are not able to open popover menu on project

apps/frontend/src/libs/hooks/use-popover-position/use-popover-position.hook.ts
kolibri753: we use absolute/relative positioning in projects module. If we don't use it, we are allowed to use this hook, but the full width of the project won't be clickable

apps/frontend/src/pages/access-management/libs/components/groups-table/groups-table.tsx
```diff
@@ -0,0 +1,112 @@
...
+		useState<GroupGetAllItemResponseDto | null>(null);
+
+	const groupColumns = getGroupColumns({
+		onDelete: (groupId: number) => {
+			const group = groups.find(({ id }) => id === groupId);
+			setGroupToDelete(group ?? null);
+
+			onOpen();
+		},
```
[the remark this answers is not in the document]
kolibri753: not sure I understand correctly, smth like this?
```
const groupColumns = getGroupColumns({
		onDelete: (groupId: number) => {
			handleDeleteClick({
				groupId,
				groups,
				onOpen: onDeleteModalOpen,
				setGroupIdToDelete,
			});
		},
	});
```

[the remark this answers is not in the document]
kolibri753: or just
```
	const onDelete = useCallback(
		(groupId: number) => {
			const groupExists = groups.some(({ id }) => id === groupId);

			if (groupExists) {
				setGroupIdToDelete(groupId);
				onDeleteModalOpen();
			}
		},
		[groups, onDeleteModalOpen],
	);

	const groupColumns = getGroupColumns(onDelete);
```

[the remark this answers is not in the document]
kolibri753: I think it's better to delete unnecessary checks and just leave this
```
const onDelete = useCallback(
	(groupId: number) => {
		setGroupId(groupId);
		onDeleteModalOpen();
	},
	[onDeleteModalOpen],
);

```

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -41,12 +41,19 @@ class GroupRepository implements Repository {
...
+			.deleteById(id)
+			.execute();
+
+		return Boolean(deletedRowsCount);
 	}
 
-	public find(): ReturnType<Repository["find"]> {
-		return Promise.resolve(null);
+	public async find(id: number): Promise<GroupModel | null> {
+		const group = await this.groupModel.query().findById(id);
```
GvoFor, line 54:
```suggestion
		const group = await this.groupModel.query().findById(id).execute();
```

apps/frontend/src/modules/groups/slices/group.slice.ts
GvoFor: Do we really need to keep track `groupsTotalCount` in state? Can't we use `groups.length` instead? Or this may lead to unwanted rerenders?

apps/frontend/src/pages/access-management/libs/components/group-menu/group-menu.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const GroupMenu = ({ groupId, onDelete }: Properties): React.ReactElement => {
+	const { isOpened, onClose, onOpen } = usePopover();
+
+	const handleDeleteClick = useCallback(() => {
+		onDelete(groupId);
+	}, [onDelete, groupId]);
+
+	return (
+		<div className={styles["options-container"]}>
```
GvoFor, line 19:
```suggestion
		<div className={styles["menu-container"]}>
```

apps/frontend/src/modules/groups/slices/group.slice.ts
perekotypole: @Fjortis It has an effect on table pagination. `groups.length` will only show the number of elements on a page.

apps/frontend/src/pages/access-management/libs/components/group-menu/group-menu.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useCallback, usePopover } from "~/libs/hooks/hooks.js";
+
+import styles from "./styles.module.css";
+
+type Properties = {
+	groupId: number;
+	onDelete: (groupId: number) => void;
+};
+
+const GroupMenu = ({ groupId, onDelete }: Properties): React.ReactElement => {
```
liza-veis, line 11:
```suggestion
const GroupMenu = ({ groupId, onDelete }: Properties): JSX.Element => {
```

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -124,6 +171,14 @@ const AccessManagement = (): JSX.Element => {
 			>
 				<GroupCreateForm onSubmit={handleGroupCreateSubmit} />
 			</Modal>
+			{groupId !== null && (
```
liza-veis, line 174:
```suggestion
			{hasSelectedGroup && (
```

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -61,6 +62,20 @@ const AccessManagement = (): JSX.Element => {
...
+		onOpen: onModalOpen,
+	} = useModal();
+
+	const {
+		isOpened: isDeleteModalOpen,
+		onClose: onDeleteModalClose,
+		onOpen: onDeleteModalOpen,
+	} = useModal();
+
+	const [groupId, setGroupId] = useState<null | number>(null);
```
liza-veis, line 77:
```suggestion
	const [selectedGroupId, setSelectedGroupId] = useState<null | number>(null);
```

apps/frontend/src/modules/groups/slices/group.slice.ts
```diff
@@ -22,6 +28,8 @@ const initialState: State = {
 	groupCreateStatus: DataStatus.IDLE,
 	groups: [],
 	groupsTotalCount: 0,
+	page: 1,
+	pageSize: 10,
```
liza-veis, line 32: As we need to handle pagination, it's too complicated to imitate it in reducer. Instead of this let's just do the same trick as when we close modals,  and call onPageSet with the initial page value on success, which will trigger the request and update the group. We want to keep the same pageSize and update page to the initial value.

```suggestion
```

apps/frontend/src/libs/hooks/use-popover-position/use-popover-position.hook.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	useEffect(() => {
+		if (!hasPositioning || !isOpen || !reference.current) {
+			setPosition({});
+
+			return;
+		}
+
+		const { top } = reference.current.getBoundingClientRect();
+		const scrollOffset = window.scrollY || document.documentElement.scrollTop;
```
liza-veis, line 22: I meant to use either `window.scrollY` or `document.documentElement.scrollTop` only in my first question

apps/frontend/src/libs/hooks/use-popover-position/use-popover-position.hook.ts
```diff
@@ -0,0 +1,32 @@
...
+	hasPositioning: boolean = false,
+): Properties => {
+	const [position, setPosition] = useState<Properties>({});
+
+	useEffect(() => {
+		if (!hasPositioning || !isOpen || !reference.current) {
+			setPosition({});
+
+			return;
+		}
```
liza-veis, line 19: I think we don't need to change positioning depending on isOpen (we will just not display it).
```suggestion
		if (!hasPositioning || !reference.current) {
			return;
		}
```

apps/frontend/src/libs/components/popover/popover.tsx
```diff
@@ -1,29 +1,55 @@
...
 	isOpened,
 	onClose,
 }: Properties): JSX.Element => {
 	const popoverReference = useRef<HTMLDivElement>(null);
+	const popoverTargetReference = useRef<HTMLDivElement>(null);
+
+	const position = usePopoverPosition(
+		popoverTargetReference,
+		isOpened,
+		hasPositioning,
```
liza-veis, line 30: let's rename it to hasAbsolutePositioning or hasFixedPositioning

liza-veis (changes requested):

apps/frontend/src/pages/access-management/libs/helpers/get-group-columns.helper.tsx
```diff
@@ -17,6 +20,14 @@ const getGroupColumns = (): TableColumn<GroupRow>[] => [
 			formatDate(new Date(group.createdAt), "d MMM yyyy HH:mm"),
 		header: "Created At",
 	},
+	{
+		accessorFn: (group: GroupRow): React.ReactNode => (
+			<GroupMenu groupId={group.id} onDelete={onDelete} />
+		),
+		header: "",
+		id: "options",
```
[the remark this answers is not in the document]
kolibri753: a non-string header (like an empty string), the column requires an id

apps/frontend/src/libs/hooks/use-popover-position/use-popover-position.hook.ts
kolibri753: I prefer to leave both, but I can choose one (window.scrollY modern option)

apps/frontend/src/modules/groups/slices/group.slice.ts
kolibri753: I deleted the pagination code from the slice (except for `groupsTotalCount`), and it is working fine without any additional changes. Should I still modify this?

apps/frontend/src/libs/hooks/use-popover-position/use-popover-position.hook.ts
liza-veis: I think we need to leave document.documentElement.scrollTop only, it should cover all the cases

kolibri753: noted

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -83,6 +100,38 @@ const AccessManagement = (): JSX.Element => {
 		[dispatch, onModalClose, userPage, userPageSize],
 	);
 
+	const handleGroupDelete = useCallback((): void => {
+		if (hasSelectedGroup) {
+			void dispatch(
+				groupActions.deleteById({
+					groupQuery: { page: groupPage, pageSize: groupPageSize },
+					id: selectedGroupId as number,
+					userQuery: { page: userPage, pageSize: userPageSize },
```
liza-veis, line 109: we can remove groups/users dispatch inside this action
```suggestion
					id: selectedGroupId as number,
```
I want it to be done like this

```tsx
	useEffect(() => {
		if (groupDeleteStatus === DataStatus.FULFILLED) {
			onGroupPageChange(INITIAL_PAGE);
	        onUserPageChange(INITIAL_PAGE);
		}
	}, [projectPatchStatus, onUserPageChange, onGroupPageChange]);
```

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -83,6 +100,38 @@ const AccessManagement = (): JSX.Element => {
...
+		dispatch,
+		groupPage,
+		groupPageSize,
+		selectedGroupId,
+		userPage,
+		userPageSize,
+		onDeleteModalClose,
+	]);
+
+	const onDelete = useCallback(
```
perekotypole, line 127: Shouldn't it be `handleDelete`? (A2)

apps/frontend/src/pages/access-management/access-management.tsx
kolibri753: Sorry, I'm having trouble understanding what `INITIAL_PAGE` refers to. If it’s just 1, the implementation is simple, but after deletion, there's an issue with pagination. I can investigate it further, but wouldn't it be better to stay on the current page where the action was performed, and only move to the previous page if the deleted group was the last one on that page?

liza-veis: agree with you, let's leave it as you currently done it

apps/frontend/src/pages/access-management/libs/components/group-menu/group-menu.tsx
liza-veis: this comment is still not fixed

liza-veis (approved):

PR #129, opened 2024-08-30
fix: Abscence of a logo on the Sign In and Sign Up pages gf-121

kolibri753:
The parent component (Auth) is responsible for deciding whether the image has loaded or failed, by passing down the isImageLoaded state and onImageError handler as props to the Logo component. The Logo component then uses these props to conditionally render either the image or a fallback (styled "LOGO" text).

image


apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -59,12 +61,20 @@ const Auth = (): JSX.Element => {
...
 			<section className={styles["auth-container"]}>
 				<div className={styles["left-side"]}>
-					{/* TODO: replace logo */}
-					<img alt="logo" className={styles["logo-wrapper"]} src="" />
+					{/* TODO: add logoSrc */}
+					<Logo
+						isImageLoaded={isImageLoaded}
+						logoSrc=""
+						onImageError={handleImageError}
+					/>
```
liza-veis, line 77: Too complicated logic for this case 😅 let's keep it as it was, the only thing that is needed is to use a random image as a placeholder. You can download "LOGO" from the design as an image for this.

liza-veis (changes requested):

apps/frontend/src/pages/auth/auth.tsx
liza-veis: I think we don't need to handle case when image is not loaded, the application won't build if we we will import invalid image path

kolibri753: It was in case we didn't import path at all :sweat_smile:

kolibri753: Oh.. I was totally thinking the same thing and asked about it earlier, but unfortunately, you didn't see it then. Okay, I'll just revert it to how it was :black_cat:

liza-veis: oh sorry, probably missed that question

kolibri753: no worries :smile_cat: :

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -1,5 +1,6 @@
 import { useNavigate } from "react-router-dom";
 
+import logoSrc from "~/assets/images/icons/logo.svg";
```
kolibri753, line 3: not sure about the path here

liza-veis (approved):

Issue #131, opened 2024-08-30
feat: Sign Up page fields validation,  Sign up button inconsistent text


VeronikaHordus:
### What feature?

1. Sign Up button text inconsistent to the Mockup. The text of the button should be Create Account.

2. Field validation rules : 
 Email  field: 

- The format should include a local part, an "@" symbol, and a domain part. 
- The length <50
Local part (before the “@”) : 
- should contain alphanumeric characters, optionally separated by dots, underscores, or hyphens;
- cannot start or end with a dot or contain consecutive dots;
Domain part (after the “@”): 
- must contain at least two characters;
- must contain at least one dot;
- should consist of alphanumeric characters (A-Z, a-z, 0-9) and hyphens;

Name field:
- The name can contain any characters, including letters, numbers, spaces, and special characters;
- at least 3 characters;
- max 50 characters

Password must include: 

- At least 5 characters ;
- Not more than 50 characters
- At least 1 uppercase letter (A-Z);
- At least 1 lowercase letter  (a-z);
- At least 1 number ;
- At least 1 special character (e.g., !@#$%)"

### Screenshots

![image](https://github.com/user-attachments/assets/5309fcbd-e0aa-4af7-bbb2-fe98174efb21)

kolibri753:
Clarification (about Email):

1. Domain Part: Should the dot (.) in the domain part be followed by at least two letters? For example, is username@example.c valid, or should it be rejected?
2. Local Part: Is a dot followed by a hyphen (.-) allowed in the local part? For instance, would user.-name@example.com be valid?

P.S. I believe both cases are invalid but wanted to confirm.

VeronikaHordus:
I believe its good to make both ways invalid (sorry for long responce). 
I will add your clarifications in main info for validation, thanks!

AnnaRomaniuta:
@kolibri753 Hello, I'm a little confused, maybe you can help me. When I try to register with an invalid address, for example without a domain part or without a special character @, I receive a message from the browser, not from the application![Image](https://github.com/user-attachments/assets/b4695128-531d-46d9-a6fc-44b18c6fc7b5)

AnnaRomaniuta:
In general, everything works and looks good, a bug has been created for some discrepancies. Good job!

PR #132, opened 2024-08-30
feat: Sign Up page fields validation, Sign up button inconsistent text gf-131

kolibri753:
Changes:

- fixed sign-up button text;
- updated existing validation rules;
- added new validation rules.


packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
```diff
@@ -11,14 +11,29 @@ const userSignUp: z.ZodType<UserSignUpRequestDto> = z
...
+				{
+					message: UserValidationMessage.EMAIL_WRONG,
+				},
+			),
 		name: z
 			.string()
 			.trim()
 			.min(UserValidationRule.NAME_MINIMUM_LENGTH, {
-				message: UserValidationMessage.NAME_REQUIRED,
+				message: UserValidationMessage.NAME_TOO_SHORT,
```
perekotypole, line 36: I'd propose to leave `NAME_REQUIRED`. We can only use`.min(1)`  to check required string with zod, and 'name is required' message is more propriate than 'need more than 1 symbol' or smth

kolibri753: Since the name must be at least 3 characters, I think `NAME_TOO_SHORT` is a better fit than just `NAME_REQUIRED`.

liza-veis: can't we pass required error additionally, like this `.string({ required_error: ''})`

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -2,10 +2,20 @@ import { UserValidationRule } from "./user-validation-rule.enum.js";
 
 const UserValidationMessage = {
 	EMAIL_REQUIRED: "Email is required.",
+	EMAIL_TOO_LONG: `Email must be less than ${String(UserValidationRule.EMAIL_MAXIMUM_LENGTH)} characters long.`,
 	EMAIL_WRONG: "Email is wrong.",
 	NAME_REQUIRED: "Name is required.",
 	NAME_TOO_LONG: `Name is too long (> ${String(UserValidationRule.NAME_MAXIMUM_LENGTH)} symbols).`,
```
liza-veis, line 8: Let's also modify this error message to keep it similar to EMAIL_TOO_LONG

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
```diff
@@ -11,14 +11,29 @@ const userSignUp: z.ZodType<UserSignUpRequestDto> = z
...
+						!!localPart &&
+						!!domainPart &&
+						UserValidationRule.EMAIL_LOCAL_PART_PATTERN.test(localPart) &&
+						UserValidationRule.EMAIL_DOMAIN_PART_PATTERN.test(domainPart)
+					);
+				},
+				{
+					message: UserValidationMessage.EMAIL_WRONG,
+				},
+			),
```
liza-veis, line 31: Doesn't Zod have built-in validation for emails? I think this logic can be fully covered by it. Also, let's add email validation for sign-in too (for password is not needed, users should know their passwords).

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
```diff
@@ -11,14 +11,29 @@ const userSignUp: z.ZodType<UserSignUpRequestDto> = z
 			.min(UserValidationRule.EMAIL_MINIMUM_LENGTH, {
```
liza-veis, line 11: Could you also call trim (built-in) for all values (name, password, email) and the same for sign in

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
```diff
@@ -28,7 +43,37 @@ const userSignUp: z.ZodType<UserSignUpRequestDto> = z
 			.trim()
 			.min(UserValidationRule.PASSWORD_MINIMUM_LENGTH, {
 				message: UserValidationMessage.PASSWORD_MINIMUM_LENGTH,
-			}),
+			})
+			.max(UserValidationRule.PASSWORD_MAXIMUM_LENGTH, {
+				message: UserValidationMessage.PASSWORD_TOO_LONG,
+			})
+			.refine(
```
liza-veis, line 50: let's use regex method for all regexes
```suggestion
			.regex(
```

liza-veis (changes requested):

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
kolibri753: Yes, but this validation break some [rules](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/issues/131):

- a hyphen should not be placed next to a dot;
- a hyphen should not be at the beginning or end of the local part;

P.S. I thought [it](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/issues/131#issuecomment-2321043852) shouldn't pass, but now I'm not sure (about Local Part)

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
kolibri753: I’ve already called trim() for all the fields, or am I mistaken :thinking:

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
kolibri753: Okay, I wasn't sure about that. I was wondering how a user could write a shorter email...

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
kolibri753: Yes, we can pass a required error using `.string({ required_error: '' })`. However, it won't be seen if the minimum length error is triggered first.
```
email: z
			.string({
				required_error: UserValidationMessage.EMAIL_REQUIRED,
			})
			.trim()
			.min(UserValidationRule.EMAIL_MINIMUM_LENGTH, {
				message: UserValidationMessage.EMAIL_TOO_SHORT,
			})
```
image

liza-veis: Hm, got this. then let's leave it

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
liza-veis: I think for our project such a strict email validation is not needed. So it's better to keep built-in email validation and ask qa to adjust the rules.

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
liza-veis: didn't noticed it 😅

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
liza-veis: Please agree this with qa first before changing

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
liza-veis: I'm thinking that maybe in this case we can change other messages with required, for example for name can: name can be at least characters long

packages/shared/src/modules/users/libs/validation-schemas/user-sign-up.validation-schema.ts
kolibri753: agreed on simplifying :+1:

liza-veis (approved):

what1s1ove (approved):

PR #137, opened 2024-09-01
feat: add an ability to edit a project gf-88

sandrvvu:
Add an options icon at the end of the project list item.
When the icon is clicked, an options menu should be opened with an Edit option.
On the Edit item click, a modal should be opened to update a project.
The data for the modal should be prepopulated with the current project data.
The validation is the same as for project creation

image
image
image


apps/backend/src/modules/projects/libs/types/types.ts
```diff
@@ -3,4 +3,6 @@ export {
 	type ProjectCreateResponseDto,
 	type ProjectGetAllItemResponseDto,
 	type ProjectGetAllResponseDto,
+	type ProjectUpdateRequestDto,
+	type ProjectUpdateResponseDto,
```
perekotypole, line 7: https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/107#discussion_r1736789709
I`m not sure, but according to my task if you use patch, make sure to use 'patch' for namings as well

apps/frontend/src/modules/projects/projects-api.ts
```diff
@@ -28,6 +32,23 @@ class ProjectApi extends BaseHTTPApi {
 
 		return await response.json<ProjectGetAllResponseDto>();
 	}
+
+	public async update(
```
perekotypole, line 36: https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/107#discussion_r1738321776
same about patch namings (I used "patch" literally for everything)

packages/shared/src/modules/projects/projects.ts
```diff
@@ -4,5 +4,8 @@ export {
 	type ProjectCreateRequestDto,
 	type ProjectGetAllItemResponseDto,
 	type ProjectGetAllResponseDto,
+	type ProjectUpdateRequestDto,
+	type ProjectUpdateResponseDto,
 } from "./libs/types/types.js";
 export { projectCreate as projectCreateValidationSchema } from "./libs/validation-schemas/validation-schemas.js";
+export { projectUpdate as projectUpdateValidationSchema } from "./libs/validation-schemas/validation-schemas.js";
```
Vitaliistf, line 11: I think you can use a single export statement for both exports.

apps/frontend/src/assets/images/icons/edit.svg
GvoFor, line 1: Remove `fill` attribute from `<path>` tag, so it does not effect on icon color. If you need to change the color of an icon, you should wrap it with some component and set `color: ...` style for that wraper to whatever color you need

apps/frontend/src/assets/images/icons/options.svg
GvoFor, line 1: Same for this icon. And here take attantion, that this icon is passed to `IconButton`, which already sets the color as it supposed to be

apps/frontend/src/pages/projects/components/project-card/styles.module.css
```diff
@@ -1,12 +1,16 @@
 .project {
 	display: flex;
 	align-items: center;
+	justify-content: space-between;
 	height: 81px;
 	padding: 0 24px;
```
GvoFor, line 6: It's better to use `min-height` instead of fixed `height`. And in Figma padding is `16px 24px`, so adjust this too please

apps/frontend/src/pages/projects/components/project-card/project-card.tsx
```diff
@@ -16,9 +18,14 @@ const ProjectCard = ({ project }: Properties): JSX.Element => {
...
 	return (
-		<NavLink className={styles["project"] ?? ""} to={projectRoute}>
-			{project.name}
-		</NavLink>
+		<div className={styles["project"] ?? ""}>
+			<NavLink className={styles["project-link"] ?? ""} to={projectRoute}>
+				{project.name}
+			</NavLink>
+			<ProjectPopover project={project}>
+				<Icon height={20} name="options" width={20} />
```
GvoFor, line 26: This should be `IconButton` instead of `Icon`

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -66,8 +68,33 @@ class ProjectService implements Service {
...
+		const existingProject = await this.projectRepository.find(id);
+
+		if (!existingProject) {
+			throw new ProjectError({
+				message: ExceptionMessage.PROJECT_NOT_FOUND,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
+
+		const projectWithSameName = await this.projectRepository.findByName(
```
liza-veis, line 84: as in other places
```suggestion
		const existingProject = await this.projectRepository.findByName(
```

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -66,8 +68,33 @@ class ProjectService implements Service {
...
+			projectData.name,
+		);
+
+		if (projectWithSameName && projectWithSameName.getId() !== id) {
```
liza-veis, line 88: remove getId method
```suggestion
		if (existingProject && existingProject.toObject().id !== id) {
```

liza-veis (changes requested):

apps/frontend/src/pages/projects/components/project-card/project-card.tsx
liza-veis: Please take a look at these two discussions regarding IconButton and the options menu overall, it would be nice to have a separate component for options menu

https://bsa2024.slack.com/archives/C07HWM8616U/p1725289146857129
https://bsa2024.slack.com/archives/C07HWM8616U/p1725290415052879

apps/frontend/src/assets/images/icons/edit.svg
GvoFor: Let's call this icon `pencil`

apps/frontend/src/assets/images/icons/options.svg
GvoFor: Let's call this icon `ellipsis`

apps/frontend/src/libs/components/menu-item/styles.module.css
```diff
@@ -0,0 +1,31 @@
...
+
+.project-menu-item-text {
+	margin: 0;
+	font-family: Inter, sans-serif;
+	font-size: 16px;
+	font-weight: 500;
+	line-height: 1.2;
+}
+
+.project-menu-item--primary {
```
GvoFor, line 25: We don't use BEM in our project, so one dash will be enough
```suggestion
.project-menu-item-primary {
```

apps/frontend/src/libs/components/menu-item/styles.module.css
```diff
@@ -0,0 +1,31 @@
...
+	color: var(--color-text-primary);
+}
+
+.project-menu-item--danger {
```
GvoFor, line 29: Same here

apps/frontend/src/libs/components/popover/popover.tsx
```diff
@@ -2,36 +2,43 @@ import {
...
 } from "~/libs/hooks/hooks.js";
 
 import styles from "./styles.module.css";
 
 type Properties = {
 	children: React.ReactNode;
 	content: React.ReactNode;
+	isPopoverOpened: boolean;
+	onPopoverClose: () => void;
+	onPopoverOpen: () => void;
```
GvoFor, line 14: This looks like we violating A5 quality criteria. Also names `onClose` and `onOpen` don't represent what they actually do: `onOpen` used to open popover, and `onClose` is used to close it -- so they are not handlers but rather actions. 
So according to all that I said:
```suggestion
	isOpened: boolean;
	close: () => void;
	open: () => void;
```
Also it would be nice to do the same changes for modal component

apps/frontend/src/libs/components/popover/popover.tsx
```diff
@@ -2,36 +2,43 @@ import {
...
 };
 
-const Popover = ({ children, content }: Properties): JSX.Element => {
-	const [isOpened, setIsOpened] = useState<boolean>(false);
+const Popover = ({
+	children,
+	content,
```
GvoFor, line 19: I don't think that now, when we have `usePopover` hook that controls visibility of popover we need `content` prop. Now we don't want to wrap a component that should response for click and open popover with `Popover` component. Instead of this, `Popover` will take only children, representing its content, and it will be placed near the component it's belong to. 

Hm, however I'm not sure how then we will position it where it suppose to be. @liza-veis Can you suggest something, or say if my suggestion is not good

apps/frontend/src/libs/hooks/use-popover-state/use-popover-state.hook.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback, useState } from "~/libs/hooks/hooks.js";
+
+type Properties = {
+	isPopoverOpened: boolean;
+	onPopoverClose: () => void;
+	onPopoverOpen: () => void;
```
GvoFor, line 6: Read my comment about `Popover` props

apps/frontend/src/pages/projects/components/project-popover/project-popover.tsx
GvoFor, line 1: Instead of this component create common `Menu` component and use it. It can take a list of elements of type `MenuItem` (not sure about name), which will looks like:
```ts
{
  icon: IconName;
  variant: "primary" | "danger";
  onClick: () => void;
}
```
and using this data map it to list of your `MenuItem` component

P.s. The modal logic should be outside of this component and will be handled in that `onClick` callbacks. This component is responsible just for visualizing a list of `MenuItem`s

apps/frontend/src/pages/projects/components/project-popover/project-popover.tsx
```diff
@@ -0,0 +1,85 @@
...
+			void dispatch(projectActions.patch({ id, payload }));
+		},
+		[dispatch],
+	);
+
+	return (
+		<>
+			<Popover
+				content={
+					<div className={styles["project-menu-popover"]}>
```
liza-veis, line 55: Let's create a separate component for menu

apps/frontend/src/pages/projects/components/project-popover/project-popover.tsx
```diff
@@ -0,0 +1,85 @@
...
+								onClick={handleEditClick}
+							/>
+						</div>
+					</div>
+				}
+				isPopoverOpened={isPopoverOpened}
+				onPopoverClose={onPopoverClose}
+				onPopoverOpen={onPopoverOpen}
+			>
+				{children}
```
liza-veis, line 69: 1. keep the trigger here as a part of component because it's a specific component, not a common one
2. Use IconButton as a trigger, not usual button. That's why we added usePopover hook, to be able to pass onClick to trigger, outside of popover.
3. Related to the 2 point, we need to remove button wrapper for trigger element inside popover. And then we will need to wrap avatar inside button from outside
```suggestion
			<IconButton onClick={onPopoverOpen} />
```

apps/frontend/src/pages/projects/components/project-popover/project-popover.tsx
```diff
@@ -0,0 +1,85 @@
...
+			<Modal
+				isModalOpened={isModalOpened}
+				onModalClose={onModalClose}
+				title="Update project"
+			>
+				<ProjectUpdateForm
+					onSubmit={handleProjectEditSubmit}
+					project={project}
+				/>
+			</Modal>
```
liza-veis, line 80: This modal, and all the logic related to it and statuses should be controlled in the project page itself. If we want to keep this project-popover component we need to pass here onEdit, onDelete etc. as props. But I think there is no need to keep it the separate component, it will be just `<Popover />` if we remove all the logic here, so I think we can remove it and define it as it is where we use it (ProjectCard). But pay attention that the edit modal should be located on Projects page, not project card
```suggestion
```

liza-veis (changes requested):

apps/frontend/src/pages/projects/components/project-popover/project-popover.tsx
sandrvvu: yeah , I thought about the general Menu component, but somehow I changed my mind because of the abundance of different functionality for each Popover, but right it's still possible to "replay" it differently

apps/frontend/src/libs/components/popover/popover.tsx
liza-veis: We need to have content to position it correctly and also we have additional wrappers for the popover and it's content

apps/frontend/src/libs/components/popover/popover.tsx
liza-veis: agree, but instead close/open should be, onClose, onOpen

apps/frontend/src/libs/components/menu-item/styles.module.css
liza-veis: And it should go without project prefix

apps/frontend/src/libs/components/menu-item/libs/constants/icon-size.constant.ts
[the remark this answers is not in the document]
liza-veis: agree

what1s1ove (approved):
lgtm after Lisa's comments

apps/frontend/src/libs/components/menu/libs/components/menu-item/menu-item.tsx
```diff
@@ -0,0 +1,38 @@
+import { Icon } from "~/libs/components/components.js";
+import { getValidClassNames } from "~/libs/helpers/helpers.js";
+
+import { type IconName } from "../../../../icon/libs/types/types.js";
```
liza-veis: Let's move IconName to common types

apps/frontend/src/libs/components/menu/libs/components/components.ts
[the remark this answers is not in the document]
sandrvvu: mhm i like this better, i'll move it

liza-veis (changes requested):

apps/frontend/src/libs/components/header/styles.module.css
```diff
@@ -29,3 +29,8 @@
 	color: var(--color-text-primary);
 	cursor: pointer;
 }
+
+.user-button {
+	all: unset;
```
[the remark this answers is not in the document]
GvoFor: Also you should add `border-radius: 50%` so popover doesn't open when click outside of avatar circle.
Or another solution (which in my opinion sounds better) is to rid of button wrapper, and add `onClick` prop to `Avatar` component itself (so we don't need to duplicate `border-radius: 50%` in two separate files)

apps/frontend/src/libs/components/menu/libs/components/components.ts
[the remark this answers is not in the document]
GvoFor: As for me, this component is placed where it supposed to be. The point is that it will be used only with `Menu` component, so it will be better to pass a list of objects with `icon`, `label` and `onClick` fields as a prop to `Menu` component, which then will convert those objects to `MenuItem` components. This will also reduce repeatition of `<MenuItem  ... />` in places where `Menu` is used

apps/frontend/src/pages/projects/components/project-card/project-card.tsx
```diff
@@ -2,23 +2,33 @@ import { NavLink } from "react-router-dom";
...
+	const handleEditClick = useCallback(() => {
+		onEditClick(project);
+	}, [onEditClick, project]);
+
 	return (
-		<NavLink className={styles["project"] ?? ""} to={projectRoute}>
-			{project.name}
-		</NavLink>
+		<div className={styles["project"]}>
+			<NavLink className={styles["project-link"] ?? ""} to={projectRoute}>
```
GvoFor, line 27: We agreed on [this approach](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/167#discussion_r1743086886)
```suggestion
			<NavLink className={styles["project-link"] as string} to={projectRoute}>
```

apps/frontend/src/libs/components/menu/menu.tsx
GvoFor, line 1: In case your `Menu` consist of both "options" button and popover, you should add `variant: "primary" | "secondary"` prop (or something else) so it can be used for both cases:
image
and
image

P.s. It seems to me, that on second screen there is a mistake in Figma design with position of popover, so I mean only color and border of button component. If I'm not mistaking, you can pass `variant: "primary" | "secondary"` prop to `IconButton`, and it already have appropriate styles

liza-veis: No need to add this for now, we will add this when we need it

apps/frontend/src/pages/projects/components/project-card/project-card.tsx
```diff
@@ -2,23 +2,33 @@ import { NavLink } from "react-router-dom";
...
+				{project.name}
+			</NavLink>
+			<ProjectMenu onEditClick={handleEditClick} />
+		</div>
```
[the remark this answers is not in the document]
sandrvvu: When you click on the options button, it redirects you to some project page. I was trying to fix it using event.preventDefault and event.stopPropagation, but it looks kinda messy, is there any other ways to trigger button?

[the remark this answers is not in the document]
liza-veis: Hm, actually I'm thinking that we cannot nest a button inside link, so it should've gone underneath as previously anyway. Usually, for such cases underneath link just has absolute positioning and is stretched to cover container sizes. I think this will be a more correct way to implement this and it should fix the problem

liza-veis (approved):

PR #143, opened 2024-09-02
feat: add pagination to the users table gf-82

Vitaliistf:
Screenshot 2024-09-02 at 13 23 15

Added pagination to the users table on the access management page, updated required actions, implemented pageable users retrieving from the backend, based on page number and page size.


apps/backend/src/modules/users/libs/types/types.ts
```diff
@@ -1,4 +1,5 @@
 export {
+	type PaginationParameters,
```
perekotypole, line 2: I think it's better to move importing `PaginationParameters` to common types. It's not related only to user model and will be used in others as well

apps/frontend/src/modules/users/libs/types/types.ts
```diff
@@ -1,4 +1,5 @@
 export {
+	type PaginationParameters,
```
perekotypole, line 2: The same with moving it to common types

Vitaliistf: Oh, yeah, I agree, thanks

apps/backend/src/libs/types/repository.type.ts
```diff
@@ -2,7 +2,7 @@ type Repository<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(id: number): Promise<boolean>;
 	find(id: number): Promise<null | T>;
-	findAll(): Promise<T[]>;
+	findAll(page: number, size: number): Promise<T[]>;
```
what1s1ove, line 5:
```suggestion
	findAll(parameters: Type): Promise<T[]>;
```

apps/frontend/src/modules/users/slices/users.slice.ts
```diff
@@ -8,12 +8,14 @@ import { loadAll, updateProfile } from "./actions.js";
...
 type State = {
 	dataStatus: ValueOf<typeof DataStatus>;
+	totalUsers: number;
 	updateProfileStatus: ValueOf<typeof DataStatus>;
 	users: UserGetAllItemResponseDto[];
 };
 
 const initialState: State = {
 	dataStatus: DataStatus.IDLE,
+	totalUsers: 0,
```
what1s1ove, line 18:
```suggestion
	usersTotalCount: 0,
```

Vitaliistf: Fixed, thanks!

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -41,8 +48,16 @@ class UserRepository implements Repository {
 		return user ? UserEntity.initialize(user) : null;
 	}
 
-	public async findAll(): Promise<UserEntity[]> {
-		const users = await this.userModel.query().execute();
+	public async findAll({
+		page,
+		pageSize,
+	}: PaginationParameters): Promise<UserEntity[]> {
+		const offset = --page * pageSize;
```
yevhenii-hladkov, line 55: it's generally better to move calculations to the service rather than keeping them in the repository

packages/shared/src/libs/types/pagination-response.type.ts
```diff
@@ -0,0 +1,5 @@
+type PaginationResponse = {
+	totalItems: number;
+};
```
liza-veis, line 3: Let's make it a generic 
```suggestion
type PaginationResponseDto = <T>{
    items: T[];
	totalItems: number;
};
```

packages/shared/src/libs/types/pagination-parameters.type.ts
```diff
@@ -0,0 +1,6 @@
+type PaginationParameters = {
```
liza-veis, line 1:
```suggestion
type PaginationQueryParameters = {
```

apps/frontend/src/pages/ui/ui.tsx
```diff
@@ -33,7 +33,7 @@ const Ui = (): JSX.Element => {
 	const modal = useModal();
 
 	useEffect(() => {
```
liza-veis, line 35: now we can remove pagination from this page

apps/frontend/src/modules/users/slices/actions.ts
```diff
@@ -13,12 +16,12 @@ import { name as sliceName } from "./users.slice.js";
 
 const loadAll = createAsyncThunk<
 	UserGetAllResponseDto,
-	undefined,
+	PaginationParameters,
 	AsyncThunkConfig
->(`${sliceName}/load-all`, (_, { extra }) => {
+>(`${sliceName}/load-all`, (paging, { extra }) => {
```
liza-veis, line 21:
```suggestion
>(`${sliceName}/load-all`, (query, { extra }) => {
```

apps/backend/src/modules/users/user.service.ts
```diff
@@ -68,11 +68,17 @@ class UserService implements Service {
 		return item.toObject();
 	}
 
-	public async findAll(): Promise<UserGetAllResponseDto> {
-		const items = await this.userRepository.findAll();
+	public async findAll(
+		parameters: PaginationParameters,
```
liza-veis, line 72:
```suggestion
		query: PaginationQueryParameters,
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -41,8 +48,16 @@ class UserRepository implements Repository {
...
+		const users = await this.userModel
```
liza-veis, line 56: As I remember there is a `.page(page, count)` method in objection query builder. And it returns results and total which we need to return from the repository. Remove count method above

liza-veis (changes requested):

apps/backend/src/modules/users/user.repository.ts
Vitaliistf: Okay, I've implemented these suggested changes, so to take advantage of getting items page and their total count at the same time, now the base repository's `findAll` returns `Promise<{ items: T[] }>` (not `Promise<PaginationResponseDto<T>>` because not every item collection should be pageable) and there are errors in other repositories now, because their `findAll` are different, they return arrays of Entities. Should I change it? Some logic from the service will be moved to the repo.

liza-veis: Let's return an object with items property for all findAll methods to make a consistent response type

what1s1ove (approved):
lgtm after Lisa's comments

apps/backend/src/modules/users/user.repository.ts
liza-veis: I think there is no need for a separate type because it may differ too much, on repository / service for now we will use either `{ items: T }` or wrap resource type into Pagination type or use ResourceGetAllResponseDto. For frontend, I think we won't need `{ items: T }`, because we will keep in the store only items as it was previously. So the only thing we will need to change on frontend is that instead of setting payload to the store we will set payload.items

Vitaliistf: I'm sorry, I just got a bit confused, and when I started working on it, I've figured out that there are already types for findAll declared, so the only backend needs to be changed a bit. Thank you!

apps/backend/src/libs/types/repository.type.ts
```diff
@@ -1,8 +1,10 @@
+import { type PaginationQueryParameters } from "~/libs/types/types.js";
+
 type Repository<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(id: number): Promise<boolean>;
 	find(id: number): Promise<null | T>;
-	findAll(): Promise<T[]>;
+	findAll(parameters: PaginationQueryParameters): Promise<{ items: T[] }>;
```
liza-veis, line 7: let's make it unknown for repository
```suggestion
	findAll(payload: unknown): Promise<{ items: T[] }>;
```

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -45,12 +46,14 @@ class GroupRepository implements Repository {
 		return Promise.resolve(null);
 	}
 
-	public async findAll(): Promise<GroupEntity[]> {
+	public async findAll(): Promise<GroupGetAllResponseDto> {
```
liza-veis, line 49: Use entity type on repository level, either `Promise<{ items: GroupEntity[] }>` or `Promise<PaginationResponseDto<UserEntity>>` for cases with pagination

apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -45,12 +46,14 @@ class GroupRepository implements Repository {
...
 		const groups = await this.groupModel
 			.query()
 			.withGraphFetched("[permissions, users]");
 
-		return groups.map((group) => GroupEntity.initialize(group));
+		return {
+			items: groups.map((group) => GroupEntity.initialize(group).toObject()),
```
liza-veis, line 55: remove .toObject call on repository level, we need to return entity
```suggestion
			items: groups.map((group) => GroupEntity.initialize(group)),
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -41,10 +51,18 @@ class UserRepository implements Repository {
...
 
-	public async findAll(): Promise<UserEntity[]> {
-		const users = await this.userModel.query().execute();
-
-		return users.map((user) => UserEntity.initialize(user));
+	public async findAll({
+		page,
+		pageSize,
+	}: PaginationQueryParameters): Promise<UserGetAllResponseDto> {
+		const result = await this.userModel.query().page(--page, pageSize);
```
liza-veis, line 58:
```suggestion
		const { results, total } = await this.userModel.query().page(--page, pageSize);
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -1,15 +1,25 @@
...
 
 class UserRepository implements Repository {
 	private userModel: typeof UserModel;
 	public constructor(userModel: typeof UserModel) {
 		this.userModel = userModel;
 	}
 
+	public async count(): Promise<number> {
+		return await this.userModel.query().resultSize();
+	}
```
liza-veis, line 21:
```suggestion
```

liza-veis (changes requested):

apps/backend/src/modules/users/user.repository.ts
Vitaliistf: Oh yeah, thank you. Today have big problems with power outages, so I'm kind of in the rush, forgetting things😅. Fixed everything.

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -40,7 +41,7 @@ class UserController extends BaseController {
 		this.userService = userService;
 
 		this.addRoute({
-			handler: () => this.findAll(),
+			handler: (request) => this.findAll(request),
```
liza-veis, line 44: As in other places
```suggestion
			handler: (options) => this.findAll(options),
```

apps/backend/src/modules/users/user.controller.ts
```diff
@@ -65,20 +66,45 @@ class UserController extends BaseController {
...
+	 *                  totalItems:
+	 *                    type: integer
+	 *                  totalPages:
+	 *                    type: integer
 	 */
-	private async findAll(): Promise<APIHandlerResponse> {
+
+	private async findAll({
+		query,
+	}: APIHandlerOptions): Promise<APIHandlerResponse> {
```
liza-veis, line 103: then you won't need to cast `query as PaginationQueryParameters` below
```suggestion
	private async findAll({
		query,
	}: APIHandlerOptions<{
			query: PaginationQueryParameters;
		}>): Promise<APIHandlerResponse> {
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -41,10 +45,18 @@ class UserRepository implements Repository {
...
 
-	public async findAll(): Promise<UserEntity[]> {
-		const users = await this.userModel.query().execute();
+	public async findAll({
+		page,
+		pageSize,
+	}: PaginationQueryParameters): Promise<PaginationResponseDto<UserEntity>> {
+		const { results, total } = await this.userModel
+			.query()
+			.page(--page, pageSize);
```
liza-veis, line 54: this should be done on service level. Also use `- PAGE_INDEX_OFFSET` instead of mutating parameters.
```suggestion
			.page(page, pageSize);
```

apps/frontend/src/modules/users/users-api.ts
```diff
@@ -21,15 +22,23 @@ class UserApi extends BaseHTTPApi {
...
-	public async getAll(): Promise<UserGetAllResponseDto> {
-		const response = await this.load(
-			this.getFullEndpoint(UsersApiPath.ROOT, {}),
+	public async getAll({
+		page,
+		pageSize,
+	}: PaginationQueryParameters): Promise<UserGetAllResponseDto> {
+		const end = this.getFullEndpoint(
+			UsersApiPath.ROOT,
+			"?page=:page&pageSize=:pageSize",
```
liza-veis, line 31: getFullEndpoint is for endpoint only (without query params). Pass query to load options. And something like private getUrl method to base http api and use `URLSearchParams` to build query string in it 
```suggestion
	public async getAll(query: PaginationQueryParameters): Promise<UserGetAllResponseDto> {
    const response = await this.load(
			this.getFullEndpoint(UsersApiPath.ROOT, {}),
			{
				contentType: ContentType.JSON,
				hasAuth: true,
				method: "GET",
				query
			},
```

liza-veis (changes requested):

apps/frontend/src/modules/users/users-api.ts
Vitaliistf: Done it. Thanks!

apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -1,3 +1,137 @@
+// import { ServerErrorType } from "~/libs/enums/enums.js";
```
liza-veis, line 1: I believe this comment should be removed 😅

apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -81,6 +215,17 @@ class BaseHTTPApi implements HTTPApi {
 		return headers;
 	}
 
+	private getUrl(path: string, query?: Record<string, string>): string {
```
liza-veis, line 218:
```suggestion
	private getUrl<T extends Record<string, unknown>>(path: string, query?: T): string {
```

apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -81,6 +215,17 @@ class BaseHTTPApi implements HTTPApi {
...
+		let fullPath = path;
+
+		if (query) {
+			const queryParameters = new URLSearchParams(query);
+			fullPath += `?${queryParameters.toString()}`;
+		}
+
+		return fullPath;
+	}
```
liza-veis, line 227: let's simplify this
```suggestion
		if(!query) {
			return path;
		}

		const queryParameters = new URLSearchParams(query);

		return `${path}?${queryParameters.toString()}`;
	}
```

apps/frontend/src/modules/users/users-api.ts
```diff
@@ -21,13 +22,19 @@ class UserApi extends BaseHTTPApi {
...
 		const response = await this.load(
 			this.getFullEndpoint(UsersApiPath.ROOT, {}),
 			{
 				contentType: ContentType.JSON,
 				hasAuth: true,
 				method: "GET",
+				query: {
+					page: String(query.page),
+					pageSize: String(query.pageSize),
+				},
```
liza-veis, line 37: I believe we don't need this after the change with typing above
```suggestion
				query
```

liza-veis (changes requested):

apps/frontend/src/libs/modules/api/base-http-api.ts
Vitaliistf: Ohhh, yeah, sure, sorry

apps/frontend/src/libs/modules/api/base-http-api.ts
Vitaliistf: I'm getting "Type parameter T is used only once in the function signature." now. So I guess I'll just use Record<string, unknown> as a type for query, it'll work the same.

Vitaliistf: I guess I'll just use Record<string, unknown> type for query, because linter does not like that we use generic only in method signature, and it would work the same.

apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -81,6 +81,18 @@ class BaseHTTPApi implements HTTPApi {
...
 	}
 
+	private getUrl(path: string, query?: Record<string, unknown>): string {
+		if (!query) {
+			return path;
+		}
+
+		const queryParameters = new URLSearchParams(
+			query as Record<string, string>,
+		);
```
liza-veis, line 91: Hm, looks like it's not allowed to pass a number, in this case the type above will be invalid. Let's use `ConstructorParameters<typeof URLSearchParams>` instead of `Record<string, unknown>`. And this case we will need to convert number query params to string (as it was before)
```suggestion
		const queryParameters = new URLSearchParams(
			query
		);
```

liza-veis (changes requested):

apps/frontend/src/libs/modules/api/base-http-api.ts
Vitaliistf: It’s very strange, but if I give `query` type of `ConstructorParameters<typeof URLSearchParams>` then I get an error: 
```
Argument of type '[init?: string | Record<string, string> | string[][] | URLSearchParams | undefined]' is not assignable to parameter of type 'string | Record<string, string> | string[][] | URLSearchParams | undefined'.
  Type '[init?: string | Record<string, string> | string[][] | URLSearchParams | undefined]' is not assignable to type 'string[][]'.
    Type 'string | Record<string, string> | string[][] | URLSearchParams | undefined' is not assignable to type 'string[]'.
      Type 'undefined' is not assignable to type 'string[]'.
```
So I need to use `ConstructorParameters<typeof URLSearchParams>[0]` (also, 0 is magic number there) as a type. Maybe it is better to leave `Record<string, string>`?

liza-veis: try to use ConstructorParameters<typeof URLSearchParams>[number]

liza-veis: We can't leave Record<string, string> because we may have nested fields and arrays

Vitaliistf: That works, thank you!

liza-veis (approved):

PR #148, opened 2024-09-02
fix: the size of the user popup and the font of the elements don't match the layout gf-136

Vitaliistf:
Changed font and set min width to 250px as in design.

### DIFF
--- a/apps/frontend/src/libs/components/header/libs/components/user-popover/styles.module.css
+++ b/apps/frontend/src/libs/components/header/libs/components/user-popover/styles.module.css
@@ -35,7 +35,7 @@
 .button {
 	width: 100%;
 	padding: 8px 12px;
-	font-family: Inter, sans-serif;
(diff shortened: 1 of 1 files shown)


apps/frontend/src/libs/components/header/libs/components/user-popover/styles.module.css
```diff
@@ -35,6 +35,7 @@
 .button {
 	width: 100%;
 	padding: 8px 12px;
+	font-family: Inter, sans-serif;
```
what1s1ove, line 38:
```suggestion
	font-family: inherit;
```

Vitaliistf: Fixed

yevhenii-hladkov (approved):

liza-veis (approved):

PR #153, opened 2024-09-02
fix: The Project description text overflows beyond the screen boundaries when 1000 characters are entered gf-150

GvoFor:
Text now wraps properly and no overflow
image


what1s1ove (approved):

liza-veis:
Don't we need the same for title?

GvoFor:
Yeah, you are right

liza-veis (approved):

PR #162, opened 2024-09-03
fix: Inconsistent location of the load component with a slow internet connection gf-147

AnnKasian:
Done: 

- centralized loader 

Preview:
image


apps/frontend/src/libs/components/protected-route/styles.module.css
```diff
@@ -0,0 +1,3 @@
+.loader {
+	height: 100vh;
+}
```
liza-veis, line 3:
```suggestion
.loader-container {
	min-height: 100vh;
}
```

liza-veis (changes requested):

liza-veis (approved):

PR #167, opened 2024-09-03
feat: Logo element on the Header component should redirect user to the Main page gf-154

AnnKasian:
Done: 
- Added navigation to root from logo

### DIFF
--- a/apps/frontend/src/libs/components/header/header.tsx
+++ b/apps/frontend/src/libs/components/header/header.tsx
@@ -1,5 +1,6 @@
 import logoSrc from "~/assets/images/logo.svg";
-import { Avatar } from "~/libs/components/components.js";
+import { Avatar, NavLink } from "~/libs/components/components.js";
+import { AppRoute } from "~/libs/enums/enums.js";
 import { useAppSelector } from "~/libs/hooks/hooks.js";
 
 import { UserPopover } from "./libs/components/components.js";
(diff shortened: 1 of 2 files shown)


liza-veis (approved):

yevhenii-hladkov (approved):

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -18,10 +19,12 @@ const Header = (): JSX.Element => {
 
 	return (
 		<header className={styles["header"]}>
-			<div className={styles["logo-container"]}>
-				<img alt="GitFit logo" className={styles["logo-img"]} src={logoSrc} />
-				<span className={styles["logo-text"]}>Logo</span>
-			</div>
+			<NavLink className={styles["logo-link"] ?? ""} to={AppRoute.ROOT}>
```
GvoFor, line 22: Maybe let's stick to the following approach
```suggestion
			<NavLink className={styles["logo-link"] as string} to={AppRoute.ROOT}>
```
Currently, in our project we have both `?? ""` and `as string`, so we should decide, which one to stick with. And `as string` looks more prefferable cause we know for sure, that this style is a string.

@liza-veis what do you think?

what1s1ove (approved):

apps/frontend/src/libs/components/header/header.tsx
what1s1ove: agreed, it will be better

PR #168, opened 2024-09-03
feat: Add an ability to delete a project gf-138

GvoFor:
## Summary
This PR adds an ability to delete a project.
1. There is ellipsis icon on the right side of the project card. When you click on this icon menu opens, where `delete` option present;
2. When you click on `delete` option, confirmation modal popups, where you can either cancel operation or confirm it;
3. If you cancel operation, modal closes. If you confirm, modal closes and project card disapear from the list.
4. Depending on successful or not was the operation, corresponding toast notification popups

## Screenshots
On ellipsis icon press
image

On delete option choose
image

On deletion confirm
image


apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -1,3 +1,4 @@
+import { NOTHING_DELETED_COUNT } from "~/libs/constants/constants.js";
```
kolibri753, line 1: Hey, you created a constant for this, while I defined a similar value directly in a function. Which approach do you think is better?

GvoFor: I think that global constant (my approach) will be more preferable, since it will be used not only in my and your issues, but in other related to ability to delete something, so this approach will prevent us of creating inplace `ZERO` constant everywhere

apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -27,8 +28,13 @@ class ProjectRepository implements Repository {
...
 
-	public delete(): ReturnType<Repository["delete"]> {
-		return Promise.resolve(true);
+	public async delete(id: number): Promise<boolean> {
+		const deletedRowsCount = await this.projectModel
+			.query()
+			.deleteById(id)
+			.execute();
+
+		return deletedRowsCount > NOTHING_DELETED_COUNT;
```
liza-veis, line 37: let's just wrap it with Boolean
```suggestion
		return Boolean(deletedRowsCount);
```

liza-veis (changes requested):

apps/frontend/src/libs/components/confirmation-modal/confirmation-modal.tsx
```diff
@@ -0,0 +1,49 @@
...
+	isOpened: boolean;
+	onClose: () => void;
+	onConfirm: () => void;
+	title: string;
+};
+
+const ConfirmationModal = ({
+	cancelLabel = "Cancel",
+	confirmationText,
+	confirmLabel = "Confirm",
```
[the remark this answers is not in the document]
GvoFor: What about title? Looks like it also won't be overriden. I'll remove it too for now

apps/backend/src/modules/projects/project.repository.ts
GvoFor: That's genious ;)

liza-veis (approved):

what1s1ove (approved):

apps/backend/src/modules/projects/project.controller.ts
```diff
@@ -124,6 +135,37 @@ class ProjectController extends BaseController {
...
+	 *      description: Deletes project by ID
+	 *      parameters:
+	 *        - in: path
+	 *          name: id
+	 *          schema:
+	 *            type: integer
+	 *          required: true
+	 *          description: Numeric ID of the project to be deleted
+	 *      responses:
+	 *        200:
```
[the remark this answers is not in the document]
liza-veis: we will return json payload, that's why we need 200

liza-veis (approved):

Issue #134, opened 2024-08-30
feat: Validation notification of the Password field on the login page


AnnaRomaniuta:
### What feature?

The message for the "Password" field should indicate that the password must be at least 5 characters long

AnnaRomaniuta:
We have the following requirements for the password field:

At least 5 characters ;
Not more than 50 characters
At least 1 uppercase letter (A-Z);
At least 1 lowercase letter (a-z);
At least 1 number ;
At least 1 special character (e.g., !@#$%)"

AnnaRomaniuta:
Good job! Thanks!

PR #173, opened 2024-09-03
feat: add only latin validation rule gf-131

kolibri753:
Changes:

- added only latin letters rule to all fields on auth pages;
- put all regex validation for one field in one rule.

Issues:

- [#131](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/issues/131)
- [#134](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/issues/134)


packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -7,6 +7,7 @@ const UserValidationMessage = {
 	EMAIL_WRONG: "Email is wrong.",
 	NAME_TOO_LONG: `Name is too long (> ${String(UserValidationRule.NAME_MAXIMUM_LENGTH)} symbols).`,
 	NAME_TOO_SHORT: `Name must be at least ${String(UserValidationRule.NAME_MINIMUM_LENGTH)} characters long.`,
+	ONLY_LATIN: "Latin letters only",
```
sandrvvu, line 10: make sure to include a period at the end of the sentence

yevhenii-hladkov (approved):

packages/shared/src/modules/users/libs/enums/user-validation-rule.enum.ts
```diff
@@ -1,6 +1,7 @@
 const UserValidationRule = {
 	EMAIL_MAXIMUM_LENGTH: 50,
 	EMAIL_MINIMUM_LENGTH: 6,
+	LATIN_PATTERN: /^[\w!"#$%&'()*+,./:;<=>?@[\]^`{|}~-]*$/,
```
liza-veis, line 4: It's not only latin and this validation looks a bit confusing and complex. It's better to avoid many regexes where it can be handled properly with one. Let's have one regext for password pattern like `/^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]*$/` (replace special chars starting from @ with needed ones). This pattern will require at least one lowercase, at least one uppercase, at least one digit, and at least one special char. The message for the password will need to combine all those rules.  And for the name field, let's have a separate pattern:
```javascript
NAME_PATTERN: /^[A-Za-z\s]*$/
```
For email, I think it's already restricted and there is no need to check it.

liza-veis (changes requested):

packages/shared/src/modules/users/libs/enums/user-validation-rule.enum.ts
kolibri753: Oh okay, it is clearer your way

P.s. it was more like Latin + special characters 😅

what1s1ove (approved):

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -7,15 +7,11 @@ const UserValidationMessage = {
 	EMAIL_WRONG: "Email is wrong.",
 	NAME_TOO_LONG: `Name is too long (> ${String(UserValidationRule.NAME_MAXIMUM_LENGTH)} symbols).`,
 	NAME_TOO_SHORT: `Name must be at least ${String(UserValidationRule.NAME_MINIMUM_LENGTH)} characters long.`,
-	PASSWORD_LOWERCASE_REQUIRED:
-		"Password must include at least one lowercase letter.",
+	ONLY_LATIN: "Latin letters only.",
```
liza-veis, line 10:
```suggestion
	NAME_PATTERN: "Name must contain only Latin letters.",
```

packages/shared/src/modules/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -7,15 +7,11 @@ const UserValidationMessage = {
...
+	PASSWORD_COMPLEXITY:
+		"Password must include at least one uppercase letter, one lowercase letter, one number, and one special character (Latin letters only).",
```
liza-veis, line 12:
```suggestion
	PASSWORD_PATTERN:
		"Password must contain only Latin letters and include at least one uppercase letter, one lowercase letter, one number, and one special character.",
```

liza-veis (changes requested):

liza-veis (approved):

PR #174, opened 2024-09-03
feat: scroll element must be shown at the right side of the table component gf-149

Vitaliistf:
Added scroll to the table. The header of the table is sticking at the top, while body is moving. 
Screenshot 2024-09-03 at 18 21 13

### DIFF
--- a/apps/frontend/src/libs/components/table/styles.module.css
+++ b/apps/frontend/src/libs/components/table/styles.module.css
@@ -1,3 +1,10 @@
+.table-container {
+	width: 100%;
+	overflow: hidden;
+	border: 1px solid var(--color-border-primary);
+	border-radius: 8px;
+}
+
 .table {
 	width: 100%;
 	font-family: Inter, sans-serif;
@@ -7,8 +14,6 @@
 	border-spacing: 0;
 	border-collapse: separate;
 	background-color: var(--color-background-secondary);
-	border: 1px solid var(--color-border-primary);
-	border-radius: 8px;
 }
 
 .table th,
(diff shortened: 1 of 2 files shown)


apps/frontend/src/libs/components/table/styles.module.css
sandrvvu, line 1: image
B10?

liza-veis: Yes, I agree, we need to pass classNames to tr, th etc.

apps/frontend/src/libs/components/table/styles.module.css
```diff
@@ -1,3 +1,10 @@
+.table-container {
+	width: 100%;
+	overflow: hidden;
```
liza-veis, line 3: Do we need overflow hidden on container if we already have scroll inside?

apps/frontend/src/libs/components/table/styles.module.css
Vitaliistf: Oh, okay then, I'll change it.

what1s1ove (approved):

apps/frontend/src/libs/components/table/styles.module.css
Vitaliistf: Yeah, it is for styling, because table has straight corners now, and some other things are getting out of borders. It's either `overflow: hidden` here, or a lot of additional styles everywhere else.

apps/frontend/src/libs/components/table/styles.module.css
Vitaliistf: Done it. Thanks for notice!

liza-veis (approved):

PR #175, opened 2024-09-03
feat: restrict users manage groups gf-84

AnutkaGn:
Only users with "Manage Access" permission should be able to view, create, edit or delete a group.
Hide the page for users without permission.
Add permissions validation for the backend to restrict this.


apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -2,18 +2,27 @@ import { createSlice, isAnyOf } from "@reduxjs/toolkit";
...
+	loadPermissions,
+	logout,
+	signIn,
+	signUp,
+} from "./actions.js";
 
 type State = {
 	authenticatedUser: null | UserAuthResponseDto;
 	dataStatus: ValueOf<typeof DataStatus>;
+	permissions: null | UserGetPermissionItemResponseDto[];
```
what1s1ove, line 19:
```suggestion
	permissions: UserGetPermissionItemResponseDto[];
```

apps/frontend/src/modules/auth/slices/auth.slice.ts
```diff
@@ -2,18 +2,27 @@ import { createSlice, isAnyOf } from "@reduxjs/toolkit";
...
 };
 
 const initialState: State = {
 	authenticatedUser: null,
 	dataStatus: DataStatus.IDLE,
+	permissions: null,
```
what1s1ove, line 25:
```suggestion
	permissions: [],
```

apps/backend/src/modules/users/user.service.ts
```diff
@@ -89,6 +90,12 @@ class UserService implements Service {
 		return item;
 	}
 
+	public async getPermissionsByUserId(
+		userId: number,
+	): Promise<null | UserGetPermissionItemResponseDto[]> {
```
what1s1ove, line 95:
```suggestion
	): Promise<UserGetPermissionItemResponseDto[]> {
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -53,6 +56,17 @@ class UserRepository implements Repository {
 		return user ? UserEntity.initialize(user) : null;
 	}
 
+	public async getPermissionsByUserId(
+		userId: number,
+	): Promise<null | UserGetPermissionItemResponseDto[]> {
```
what1s1ove, line 61:
```suggestion
	): Promise<UserGetPermissionItemResponseDto[]> {
```

apps/backend/src/modules/auth/auth.service.ts
```diff
@@ -34,6 +35,12 @@ class AuthService {
 		return await this.userService.find(userId);
 	}
 
+	public async getPermissions(
+		userId: number,
+	): Promise<null | UserGetPermissionItemResponseDto[]> {
```
what1s1ove, line 40:
```suggestion
	): Promise<UserGetPermissionItemResponseDto[]> {
```

what1s1ove (changes requested):
permissions should be a separate module 
(module, slice, etc)

apps/backend/src/libs/modules/controller/base-controller.module.ts
```diff
@@ -48,13 +48,14 @@ class BaseController implements Controller {
...
 	public addRoute(options: ControllerRouteParameters): void {
-		const { handler, path } = options;
+		const { handler, path, preHandler } = options;
 		const fullPath = this.apiUrl + path;
 
 		this.routes.push({
 			...options,
 			handler: (request, reply) => this.mapHandler(handler, request, reply),
 			path: fullPath,
+			...(preHandler && { preHandler }),
```
GvoFor, line 58: This line is redudent cause of `...options,`
```suggestion
```

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -166,17 +170,23 @@ class BaseServerApplication implements ServerApplication {
 	}
 
 	public addRoute(parameters: ServerApplicationRouteParameters): void {
-		const { handler, method, path, validation } = parameters;
+		const { handler, method, path, preHandler, validation } = parameters;
 
-		this.app.route({
+		const route: RouteOptions = {
```
GvoFor, line 175:
```suggestion
		const routeOptions: RouteOptions = {
```

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -8,9 +10,21 @@ type Properties = {
...
 
 const Sidebar = ({ items }: Properties): JSX.Element => {
+	const { permissions } = useAppSelector(({ auth }) => auth);
+
+	const filteredItems = items.filter((item) => {
+		if (item.label === "Access Management") {
+			return permissions?.some(
+				(permission) => permission.key === Permission.MANAGE_USER_ACCESS,
+			);
+		}
```
GvoFor, line 20: It'll be better to create some `map`, which will map page name to an array of requeired permissions. Also use (create if it's not exist) enum for pages names instead of hardcoded string literal (do not forget to use this enum in `navigation-items.constant.ts`)

packages/shared/src/modules/users/libs/types/user-get-permission-item-response-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+type UserGetPermissionItemResponseDto = {
+	id: number;
+	key: string;
```
GvoFor, line 3:
```suggestion
	key: Permisson;
```

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -166,17 +170,23 @@ class BaseServerApplication implements ServerApplication {
...
 			method,
 			schema: {
 				body: validation?.body,
 			},
 			url: path,
+		};
+
+		const routeOptionsWithPreHandler = Object.assign({}, route, {
+			...(preHandler ? { preHandler } : {}),
 		});
```
GvoFor, line 186: I think you can do just:
```ts
const routeOptions = {
  handler,
  method,
  shema: {
    body: validation?.body,
  },
  url: path,
  ...(preHandler ? { preHandler } : {}),
}
```

perekotypole:
https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/176 Serhii has done a lot with permission modules, you can check it as well

apps/backend/src/modules/auth/auth.controller.ts
```diff
@@ -87,6 +99,46 @@ class AuthController extends BaseController {
...
+	 *                      format: date-time
+	 */
+
+	private async getPermissions(
+		options: APIHandlerOptions<{
+			user: UserAuthResponseDto;
+		}>,
+	): Promise<APIHandlerResponse> {
+		return {
+			payload: await this.authService.getPermissions(options.user.id),
```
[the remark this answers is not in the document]
AnutkaGn: Yes, I agree, it is being worked on now

what1s1ove (changes requested):
permissions should be a separate module
(module, slice, etc)

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -26,11 +41,15 @@ const ProtectedRoute = ({ children }: Properties): JSX.Element => {
 		);
 	}
 
-	return hasAuthenticatedUser ? (
-		<>{children}</>
-	) : (
-		<Navigate replace to={AppRoute.SIGN_IN} />
-	);
+	if (!hasRequiredPermission) {
+		return <Navigate replace to={AppRoute.ANY} />;
```
yevhenii-hladkov, line 45: If I get it right when user doesn't have the required permission (status 403 forbidden) then the user is redirected to ANY route which displays not found page with status 404.

AnutkaGn: Yes, that's right. I don't know what better error to return, since 403 is correct because access is forbidden. I can change to 404 not Found, but I'm not sure that this is a good decision

apps/backend/src/libs/helpers/check-permission.ts
```diff
@@ -0,0 +1,42 @@
...
+		const userId = request.user?.id;
+
+		if (!userId) {
+			throw new HTTPError({
+				message: ExceptionMessage.USER_NOT_FOUND,
+				status: HTTPCode.UNAUTHORIZED,
+			});
+		}
+
+		const permissions = await permissionService.getPermissions(userId);
```
liza-veis, line 20: I think there was a misunderstanding with permissions in the app. We need the permissions service for getting and manipulating all the permissions (for example, for the permissions dropdown when we create a group). But for permissions for the current user, we need to attach them to the current user through the groups. So that UserAuthResponseDto contains groups field with permissions which we need to include using relation in users getById method. We don't need a separate call for user permissions, we will use get permissions for getting all available in db.

liza-veis (changes requested):

apps/backend/src/libs/helpers/check-permission.ts
liza-veis: For this task, we need to attach permissions to the authenticated user itself. So instead of loading them separately, you already will have groups and permissions inside authenticated user. You need to include them using relations in one query (not a separate call) in getById method of users service. We don't need service/api/slice to get all permissions in this task.

AnutkaGn: If I understand correctly, then the UserAuthResponseDto will look like this:
```
type UserAuthResponseDto = {
	createdAt: string;
	email: string;
	id: number;
	name: string;
	groups: [
		{
			groupName: string;
			permissions: [
				{
					name: string;
					key: string;
				}
			]
		}
	]
};
```

liza-veis: yes, similar to this

apps/frontend/src/modules/permissions/slices/permission.slice.ts
```diff
@@ -0,0 +1,39 @@
...
+	dataStatus: DataStatus.IDLE,
+	permissions: {
+		items: [],
+	},
+};
+
+const { actions, name, reducer } = createSlice({
+	extraReducers(builder) {
+		builder.addCase(loadPermissions.fulfilled, (state, action) => {
+			state.permissions = action.payload;
```
what1s1ove, line 24:
```suggestion
			state.permissions = action.payload.items;
```

apps/frontend/src/modules/permissions/slices/permission.slice.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { DataStatus } from "~/libs/enums/enums.js";
+import { type ValueOf } from "~/libs/types/types.js";
+
+import { type PermissionGetAllResponseDto } from "../libs/types/types.js";
+import { loadPermissions } from "./actions.js";
+
+type State = {
+	dataStatus: ValueOf<typeof DataStatus>;
+	permissions: PermissionGetAllResponseDto;
```
what1s1ove, line 11:
```suggestion
	permissions: ItemType[];
```

apps/frontend/src/modules/permissions/slices/permission.slice.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const initialState: State = {
+	dataStatus: DataStatus.IDLE,
+	permissions: {
+		items: [],
+	},
```
what1s1ove, line 18:
```suggestion
	permissions: []
```

apps/frontend/src/libs/constants/page-name.constant.ts
```diff
@@ -0,0 +1,6 @@
+const PAGE_NAME = {
```
what1s1ove, line 1: seems like an enum

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -1,23 +1,39 @@
...
 import { useAppSelector } from "~/libs/hooks/hooks.js";
+import { type PermissionGetAllItemResponseDto } from "~/modules/permissions/libs/types/types.js";
+import { NotFound } from "~/pages/not-found/not-found.jsx";
 
 import styles from "./styles.module.css";
 
 type Properties = {
-	children: ReactNode;
+	children: React.ReactNode;
+	pagePermissions?: string[];
```
what1s1ove, line 13: should be typed values here (as an enum's values, i guess)

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -1,23 +1,39 @@
...
+}: Properties): JSX.Element => {
 	const { authenticatedUser, dataStatus } = useAppSelector(({ auth }) => auth);
+	const { permissions } = useAppSelector(({ permissions }) => permissions);
 
 	const hasAuthenticatedUser = Boolean(authenticatedUser);
+
 	const isLoading =
 		dataStatus === DataStatus.PENDING || dataStatus === DataStatus.IDLE;
 
+	const hasRequiredPermission = pagePermissions
```
what1s1ove, line 28: i think we need to create a helper for it and use it on both back/frond apps

liza-veis (changes requested):

what1s1ove (approved):
lgtm after Lisa's comments

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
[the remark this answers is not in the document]
liza-veis: It's how usually pre-handlers are called on backend. We don't need to name it with use because it's not a frontend hook

apps/backend/src/libs/modules/controller/libs/types/controller-route-parameters.type.ts
```diff
@@ -2,11 +2,13 @@ import { type HTTPMethod } from "~/libs/modules/http/http.js";
 import { type ValidationSchema } from "~/libs/types/types.js";
 
 import { type APIHandler } from "./api-handler.type.js";
+import { type APIPreHandler } from "./api-pre-handler.type.js";
 
 type ControllerRouteParameters = {
 	handler: APIHandler;
 	method: HTTPMethod;
 	path: string;
+	preHandler?: APIPreHandler;
```
liza-veis, line 11: we need to be able to pass multiple hooks here
```suggestion
	preHandlers?: APIPreHandler[];
```

apps/backend/src/modules/users/user.entity.ts
```diff
@@ -1,26 +1,28 @@
...
-
 	private deletedAt: null | string;
-
 	private email: string;
-
+	private groups: Array<
+		{
+			permissions: Array<Pick<PermissionModel, "key" | "name">>;
+		} & Pick<GroupModel, "name">
+	>;
```
liza-veis, line 15: I see that in other places we use model in entity, i think we will need to change it to entity. this way we can do it like this
```suggestion
	private groups: Pick<GroupEntity, "name" | "permissions">[]
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -47,10 +49,22 @@ class UserRepository implements Repository {
...
 			.whereNull("deletedAt")
-			.returning("*")
-			.execute();
+			.withGraphFetched("groups.permissions")
+			.modifyGraph("groups", (builder) => {
+				builder.select("name");
+			})
+			.modifyGraph("groups.permissions", (builder) => {
+				builder.select("name", "key");
+			})
```
liza-veis, line 58: we need to include group id and permission id, just best practice to return identifier for all resources

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -47,10 +49,22 @@ class UserRepository implements Repository {
...
+			.castTo<
+				{
+					deletedAt: string;
+					passwordHash: string;
+					passwordSalt: string;
+				} & UserAuthResponseDto
```
liza-veis, line 64: Can't we infer this type from initialize parameters?

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -47,10 +49,22 @@ class UserRepository implements Repository {
...
+			>();
 
-		return user ? UserEntity.initialize(user) : null;
+		return UserEntity.initialize(user);
```
liza-veis, line 67: what if we didn't find a user

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -26,11 +37,22 @@ const ProtectedRoute = ({ children }: Properties): JSX.Element => {
...
-		<>{children}</>
-	) : (
-		<Navigate replace to={AppRoute.SIGN_IN} />
-	);
+	if (!hasAuthenticatedUser) {
+		return <Navigate replace to={AppRoute.SIGN_IN} />;
+	}
+
+	const hasRequiredPermission = authenticatedUser
+		? checkHasPermission(
```
liza-veis, line 45: this check is no longer needed
```suggestion
	const hasRequiredPermission = checkHasPermission(
```

apps/frontend/src/libs/constants/navigation-items.constant.ts
```diff
@@ -11,6 +11,7 @@ const SIDEBAR_ITEMS: NavigationItem[] = [
 		href: AppRoute.ACCESS_MANAGEMENT,
 		icon: "access",
 		label: "Access Management",
+		pagePermissions: ["manage_user_access"],
```
liza-veis, line 14:
```suggestion
		pagePermissions: [PermissionKey.MANAGE_USER_ACCESS],
```

apps/frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -1,16 +1,32 @@
...
+			if (item.pagePermissions) {
+				return item.pagePermissions.every((permission) =>
+					userPermissions.some(
+						(userPermission) => userPermission.key === permission,
+					),
+				);
+			}
+
+			return true;
+		});
```
liza-veis, line 24: let's simplify it
```suggestion
		return items.filter(({ pagePermission = [] }) => checkHasPermission(pagePermissions, userPermissions));
```

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -1,20 +1,31 @@
...
 type Properties = {
-	children: ReactNode;
+	children: React.ReactNode;
+	permissionKeys?: ValueOf<typeof PermissionKey>[];
 };
 
-const ProtectedRoute = ({ children }: Properties): JSX.Element => {
+const ProtectedRoute = ({
+	children,
+	permissionKeys = [],
```
liza-veis, line 23: let's name it by role
```suggestion
	routePermissions = [],
```

packages/shared/src/libs/helpers/check-has-permission/check-has-permission.helper.ts
```diff
@@ -0,0 +1,10 @@
+const checkHasPermission = (
+	requiredPermissions: string[],
+	permissions: { key: string; name: string }[],
```
liza-veis, line 3: user PermissionGetAllItem type here

liza-veis (changes requested):

apps/backend/src/modules/users/user.repository.ts
AnutkaGn: I wanted to leave it, but the linter gave the error `Unnecessary conditional, value is always truthful.`

apps/frontend/src/libs/components/protected-route/protected-route.tsx
AnutkaGn: I can't do that because I get a `Forbidden non-null assertion` error

apps/backend/src/modules/users/user.entity.ts
AnutkaGn: It doesn't work, I tried to do it that way, it doesn't work.

apps/backend/src/modules/users/user.repository.ts
liza-veis: we can't leave it as it is now, because this will cause unexpected application error

liza-veis: The problem is that you cast the return query type to be found user so linter thinks it is always be found

apps/backend/src/libs/modules/controller/libs/types/controller-route-parameters.type.ts
```diff
@@ -2,11 +2,13 @@ import { type HTTPMethod } from "~/libs/modules/http/http.js";
 import { type ValidationSchema } from "~/libs/types/types.js";
 
 import { type APIHandler } from "./api-handler.type.js";
+import { type APIPreHandler } from "./api-pre-handler.type.js";
 
 type ControllerRouteParameters = {
 	handler: APIHandler;
 	method: HTTPMethod;
 	path: string;
+	preHandler?: APIPreHandler[];
```
liza-veis, line 11: we our property we always know it will be an array, so it should be plural
```suggestion
	preHandlers?: APIPreHandler[];
```

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -166,16 +166,19 @@ class BaseServerApplication implements ServerApplication {
...
-		this.app.route({
+		const routeOptions = {
 			handler,
 			method,
 			schema: {
 				body: validation?.body,
 			},
 			url: path,
-		});
+			...(preHandler ? { preHandler } : {}),
```
liza-veis, line 178: now it's array so we can do it simpler
```suggestion
			preHandler: preHandlers ?? [],
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -47,10 +49,22 @@ class UserRepository implements Repository {
...
-			.returning("*")
-			.execute();
+			.withGraphFetched("groups.permissions")
+			.modifyGraph("groups", (builder) => {
+				builder.select("user_groups.id", "name");
+			})
+			.modifyGraph("groups.permissions", (builder) => {
+				builder.select("permissions.id", "name", "key");
+			})
+			.castTo<
```
liza-veis, line 59: I've just checked it locally and we don't need this castTo at all, we can just add .execute() at the end

liza-veis (changes requested):

apps/backend/src/modules/users/user.repository.ts
AnutkaGn: Oh, thanks a lot for the help, I don't know why I didn't try it right away

apps/backend/src/libs/modules/server-application/base-server-application.ts
```diff
@@ -166,16 +166,19 @@ class BaseServerApplication implements ServerApplication {
...
 
 	public addRoute(parameters: ServerApplicationRouteParameters): void {
-		const { handler, method, path, validation } = parameters;
+		const { handler, method, path, preHandlers, validation } = parameters;
 
-		this.app.route({
+		const routeOptions = {
 			handler,
 			method,
+			preHandlers: preHandlers ?? [],
```
liza-veis, line 174: I think fastify accepts preHandler only property, preHandlers is like our internal name, because we know it will be always an array
```suggestion
			preHandler: preHandlers ?? [],
```

apps/backend/src/libs/modules/server-application/libs/types/server-application-route-parameters.type.ts
```diff
@@ -1,5 +1,6 @@
 import { type FastifyReply, type FastifyRequest } from "fastify";
 
+import { type APIPreHandler } from "~/libs/modules/controller/libs/types/api-pre-handler.type.js";
```
liza-veis, line 3: this type shoud be re-exported from `~/libs/modules/controller/controller.js`, we can't access module internal files

liza-veis (changes requested):

yevhenii-hladkov (approved):

liza-veis (approved):

PR #176, opened 2024-09-03
feat: add an ablity to create a group gf-79

s1rserg:
* Added a "Create New" button near the Groups table on the Access Management page.
* When the button is clicked, a modal to create a new group should be displayed.
* Updated table component to have optional checkboxes; select component to handle changes.
* Implemented backend and frontend route handling for getAll permissions.
* After group creation, the new group appears at the top of the Groups table.
* Updated users backend module to fetch groups with user.
* Added groups column to the users table.
* Added pagination to a modal users table.
image
image
image


apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -44,13 +74,28 @@ const AccessManagement = (): JSX.Element => {
...
 			</section>
 			<section>
-				<h2 className={styles["section-title"]}>Groups</h2>
+				<div className={styles["section-header"]}>
+					<h2 className={styles["section-title"]}>Groups</h2>
+					<Button label="Create New" onClick={onModalOpen} />
+				</div>
 				<Table<GroupRow> columns={groupColumns} data={groupData} />
 			</section>
+			<Modal
```
perekotypole, line 89: I don't know, but I have separated users and groups tables  into components and put modal there locally for groups. It's just an idea, I'm not sure about it, but you can think about it too.

s1rserg: Yeah, it seems pretty logical. I will separate them. Thanks!

kolibri753:
Good job on separating permissions from groups! :+1: 

P.S. Sorry I didn't do that when creating the groups module. I thought it would add too much code for one pull request.

s1rserg:
I need to somehow separate fetching users for access management page and for create group modal. My solution is just to create another action and another users in the store for this action. I am not sure this is right. Maybe there is a more appropriate way to do this?

liza-veis:
> I need to somehow separate fetching users for access management page and for create group modal. My solution is just to create another action and another users in the store for this action. I am not sure this is right. Maybe there is a more appropriate way to do this?

Yes, this will be ok, something like configureGroupUsers (because we will be using it for creation and edit)

> upd. I just noticed that I can avoid creating duplicate action just by using some flag, but I cannot remove duplicate modalUsers from the store

Let's use separate action, we may have different logic and different endpoints

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -48,6 +50,39 @@ const Select = <TFieldValues extends FieldValues, TOptionValue>({
 		isLabelHidden && "visually-hidden",
 	);
 
+	const handleChange = useCallback(
+		(
+			selectedOptions:
```
perekotypole, line 55: How about making it easier and working only with values? And I did it without `onChange`, it can still work correctly
```
	const handleChange = useCallback(
		(
			selectedOptions:
				| MultiValue<SelectOption<TOptionValue> | undefined>
				| SingleValue<SelectOption<TOptionValue> | undefined>,
		): void => {
			const selectedValues = isMulti
				? (selectedOptions as MultiValue<SelectOption<TOptionValue>>).map(
						({ value }) => value,
					)
				: (selectedOptions as SingleValue<SelectOption<TOptionValue>>)?.value;

			field.onChange(selectedValues);
		},
		[field, isMulti],
	);

```

but for this we need to do changes in table pagination:
```
type FormData = {
	pageSize: number;
};
````

Comment (s1rserg) [reply apps/frontend/src/libs/components/select/select.tsx:53]: It was really overcomplicated, but this change in table pagination solves the issue. Thanks!

File: apps/frontend/src/libs/components/select/select.tsx [hunk 3]
```diff
@@ -48,6 +50,40 @@ const Select = <TFieldValues extends FieldValues, TOptionValue>({
...
+
+				if (onChange) {
+					onChange(filteredOptions as SelectOption<TOptionValue>[]);
+				}
+			} else {
+				field.onChange(selectedOptions);
+			}
+		},
+		[field, onChange, isMulti],
+	);
```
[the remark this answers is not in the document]
s1rserg: After previous changes the only thing I can separate from this handler is 
`const selectedValues = isMulti
				? (selectedOptions as MultiValue<SelectOption<TOptionValue>>).map(
						({ value }) => value,
					)
				: (selectedOptions as SingleValue<SelectOption<TOptionValue>>)?.value;`
And I'm not sure it is worth hiding.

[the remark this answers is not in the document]
Fjortis: You're right, not worth it at all. Agree with you totally.

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -43,6 +62,20 @@ const Table = <T extends object>({
 				<tbody className={styles["table-body"]}>
 					{table.getRowModel().rows.map((row) => (
 						<tr className={styles["table-row"]} key={row.id}>
+							{onRowSelect && (
+								<td className={styles["table-data"]}>
```
liza-veis, line 66: Create a separate subcomponent - SelectRowCell

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -43,6 +62,20 @@ const Table = <T extends object>({
...
+									<input
```
liza-veis, line 67: We need a separate common component for checkbox and it should be controlled using useForm, similar to search field

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -43,6 +62,20 @@ const Table = <T extends object>({
...
+										checked={selectedIds?.has(
+											(row.original as unknown as Identifiable).id,
```
liza-veis, line 69: we can't be sure it will be id. We need to pass a property like `rowIdKey` - "id" or `getRowId` - a function that accepts a row and returns id.

apps/frontend/src/libs/enums/notification-message.enum.ts
```diff
@@ -1,4 +1,5 @@
 const NotificationMessage = {
+	GROUP_CREATE_SUCCESS: "Group was successfully created",
```
liza-veis, line 2: End messages with a period
```suggestion
	GROUP_CREATE_SUCCESS: "Group was successfully created.",
```

apps/frontend/src/modules/users/slices/users.slice.ts
```diff
@@ -4,34 +4,64 @@ import { DataStatus } from "~/libs/enums/enums.js";
 import { type ValueOf } from "~/libs/types/types.js";
 import { type UserGetAllItemResponseDto } from "~/modules/users/users.js";
 
-import { loadAll, updateProfile } from "./actions.js";
+import { configureGroupUsers, loadAll, updateProfile } from "./actions.js";
 
 type State = {
-	dataStatus: ValueOf<typeof DataStatus>;
+	groupUsers: {
```
liza-veis, line 10: I think we can keep users for group in groups slice as just users

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -1,54 +1,73 @@
...
+	const handleGroupCreateSubmit = useCallback(
+		(payload: GroupCreateRequestDto): void => {
+			dispatch(groupActions.create(payload))
+				.then(() => {
+					onModalClose();
+					void dispatch(
+						userActions.loadAll({ page: userPage, pageSize: userPageSize }),
+					);
+				})
+				.catch(() => {});
```
liza-veis, line 67: We shouldn't await dispatch. `userActions.loadAll` can be dispatched inside `groupActions.create`. For modal flow, take a look at how it's done with status during project creation
```suggestion
```

apps/frontend/src/modules/users/slices/users.slice.ts
```diff
@@ -4,34 +4,64 @@ import { DataStatus } from "~/libs/enums/enums.js";
...
+		totalCount: 0,
+	},
 	updateProfileStatus: DataStatus.IDLE,
-	users: [],
-	usersTotalCount: 0,
+	users: {
+		data: [],
+		dataStatus: DataStatus.IDLE,
+		totalCount: 0,
+	},
```
liza-veis, line 34: reducers should have plain structure, let's return logic with users here back, and move groupUsers as `users` to groups slice because it's related to group creation

liza-veis (changes requested):

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -48,6 +48,34 @@ const Select = <TFieldValues extends FieldValues, TOptionValue>({
 		isLabelHidden && "visually-hidden",
 	);
 
+	const handleChange = useCallback(
```
liza-veis, line 51: We need to return changes in select component back, it should be controlled using useForm

liza-veis (changes requested):

apps/frontend/src/libs/components/select/select.tsx
s1rserg: Could you please explain a bit more what should be here? I added handleChange just so select would add only value to formData fields, not label value object.

liza-veis: Oh, I see, I was thinking you added custom handlers. in this case we need to simplify this logic, because it's not very readable. Something like:
```typescript
	const findOptionByValue = useCallback(
		(value: TOptionValue): SelectOption<TOptionValue> | undefined => {
			return options.find((option) => option.value === value);
		},
		[options],
	);
	
	const selectedValue = useMemo(() => {
		if (isMulti) {
			return (field.value as TOptionValue[])
				.map(findOptionByValue)
				.filter(Boolean);
		}

		return findOptionByValue(field.value as TOptionValue);
	}, []);
```

And for handleChange, can't we use `let` with if instead of single multiline condition? or linter doesn't allow this?

s1rserg: Simplified the logic. For handleChange linter allows only ternary expression.

apps/frontend/src/libs/components/table/table.tsx
s1rserg: I've been working on it for quite some time and I really cannot figure this out. 
So table should have SelectRowCell and inside SelectRowCell should be Checkbox. Can you please explain this more?Because now for me it seems redundant.
Also, if i use useFormController in Checkbox component, then should the useAppForm for that be in Table component? And how do I handle multiple checkboxes as a form? I believe every checkbox should have its unique name to use in useController but how do I watch all them like in search component.
Sorry for too many questions, but I'm really stuck.

apps/frontend/src/pages/access-management/access-management.tsx
s1rserg: So this is what's needed?

` const create = createAsyncThunk<
	GroupCreateResponseDto,
	{ payload: GroupCreateRequestDto; query: PaginationQueryParameters },
	AsyncThunkConfig
>(${sliceName}/create, async ({ payload, query }, { dispatch, extra }) => {
	const { groupApi, toastNotifier } = extra;

	const response = await groupApi.create(payload);

	toastNotifier.showSuccess(NotificationMessage.GROUP_CREATE_SUCCESS);

	void dispatch(userActions.loadAll(query));

	return response;
}); `

liza-veis: yes

apps/frontend/src/libs/components/table/table.tsx
s1rserg: @liza-veis, please check the comment above🙏

liza-veis: > So table should have SelectRowCell and inside SelectRowCell should be Checkbox. Can you please explain this more?Because now for me it seems redundant.

Because if we add useForm, we may need to pass handlers or additional logic, etc., and to avoid overcomplicating the table component, it's better to move this to a separate component (not common but internal to the table component). This is a special cell, and for all other special cells, we will create a custom component, that's also why.

> Also, if i use useFormController in Checkbox component, then should the useAppForm for that be in Table component? And how do I handle multiple checkboxes as a form? I believe every checkbox should have its unique name to use in useController but how do I watch all them like in search component.
Sorry for too many questions, but I'm really stuck.

1. let's create useSelectedItems hook to handle checkboxes selection, something like  
```tsx
const useSelectedItems = <T>(items: T[]): UseSelectedItemsHook<T> => {
  const [selectedItems, setSelectedItems] = useState<T[]>(items);

  const handleAdd = (id: T): void => {
    setSelectedItems((prevState: T[]) => prevState.concat(id));
  };

  const handleRemove = (id: T): void => {
    setSelectedItems((prevState: T[]) => prevState.filter((it) => it !== id));
  };

  const handleToggle = (id: T): void => {
    selectedItems.includes(id) ? handleRemove(id) : handleAdd(id);
  };

  return { onToggle: handleToggle, items: selectedItems };
};
```
items is our selected items, onToggle is our onRowSelect

2. then if we have this select cell as separate component, we will need to pass onToggle and isChecked to it. And inside it we can have something like this
```
  const { control, errors, reset } = useAppForm({
    defaultValues: {
      [name]: isChecked,
    },
  });

  useEffect(() => {
    reset({ // or set value
      [name]: isChecked,
    });
  }, [name, isChecked]);
```
now we will be able to pass control, errors, and name to checkbox

liza-veis: the name that we need to somehow pass here should be our userIds

apps/frontend/src/pages/access-management/components/group-create-form/group-create-form.tsx
```diff
@@ -0,0 +1,142 @@
...
+		<form className={styles["form-wrapper"]} onSubmit={handleFormSubmit}>
+			<Input
+				autoComplete="given-name"
+				control={control}
+				errors={errors}
+				label="Name"
+				name="name"
+			/>
+			<span className={styles["table-title"]}>Users</span>
+			<UsersTable
```
perekotypole, line 113: How about creating separate `GroupUsersTable` component and manage loading data and pagination there?

apps/frontend/src/libs/components/table/table.tsx
s1rserg: Thank you so much for your assistance! I hope the changes I've made are correct.

apps/backend/src/modules/permissions/permission.entity.ts
```diff
@@ -0,0 +1,45 @@
+import { type PermissionGetAllItemResponseDto } from "@git-fit/shared";
```
Fjortis, line 1: You've created a local types.ts for re-exports, you should use it here

s1rserg: Good catch! Thanks!

apps/frontend/src/pages/access-management/components/group-create-form/group-create-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+		[handleSubmit, onSubmit],
+	);
+
+	const permissionOptions = getPermissionOptions(permissions);
+
+	const isLoading =
+		permissionsDataStatus === DataStatus.IDLE ||
+		permissionsDataStatus === DataStatus.PENDING;
+
+	const { items: selectedUserIds, onToggle: handleToggle } =
```
liza-veis, line 61: let's name it consistently
```suggestion
	const { items: selectedUserIds, onToggle: handleUserIdsToggle } =
```

apps/frontend/src/pages/access-management/components/group-users-table/group-users-table.tsx
```diff
@@ -0,0 +1,83 @@
...
+				onPageSizeChange={onPageSizeChange}
+				onRowSelect={onToggle}
+				page={page}
+				pageSize={pageSize}
+				paginationBackground="secondary"
+				selectedIds={selectedUserIds}
+				totalItemsCount={usersTotalCount}
+				users={users}
+			/>
+			{selectedUserIds.length === EMPTY_LENGTH && errors.userIds && (
```
liza-veis, line 76: why not to check for error message only? (the same logic as for inputs)
```tsx
	const error = errors["userIds"]?.message;
	const hasError = Boolean(error);
```
```suggestion
			{hasError && (
```

liza-veis (changes requested):

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -6,16 +6,25 @@ import {
...
+type Properties<T> = {
 	columns: TableColumn<T>[];
 	data: T[];
+	getRowId?: (row: T) => number;
+	name?: string;
+	onRowSelect?: (rowId: number) => void;
+	selectedIds?: number[];
 };
 
 const Table = <T extends object>({
```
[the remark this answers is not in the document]
s1rserg: Check [this](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/176#discussion_r1745551711)

apps/frontend/src/libs/components/table/components/select-row-cell/select-row-cell.tsx
```diff
@@ -0,0 +1,43 @@
...
+		name,
+	});
+
+	useEffect(() => {
+		if (watchedValue !== isChecked) {
+			onToggle(id);
+		}
+	}, [watchedValue, isChecked, id, onToggle]);
+
+	return <Checkbox control={control} errors={errors} name={name} />;
```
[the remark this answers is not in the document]
s1rserg: I get this warning in this case
image

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -22,11 +23,12 @@ type Properties<T extends FieldValues> = {
 	placeholder?: string;
 	rightIcon?: JSX.Element;
 	rowsCount?: number;
-	type?: "email" | "password" | "search" | "text";
+	type?: "checkbox" | "email" | "password" | "search" | "text";
```
what1s1ove, line 26:
```suggestion
	type?: "email" | "password" | "search" | "text";
```

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -54,6 +54,21 @@
 	resize: none;
 }
 
+.input-checkbox {
```
what1s1ove, line 57:
```suggestion
```

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -48,6 +48,40 @@ const Select = <TFieldValues extends FieldValues, TOptionValue>({
...
+						({ value }) => value,
+					)
+				: (selectedOptions as SingleValue<SelectOption<TOptionValue>>)?.value;
+
+			field.onChange(selectedValues);
+		},
+		[field, isMulti],
+	);
+
+	const findOptionByValue = useCallback(
```
what1s1ove, line 68: a10

apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -48,6 +48,40 @@ const Select = <TFieldValues extends FieldValues, TOptionValue>({
...
+		(value: TOptionValue): SelectOption<TOptionValue> | undefined => {
+			return options.find((option) => option.value === value);
+		},
+		[options],
+	);
+
+	const selectedValue = useMemo(() => {
```
what1s1ove, line 75: a10

apps/frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
```diff
@@ -43,13 +45,15 @@ const useAppForm = <T extends FieldValues = FieldValues>({
...
 		handleSubmit,
+		setValue,
 	} = useForm<T>(parameters);
 
 	return {
 		control,
 		errors,
 		handleSubmit,
 		isDirty,
+		setValue,
```
what1s1ove, line 56:
```suggestion
		handleValueSet: setValue,
```

what1s1ove (commented):
merge the main into your branch

apps/frontend/src/libs/components/table/components/select-row-cell/select-row-cell.tsx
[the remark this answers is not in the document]
liza-veis: oh ok got this, then let's leave it

apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -12,6 +12,7 @@ import styles from "./styles.module.css";
 
 type Properties<T extends FieldValues> = {
 	autoComplete?: string;
+	checked?: boolean;
```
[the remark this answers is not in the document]
s1rserg: I won't use Input component

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -80,31 +89,39 @@ const AccessManagement = (): JSX.Element => {
...
-				<div className={styles["groups-table"]}>
-					<Table<GroupRow> columns={groupColumns} data={groupData} />
-					<TablePagination
-						onPageChange={onGroupPageChange}
-						onPageSizeChange={onGroupPageSizeChange}
-						page={groupPage}
-						pageSize={groupPageSize}
-						totalItemsCount={groupsTotalCount}
-					/>
+				<div className={styles["section-header"]}>
```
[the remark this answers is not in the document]
s1rserg: Yes. There is the rule
`.section-header {
	display: flex;
	align-items: center;
	justify-content: space-between;
	padding: 32px 0 24px;
}`

apps/frontend/src/pages/access-management/components/group-users-table/group-users-table.tsx
```diff
@@ -0,0 +1,83 @@
...
+	setValue,
+}: Properties): JSX.Element => {
+	const dispatch = useAppDispatch();
+
+	const { users, usersDataStatus, usersTotalCount } = useAppSelector(
+		({ groups }) => groups,
+	);
+
+	const { onPageChange, onPageSizeChange, page, pageSize } = usePagination({
+		queryParameterPrefix: "user",
```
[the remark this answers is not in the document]
s1rserg: It didn't conflict but I changed it

apps/frontend/src/pages/access-management/components/group-users-table/group-users-table.tsx
s1rserg: Errors are refreshed on submit of the form and If we keep track of selectedUserIds.length it allows to not display error message when after unsuccessful form submit we select some user from the table. I thought it will be more responsible this way.

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -43,6 +53,16 @@ const Table = <T extends object>({
 				<tbody className={styles["table-body"]}>
 					{table.getRowModel().rows.map((row) => (
 						<tr className={styles["table-row"]} key={row.id}>
+							{onRowSelect && selectedIds && name && getRowId && (
+								<td className={styles["table-data"]}>
+									<SelectRowCell
+										id={getRowId(row.original)}
+										isChecked={selectedIds.includes(getRowId(row.original))}
+										name={name}
+										onToggle={onRowSelect}
```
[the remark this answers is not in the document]
s1rserg: Can you explain what you mean by this, please?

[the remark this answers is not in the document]
perekotypole: I was just thinking about the ability of the table component to return a whole array of selected rows, not just row by row and to process it outside each time we need to. As an example of implementation, we can add `useSelectedItems` to the table component and replace `onRowSelect: (id: number) => void` with `onSelect: (selectedItems: number[]) => void`.

I'm not quite sure about that, it's just my logic of how it should work. But on the other hand, we will have to change a lot of things if we need some specific data processing in the future

s1rserg:
@liza-veis, there seems to be an issue introduced with the creation of preHandler on create groups route in backend. Pre handler executes properly, but handler doesn't execute. On the frontend it is just pending and on backend there is no log of request. 

With empty array like that everything is working properly
```
this.addRoute({
			handler: (options) =>
				this.createGroup(
					options as APIHandlerOptions<{ body: GroupCreateRequestDto }>,
				),
			method: "POST",
			path: GroupsApiPath.ROOT,
			preHandlers: [],
			validation: {
				body: groupCreateValidationSchema,
			},
		}); 
```

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -7,16 +7,28 @@ import {
 import { EMPTY_LENGTH } from "~/libs/constants/constants.js";
 import { type TableColumn } from "~/libs/types/types.js";
 
+import { SelectRowCell } from "./components/select-row-cell/select-row-cell.js";
```
perekotypole, line 10:
```suggestion
import { SelectRowCell } from "./components/components.js";
```

apps/frontend/src/libs/components/table/table.tsx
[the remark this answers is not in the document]
s1rserg: Your idea makes sense, but I'm not sure it's needed right now. As you said it might require more changes later if we need specific data processing.

apps/frontend/src/libs/components/table/table.tsx
s1rserg: SelectRowCell is a subcomponent of Table. There is no need to export it. Check [this](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/176#discussion_r1745533946)

apps/frontend/src/libs/components/table/components/select-row-cell/select-row-cell.tsx
```diff
@@ -0,0 +1,41 @@
+import { useEffect } from "react";
+
+import { Checkbox } from "~/libs/components/components.js";
+import { useAppForm, useFormWatch } from "~/libs/hooks/hooks.js";
```
AnutkaGn, line 4: I think useEffect needs to be imported from `~/libs/hooks/hooks.js`

apps/frontend/src/pages/access-management/components/group-users-table/group-users-table.tsx
```diff
@@ -0,0 +1,100 @@
+import { useEffect } from "react";
```
AnutkaGn, line 1: The same, needs to import from `~/libs/hooks/hooks.js`

apps/frontend/src/libs/components/table/table.tsx
perekotypole: Yep, I know. I'm about creating `components.ts` inside `table/components` folder

s1rserg: Oh, my bad. Thanks for the clarification!

what1s1ove (approved):

liza-veis:
> @liza-veis, there seems to be an issue introduced with the creation of preHandler on create groups route in backend. Pre handler executes properly, but handler doesn't execute. On the frontend it is just pending and on backend there is no log of request.
> 
> With empty array like that everything is working properly
> 
> ```
> this.addRoute({
> 			handler: (options) =>
> 				this.createGroup(
> 					options as APIHandlerOptions<{ body: GroupCreateRequestDto }>,
> 				),
> 			method: "POST",
> 			path: GroupsApiPath.ROOT,
> 			preHandlers: [],
> 			validation: {
> 				body: groupCreateValidationSchema,
> 			},
> 		}); 
> ```

I will create a bug ticket for the issue you described

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -7,16 +7,28 @@ import {
...
+};
+
+type Properties<T> =
+	| BaseProperties<T>
+	| (BaseProperties<T> & SelectableProperties<T>);
+
 const Table = <T extends object>({
 	columns,
 	data,
+	...selectableProperties
```
liza-veis, line 31: let's get rid of this grouping, it's a bit confusing. We only need this for the types
```suggestion
	getRowId,
	onRowSelect,
	selectedRowIds
```

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -0,0 +1,100 @@
...
+import {
+	actions as groupActions,
+	type GroupCreateRequestDto,
+} from "~/modules/groups/groups.js";
+
+import { getUserColumns, getUserRows } from "../../helpers/helpers.js";
+import { type UserRow } from "../../types/types.js";
+import styles from "./styles.module.css";
+
+const getRowId = (row: UserRow): number => row.id;
```
liza-veis, line 25: the logic for it will always be really simple, so let's define it inside the component itself, without a separate function

```
getRowId={(row): number => row.id}
```

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -0,0 +1,100 @@
...
+		void dispatch(groupActions.loadUsers({ page, pageSize }));
+	}, [dispatch, page, pageSize]);
+
+	useEffect(() => {
+		setValue("userIds", selectedUserIds);
+	}, [selectedUserIds, setValue]);
+
+	const error = errors["userIds"]?.message;
+	const hasError = Boolean(error);
+	const hasSelectedUserIds = selectedUserIds.length === EMPTY_LENGTH;
```
liza-veis, line 64: do we need this additional check at all? I think the error above should fit all the cases
```suggestion
```

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -0,0 +1,100 @@
...
+					background="secondary"
+					onPageChange={onPageChange}
+					onPageSizeChange={onPageSizeChange}
+					page={page}
+					pageSize={pageSize}
+					totalItemsCount={usersTotalCount}
+				/>
+			</div>
+			{hasError && hasSelectedUserIds && (
+				<div className={styles["error-message"]}>{error}</div>
```
liza-veis, line 94:
```suggestion
				<span className={styles["error-message"]}>{error}</div>
```

liza-veis (changes requested):

apps/frontend/src/pages/access-management/components/group-users-table/group-users-table.tsx
liza-veis: Can't we use something more appropriate for this? for example reset method from react hook form

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
liza-veis: then let's just keep it inside the component

apps/frontend/src/libs/components/table/table.tsx
s1rserg: If I do it like that I get errors: Property 'getRowId' does not exist on type 'Properties<T>' ... Cannot make this and [that](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/176#discussion_r1749321117) work together

apps/frontend/src/pages/access-management/components/group-users-table/group-users-table.tsx
s1rserg: I can use clearErrors from react hook form like this
```
	useEffect(() => {
		if (errors["userIds"] && selectedUserIds.length > EMPTY_LENGTH) {
			handleErrorsClear("userIds")
		}
	}, [selectedUserIds, errors, handleErrorsClear]);
```

s1rserg: I've tried reset but it requires another method to get values and resetField just isn't resetting errors

AnutkaGn:
Oh, it's my fault. It's a naming problem. 
To fix this, you need to change the file `controller-route-parameters.type.ts` to `preHandler?: APIPreHandler[];` 
and in the file `group.controller.ts` change to 
```
this.addRoute({
  handler: (options) =>
  this.createGroup(
  options as APIHandlerOptions<{ body: GroupCreateRequestDto }>,
  ),
  method: "POST",
  path: GroupsApiPath.ROOT,
  preHandler: [checkUserPermissions([PermissionKey.MANAGE_USER_ACCESS])],
  validation: {
  body: groupCreateValidationSchema,
  },
  });
```
I tested it on myself as if everything should work then

apps/frontend/src/libs/components/table/table.tsx
liza-veis: In this case let's retrieve those property right at the start of the function body:
```
const { getRowId } = selectableProperties
````

Comment (s1rserg) [reply apps/frontend/src/libs/components/table/table.tsx:31]: Now there are errors like
`Property 'getRowId' does not exist on type '{} | { getRowId: (row: T) => number; onRowSelect: (rowId: number) => void; selectedRowIds: number[]; }'.`

A way to fix it is 
```
	const { getRowId, onRowSelect, selectedRowIds } = 
    "getRowId" in selectableProperties && 
    "onRowSelect" in selectableProperties && 
    "selectedRowIds" in selectableProperties 
    ? selectableProperties 
    : { getRowId: undefined, onRowSelect: undefined, selectedRowIds: undefined };

	const isRowSelectable = Boolean(getRowId);
```

But here I need to use getRowId or other selectable prop because isRowSelectable gives a bunch of ts errors
```
<tr className={styles["table-row"]} key={row.id}>
								{getRowId && (
									<td className={styles["table-data"]}>
										<SelectRowCell
											id={getRowId(row.original)}
											isChecked={selectedRowIds.includes(
												getRowId(row.original),
											)}
											onToggle={onRowSelect}
										/>
									</td>
								)}
```

Comment (liza-veis) [reply apps/frontend/src/libs/components/table/table.tsx:31]: oh, I see what's the issue. let's try to use it this way. A small trick to handle ts discriminated unions with destructuring.

```tsx
	const { getRowId, onRowSelect, selectedRowIds } = selectableProperties as
		| SelectableProperties<T>
		| Record<keyof SelectableProperties<T>, undefined>;

	const isRowSelectable = typeof onRowSelect === "function";
```

s1rserg: Thanks! That works perfectly. I'll definitely remember this trick.

liza-veis (approved):

PR #184, opened 2024-09-04
feat: Add pagination to the groups table gf-83

Fjortis:
Implemented Pagination: Added pagination functionality to the group table.
Consistency with User Table: Adjusted the implementation to follow the pagination logic of the user table, ensuring uniformity across the application.
Shared Constants and Types: Utilized the same constants and types for pagination to maintain consistency.
Grouptablepag1 9


apps/backend/src/modules/groups/group.repository.ts
```diff
@@ -45,13 +49,19 @@ class GroupRepository implements Repository {
...
+	public async findAll({
+		page,
+		pageSize,
+	}: PaginationQueryParameters): Promise<PaginationResponseDto<GroupEntity>> {
+		const { results, total } = await this.groupModel
 			.query()
+			.page(page, pageSize)
 			.withGraphFetched("[permissions, users]");
 
+		// Return the paginated response with mapped entities
```
Vitaliistf, line 61: I think that this comment can be removed, it is clear what we are returning from the signature.
```suggestion
```

apps/frontend/src/libs/components/table/styles.module.css
```diff
@@ -40,7 +40,6 @@
 
 .table .table-body {
 	display: block;
-	max-height: 450px;
```
Vitaliistf, line 43: You shouldn't change this. This is limited to make the table scrollable.

packages/shared/src/libs/types/pagination-response.type.ts
Vitaliistf, line 1: File name is not corresponding to the content, and we actually already have such type in `pagination-query-parameters.type.ts`. Check your exports and imports after deleting.

tsx
Vitaliistf, line 1: I think this file is also redundant.

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -68,7 +68,6 @@ class ProjectService implements Service {
 			items: projects.items.map((item) => item.toObject()),
 		};
 	}
-
 	public async findAllbyName(
```
kolibri753, line 71: Better add that line back here, so this file won't be changed :sweat_smile:

packages/shared/src/libs/types/pagination-response.type.ts
Fjortis: Oh, that was an earlier implementation of my backend, i now cleaned it out, thanks for pointing it out :)

apps/frontend/src/libs/components/table/styles.module.css
Fjortis: You're right, will get this back to the orignal version

apps/backend/src/modules/groups/group.repository.ts
Fjortis: Yep, it is now removed from the code

apps/backend/src/modules/projects/project.service.ts
Fjortis: The line has now been added back in :D now the PR is cleaner, thanks!

apps/frontend/src/libs/components/table-pagination/libs/constants/page-size-options.constant.ts
```diff
@@ -1,4 +1,5 @@
 const PAGE_SIZE_OPTIONS = [
+	{ label: "5", value: 5 },
```
liza-veis, line 2: let's have 10 as minimum
```suggestion
```

apps/frontend/src/modules/groups/slices/group.slice.ts
```diff
@@ -23,6 +25,7 @@ const { actions, name, reducer } = createSlice({
 		});
 		builder.addCase(loadAll.fulfilled, (state, action) => {
 			state.groups = action.payload.items;
+			state.totalGroupsCount = action.payload.totalItems;
```
liza-veis, line 28: the same as for users
```suggestion
			state.groupsTotalCount = action.payload.totalItems;
```

tsx
liza-veis, line 1: Remove this file

apps/frontend/src/pages/access-management/access-management.tsx
```diff
@@ -62,17 +82,26 @@ const AccessManagement = (): JSX.Element => {
 				<div className={styles["users-table"]}>
 					<Table<UserRow> columns={userColumns} data={userData} />
 					<TablePagination
```
liza-veis, line 84: What about query parameters for page and pageSize for both tables? they won't work properly with the same name, so we need to rename them

liza-veis (changes requested):

apps/frontend/src/pages/access-management/access-management.tsx
Fjortis: This has now been adjusted, it now uses queryParameterPrefix to create the query parameters separetly for both tables
`	const {
		onPageChange: onGroupPageChange,
		onPageSizeChange: onGroupPageSizeChange,
		page: groupPage,
		pageSize: groupPageSize,
	} = usePagination({
		queryParameterPrefix: "group",
		totalItemsCount: groupsTotalCount,
	});`

tsx
Fjortis: Removed

apps/frontend/src/modules/groups/slices/group.slice.ts
Fjortis: I've adjusted this to match users.

apps/frontend/src/libs/components/table-pagination/libs/constants/page-size-options.constant.ts
Fjortis: I put it back to 10 as minimum

what1s1ove (approved):
lgtm after Lisa's comments

liza-veis (approved):

PR #185, opened 2024-09-04
feat: Add an ability to delete personal account gf-146

HoroshkoMykhailo:
Summary:

- Added column deleted_at in table users for soft delete.
- Added an ability to soft delete Personal Account .
- Created a component and confirmation modal for Account Delete.
- Upon confirmation, user is logged out and redirected to the Sign In page.

image

image

- When user attempt to log in with deleted account it result in "Invalid credentials" toast notification. 
- When user attempt to sign up using the same email it result in "Email already in use" toast notification because there can`t be two records with the same email in database.


apps/frontend/src/pages/profile/styles.module.css
```diff
@@ -1,16 +1,30 @@
...
-	max-width: 800px;
-	margin: 0 auto;
+.title {
+	margin: 0;
+	font-size: 30px;
+	line-height: 120%;
+}
+
+.divider {
+	width: 800px;
```
Vitaliistf, line 27: Avoid strict width and height, use` max-width` or `min-width`.

apps/frontend/src/libs/enums/notification-message.enum.ts
```diff
@@ -2,6 +2,7 @@ const NotificationMessage = {
 	PROJECT_CREATE_SUCCESS: "Project was successfully created",
 	PROJECT_UPDATE_SUCCESS: "Project was successfully updated.",
 	SUCCESS_PROFILE_UPDATE: "Successfully updated profile information.",
+	SUCCESS_USER_DELETE: "User deleted successfully.",
```
Vitaliistf, line 5: I think we should agree how to name this messages, because they are written in different patterns here.

apps/frontend/src/libs/components/confirmation-modal/confirmation-modal.tsx
```diff
@@ -0,0 +1,45 @@
+import { Button, Modal } from "~/libs/components/components.js";
+import { useCallback } from "~/libs/hooks/hooks.js";
+
+import styles from "./styles.module.css";
+
+type ConfirmationModalProperties = {
```
AnutkaGn, line 6: It is better to follow the same code style and simply write Properties

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -31,14 +32,22 @@ class UserRepository implements Repository {
...
-		return Promise.resolve(true);
+	public async delete(id: number): Promise<boolean> {
+		const numberDeletedRows = await this.userModel
+			.query()
+			.patch({ deletedAt: new Date().toISOString() })
+			.where({ id })
+			.whereNull("deletedAt")
+			.execute();
+
+		return numberDeletedRows > NOTHING_DELETED_COUNT;
```
what1s1ove, line 43:
```suggestion
		return Boolean(numberDeletedRows);
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -31,14 +32,22 @@ class UserRepository implements Repository {
 		return UserEntity.initialize(user);
 	}
 
-	public delete(): ReturnType<Repository["delete"]> {
-		return Promise.resolve(true);
+	public async delete(id: number): Promise<boolean> {
+		const numberDeletedRows = await this.userModel
```
what1s1ove, line 36:
```suggestion
		const deletedRowsCount = await this.userModel
```

apps/frontend/src/libs/enums/notification-message.enum.ts
```diff
@@ -1,7 +1,8 @@
 const NotificationMessage = {
+	PROFILE_UPDATE_SUCCESS: "Successfully updated profile information.",
 	PROJECT_CREATE_SUCCESS: "Project was successfully created",
 	PROJECT_UPDATE_SUCCESS: "Project was successfully updated.",
-	SUCCESS_PROFILE_UPDATE: "Successfully updated profile information.",
+	USER_DELETE_SUCCESS: "User deleted successfully.",
```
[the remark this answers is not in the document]
HoroshkoMykhailo: We agreed in Slack that (Subject) (Action) (Result) fits best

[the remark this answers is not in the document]
Fjortis: Yep, sorry, this was among my pending request i never sent :*(

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -60,7 +69,17 @@ class UserRepository implements Repository {
...
+		const user = await this.userModel
+			.query()
+			.findOne({ email })
+			.whereNull("deletedAt")
+			.execute();
+
+		return user ? UserEntity.initialize(user) : null;
+	}
+
+	public async findByEmailCreate(email: string): Promise<null | UserEntity> {
```
liza-veis, line 81: why did you rename it 🤔 
```suggestion
	public async findByEmail(email: string): Promise<null | UserEntity> {
```

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -60,7 +69,17 @@ class UserRepository implements Repository {
...
+		const user = await this.userModel.query().findOne({ email }).execute();
```
liza-veis, line 82: Don't we need to add a check with deleted at here?

packages/shared/src/libs/modules/http/libs/enums/http-code.enum.ts
```diff
@@ -2,6 +2,7 @@ const HTTPCode = {
 	CONFLICT: 409,
 	CREATED: 201,
 	INTERNAL_SERVER_ERROR: 500,
+	NO_CONTENT: 204,
```
liza-veis, line 5: Let's use 200 status and return isDeleted in payload as in this pr #168

apps/frontend/src/pages/profile/styles.module.css
```diff
@@ -16,3 +16,17 @@
 	font-weight: 600;
 	line-height: 120%;
 }
+
+.profile-update {
```
liza-veis, line 20:
```suggestion
.profile-info {
```

liza-veis (changes requested):

apps/backend/src/modules/users/user.repository.ts
HoroshkoMykhailo: We don`t need a check when we use this method in sign up, so user can`t sign up with the soft deleted email. I will change this methods, I think I made it wrong.

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -59,8 +68,17 @@ class UserRepository implements Repository {
 		};
 	}
 
-	public async findByEmail(email: string): Promise<null | UserEntity> {
-		const user = await this.userModel.query().findOne({ email });
+	public async findByEmail(
```
liza-veis, line 71: Oh now I see why it was without deletedAt filter, looks like for now we only need to use this method to find users including deleted, so let's remove includeDeleted parameter and return it as it was

apps/frontend/src/pages/profile/profile.tsx
```diff
@@ -11,9 +11,14 @@ const Profile = (): JSX.Element => {
...
 			<div className={styles["profile-layout"]}>
-				<h1 className={styles["title"]}>Profile</h1>
-
-				<EditUserForm user={authenticatedUser as UserAuthResponseDto} />
+				<div className={styles["profile-info"]}>
+					<h1 className={styles["title"]}>Profile</h1>
+					<EditUserForm user={authenticatedUser as UserAuthResponseDto} />
+				</div>
+				<div className={styles["divider"]} />
+				{authenticatedUser?.id && (
```
liza-veis, line 19: Let's cast this type, we can only get to this page when we have authenticatedUser

liza-veis (changes requested):

apps/backend/src/modules/users/user.repository.ts
HoroshkoMykhailo: Just to clarify, return to two separate methods? Because we need to find users without deleted ones in sign-in flow

liza-veis: then you did it right, let's just rename the parameter

apps/backend/src/modules/users/user.repository.ts
```diff
@@ -59,8 +68,17 @@ class UserRepository implements Repository {
...
+		email: string,
+		includeDeleted = false,
```
liza-veis, line 73: Let's use more common prefix for boolean variables
```suggestion
		hasDeleted = false,
```

liza-veis (approved):

what1s1ove (approved):

PR #194, opened 2024-09-05
fix: Font size of the "Profile" heading doesn`t match the design mockup gf-192

perekotypole:
- fix title font size and add additional settings (line-height, word-break)
- change profile layout styles

image

image

### DIFF
--- a/apps/frontend/src/pages/profile/styles.module.css
+++ b/apps/frontend/src/pages/profile/styles.module.css
@@ -13,6 +13,6 @@
 .title {
 	margin: 0;
 	font-size: 30px;
+	font-weight: 600;
 	line-height: 120%;
-	word-break: break-word;
 }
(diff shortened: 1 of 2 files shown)


what1s1ove (approved):

apps/frontend/src/pages/profile/styles.module.css
```diff
@@ -1,16 +1,18 @@
...
+	padding: 0;
 	margin: 0 auto;
+	color: var(--color-text-primary);
+}
+
+.title {
+	margin: 0;
+	font-size: 30px;
+	line-height: 120%;
+	word-break: break-word;
```
liza-veis, line 17: We need this only for user input values, like project title etc.
```suggestion
```

perekotypole: oh, got it

liza-veis (approved):

PR #203, opened 2024-09-06
fix: inappropriate appearance of the scroll element at the Table component on Windows OS gf-197

Vitaliistf:
Added global styles for scroll element. As an example for styles I took the only scrollbar which is shown in design.
Screenshot 2024-09-06 at 08 13 33
Here's how does scrollbar looks across different browsers on Mac:
Webkit browsers (Chrome, Safari, new Edge):
Screenshot 2024-09-06 at 08 23 06
Firefox:
Screenshot 2024-09-06 at 08 24 04
They look a little bit different, because there is no ability to customise one in Firefox more than that.
Unfortunately, have no ability to test it on windows, but it should there too work.

### DIFF
--- a/apps/frontend/src/assets/css/scaffolding.css
+++ b/apps/frontend/src/assets/css/scaffolding.css
@@ -1,5 +1,7 @@
 * {
 	box-sizing: border-box;
+	scrollbar-width: thin;
+	scrollbar-color: var(--color-border-primary) var(--color-background-secondary);
 }
 
 body {
@@ -31,4 +33,21 @@ body {
 	clip-path: inset(100%);
 	white-space: nowrap;
 	border: 0;
+}
+
+::-webkit-scrollbar {
+	width: 8px;
+	height: 8px;
+}
+
+::-webkit-scrollbar-thumb {
+	background-color: var(--color-border-primary);
+}
+
+::-webkit-scrollbar-thumb:hover {
+	background-color: var(--color-brand-primary);
+}
+
+::-webkit-scrollbar-track {
+	background: var(--color-background-secondary);
 }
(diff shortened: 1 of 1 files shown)


apps/frontend/src/assets/css/scaffolding.css
```diff
@@ -32,3 +32,25 @@ body {
 	white-space: nowrap;
 	border: 0;
 }
+
+html {
+	scrollbar-width: thin;
+	scrollbar-color: var(--color-border-primary) var(--color-background-secondary);
```
liza-veis, line 38: Let's move this to the first block with *

apps/frontend/src/assets/css/scaffolding.css
```diff
@@ -32,3 +32,25 @@ body {
...
+}
+
+html::-webkit-scrollbar {
```
liza-veis, line 41: remove html prefix here, we want to apply it for every element
```suggestion
::-webkit-scrollbar {
```

apps/frontend/src/assets/css/scaffolding.css
```diff
@@ -32,3 +32,25 @@ body {
...
+	width: 8px;
+	height: 8px;
+}
+
```
liza-veis, line 45: By the way, don't we need to remove scrollbar styles from textarea? please check that new styles apply to it and to tables

liza-veis (changes requested):

apps/frontend/src/assets/css/scaffolding.css
Vitaliistf: Oh, yeah, I thought we'll leave them for the textarea. New styles apply to tables, and textarea - confirmed 😅.

liza-veis (approved):

PR #210, opened 2024-09-06
feat: Add a "No projects found" message when search field no results on the Projects page gf-187

HoroshkoMykhailo:
Summary:

- Add the message "No projects found matching your search criteria. Please try different keywords." if no results were found during the search on the "Projects" page.

image

### DIFF
--- a/apps/frontend/src/pages/projects/projects.tsx
+++ b/apps/frontend/src/pages/projects/projects.tsx
@@ -145,7 +145,7 @@ const Projects = (): JSX.Element => {
 							/>
 						))
 					) : (
-						<p className={styles["no-projects-message"]}>
+						<p className={styles["empty-placeholder"]}>
 							No projects found matching your search criteria. Please try
 							different keywords.
 						</p>
--- a/apps/frontend/src/pages/projects/projects.tsx
+++ b/apps/frontend/src/pages/projects/projects.tsx
@@ -50,7 +50,7 @@ const Projects = (): JSX.Element => {
 		[dispatch],
 	);
 
-	const hasProject = projects.length === EMPTY_LENGTH;
+	const hasProject = projects.length !== EMPTY_LENGTH;
 
 	const {
 		isOpened: isCreateModalOpen,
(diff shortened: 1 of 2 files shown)


apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -131,20 +131,27 @@ const Projects = (): JSX.Element => {
...
+						projects.map((project) => (
+							<ProjectCard
+								key={project.id}
+								onDelete={handleDeleteClick}
+								onEdit={handleEditClick}
+								project={project}
+							/>
+						))
+					) : (
+						<p className={styles["no-projects-message"]}>
```
liza-veis, line 148:
```suggestion
						<p className={styles["empty-placeholder"]}>
```

apps/frontend/src/pages/projects/styles.module.css
```diff
@@ -17,3 +17,9 @@
 	gap: 12px;
 	padding: 20px 0;
 }
+
+.no-projects-message {
```
liza-veis, line 21: I think the placeholder should be not so bright, also the font-size should be the same as for body, and it I think it will look better if it will be horizontally centered

liza-veis (changes requested):

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -50,7 +50,7 @@ const Projects = (): JSX.Element => {
 		[dispatch],
 	);
 
-	const hasProject = projects.length === EMPTY_LENGTH;
+	const hasProject = projects.length !== EMPTY_LENGTH;
```
perekotypole, line 53: How about using this?
```suggestion
	const hasProject = Boolean(projects.length)
```

HoroshkoMykhailo: Good idea

HoroshkoMykhailo: But eslint doesn`t allow it

image

perekotypole: okay, got it

what1s1ove (approved):

PR #218, opened 2024-09-06
feat: save search value to query parameters gf-200

Fjortis:
Implemented  saving of search value to query parameters.
 After reloading the page initial state is taken from query params and filter the projects accordingly.
 adjusted GET all projects request for cases where name is missing.
Screenshot 2024-09-06 173453
Screenshot 2024-09-06 173501
Screenshot 2024-09-06 173507
Screenshot 2024-09-06 175643
Screenshot 2024-09-06 175654


apps/frontend/src/libs/components/search/libs/hooks/use-search-query-parameter.hook.ts
```diff
@@ -0,0 +1,34 @@
+import { useLocation, useNavigate } from "react-router-dom";
+
+import { useCallback } from "~/libs/hooks/hooks.js";
```
Fjortis, line 3: Why can i not grab these hooks from the same location? I am unable to do it without lint errors?

apps/frontend/src/libs/components/search/libs/constants/search-parameter-key.constant.ts
```diff
@@ -0,0 +1,3 @@
+const SEARCH_PARAMETER_KEY = "search";
```
liza-veis, line 1: we alread have QueryParameterName enum and it should be in it, also let's use q keyword for this, it's more common for search. Like below

```typescript
const QueryParameterName = {
	SEARCH: "q",
} as const;
```

apps/frontend/src/libs/components/search/libs/hooks/use-search-query-parameter.hook.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { SEARCH_PARAMETER_KEY } from "../constants/constants.js";
+
+type Properties = {
+	searchQuery: string;
+	updateSearchParams: (newSearchValue: string) => void;
+};
+
+const useSearchQueryParameter = (defaultSearch: string): Properties => {
```
liza-veis, line 12: It should be consistent with the usePagination hook. We need to take values from query string inside it and the whole logic should be the same as for pagination. Also don't forget to keep the same naming and code style across the app
```suggestion
const useSearch = (): {
  search: string;
  onSearchChange: (search: string) => void;
} => {
```

apps/backend/src/modules/projects/project.controller.ts
```diff
@@ -210,12 +216,18 @@ class ProjectController extends BaseController {
 	): Promise<APIHandlerResponse> {
 		const { query } = options;
 
+		if (!query.name) {
+			return {
+				payload: await this.projectService.findAll(),
+				status: HTTPCode.OK,
+			};
+		}
```
liza-veis, line 224: instead let's merge findAll and findByName methods for projects service and projects repository and keep only findAll. For repository type let's define find like the following. And inside projects repository we need to do the same as we did for findByEmail method in users repository - in case the query is true, only then we attach `whereILike` to the query
```suggestion
  findAll(payload: unknown): Promise<{ items: T[] }>;
```

liza-veis (changes requested):

what1s1ove (commented):
lgtm after Lisa's comments

apps/backend/src/modules/projects/project.controller.ts
Fjortis: Just to make sure, before i do any more changes, do you mean findByName, or the findAllbyName?

liza-veis: I mean findAllbyName

apps/backend/src/libs/types/repository.type.ts
```diff
@@ -2,7 +2,7 @@ type Repository<T = unknown> = {
 	create(payload: unknown): Promise<T>;
 	delete(id: number): Promise<boolean>;
 	find(id: number): Promise<null | T>;
-	findAll(parameters: unknown): Promise<{ items: T[] }>;
+	findAll(payload: unknown): Promise<{ items: T[] }>;
```
perekotypole, line 5: Wouldn't `query`/`queries` be more appropriate instead of `payload` here?

apps/frontend/src/libs/enums/query-parameter-name.enum.ts
```diff
@@ -1,6 +1,7 @@
 const QueryParameterName = {
 	PAGE: "page",
 	PAGE_SIZE: "pageSize",
+	SEARCH: "q",
```
perekotypole, line 4: I'd make search param more semantic
```suggestion
	SEARCH: "seach"
```

apps/frontend/src/libs/components/search/libs/hooks/use-search-query-parameter.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+	useSearchParams,
+	useState,
+} from "~/libs/hooks/hooks.js";
+
+type UseSearch = () => {
+	onSearchChange: (search: string) => void;
+	search: string;
+};
+
+const useSearch: UseSearch = () => {
```
perekotypole, line 14:
```suggestion
const useSearch = (): UseSearch => {
```
Use it because you explicitly specify the return type of the function, not the value type

apps/frontend/src/libs/components/search/libs/hooks/use-search-query-parameter.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+	const [searchParameters, setSearchParameters] = useSearchParams();
+
+	const searchParameterName = QueryParameterName.SEARCH;
+
+	const NamedSearchParameter = searchParameters.get(searchParameterName);
+
+	const [search, setSearch] = useState<string>(NamedSearchParameter || "");
```
perekotypole, line 21: A1
Use `??` instead of `||`

apps/frontend/src/libs/components/search/libs/hooks/use-search-query-parameter.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+	onSearchChange: (search: string) => void;
+	search: string;
+};
+
+const useSearch: UseSearch = () => {
+	const [searchParameters, setSearchParameters] = useSearchParams();
+
+	const searchParameterName = QueryParameterName.SEARCH;
+
+	const NamedSearchParameter = searchParameters.get(searchParameterName);
```
liza-veis, line 19:
```suggestion
	const searchParameter = searchParameters.get(searchParameterName);
```

liza-veis (changes requested):

apps/frontend/src/libs/components/search/libs/hooks/use-search-query-parameter.hook.ts
```diff
@@ -0,0 +1,45 @@
...
+	const [search, setSearch] = useState<string>(NamedSearchParameter || "");
+
+	useEffect(() => {
+		const updatedSearchParameters = new URLSearchParams(searchParameters);
+
+		if (search) {
+			updatedSearchParameters.set(searchParameterName, search);
+		} else {
+			updatedSearchParameters.delete(searchParameterName);
+		}
```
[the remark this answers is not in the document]
Fjortis: I tought it would make the url look cleaner when it is no search.

apps/frontend/src/libs/components/search/libs/hooks/use-search-query-parameter.hook.ts
Fjortis: Thanks, this has been changed.

liza-veis (approved):

yevhenii-hladkov (approved):

apps/frontend/src/libs/components/search/libs/hooks/use-search.hook.ts
```diff
@@ -0,0 +1,30 @@
+import { QueryParameterName } from "~/libs/enums/enums.js";
+import { useEffect, useSearchParams, useState } from "~/libs/hooks/hooks.js";
+
+type UseSearch = {
+	search: string;
+	setSearch: (search: string) => void;
+};
+
+const useSearch = (): UseSearch => {
```
liza-veis, line 9: we need to rename setSearch to onSearch before exposing it
```suggestion
const useSearch = (): {
	search: string;
	onSearch: (search: string) => void;
} => {
```
```tsx
	return {
		search,
		onSearch: setSearch,
	};
```

liza-veis (changes requested):

liza-veis (approved):

what1s1ove (approved):

PR #219, opened 2024-09-06
 feat: Change the text message and the button content of Confirmation modal window on Deleting user account gf-216

HoroshkoMykhailo:
Summary:

- Changed text message and button content

image

### DIFF
--- a/apps/frontend/src/libs/components/confirmation-modal/confirmation-modal.tsx
+++ b/apps/frontend/src/libs/components/confirmation-modal/confirmation-modal.tsx
@@ -4,6 +4,7 @@ import { useCallback } from "~/libs/hooks/hooks.js";
 import styles from "./styles.module.css";
 
 type Properties = {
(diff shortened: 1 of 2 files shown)


apps/frontend/src/libs/components/confirmation-modal/confirmation-modal.tsx
```diff
@@ -4,13 +4,15 @@ import { useCallback } from "~/libs/hooks/hooks.js";
 import styles from "./styles.module.css";
 
 type Properties = {
+	confirmLabel?: string;
```
liza-veis, line 7: I think for now we can just use "Confirm" in such modals for all cases (agree with qa first).
```suggestion
```

liza-veis (changes requested):

what1s1ove (commented):
lgtm after Lisa's comments

liza-veis (approved):

Issue #180, opened 2024-09-04
feat: Add an ability to generate and retrieve project api key


liza-veis:
### What feature?

- [x] We need to add a new table that will store project_api_keys, it should contain encoded_key, project_id, and user_id columns. project_id - user_id pair should be unique.
- [x] Before we insert the key into the database, we need to encrypt it. And when we retrieve it (to send it to the client), we will need to decrypt it.
- [x] We need to attach this token in response when we get a project by id.
- [x] On the frontend we need to add the "Setup Analytics" button which will open a modal.
- [x] For now, the modal will only contain a field for displaying api token and "Generate" button.
- [x] When the token is present, "Generate" button should be disabled.

Notes: 

- We might be able to implement this using our current token service, but this is to be determined during implementation.
- An ability to re-generate API tokens will be added in future iterations.

### Screenshots

![Setup analytics modal](https://github.com/user-attachments/assets/aa9e208f-732c-497a-8880-48c16758350e)

GvoFor:
>...an icon indicator showing loading and green tick is shown.

Can you specify, how indicator should look like (should it be green "broken" ring?), where this indicatior should be placed, and where green tick should be shown

GvoFor:
@liza-veis
What should be the route for generation api key?
Should it be `.../api/v1/projects/{id}/api-key` or something else?

liza-veis:
> What should be the route for generation api key?
> Should it be .../api/v1/projects/{id}/api-key or something else?

Let's use `/api/v1/project-api-keys` as a separate resource

GvoFor:
Will we need to be able to retrieve all the keys (`findAll`) and retrieve by id (`getById`) for this resource?

I'm writing because I already add all the stuff needed for this (and this even already works), but now I realized, that for `"We need to attach this token in response when we get a project by id." ` I can just use some `findByProjectId` method from `projectApiKeysService` inside `projectService`, without any additional REST API, so I it seems to me, that I was supposed only to add `POST` method for api key creation for that route...

So should I delete `GET` methods responsible for retrieving all api keys, and api key by id?

Also want to clarify. I created separate module `project-api-keys` for this. Is it right?

liza-veis:
> Will we need to be able to retrieve all the keys (findAll) and retrieve by id (getById) for this resource?

In the next two weeks, we are not planning to add something like this. We only will need to search by the encrypted_key field. So let's remove unused methods.

[...]

PR #221, opened 2024-09-07
feat: add an ability to save and retrieve activity logs gf-181

sandrvvu:
**Add contributors table**: This table includes the following columns: id, created_at, updated_at, and name.

**Add git_emails table**: This table includes the following columns: id, created_at, updated_at, contributor_id, and email.

**Add activity_logs table**: This table includes the following columns: id, created_at, updated_at, git_email_id, project_id, created_by_user_id, date, and commits_number.

**_[BLOCKER: Issue #180]_**.  For the POST method, we will implement a different type of authorization. The request will include a Bearer token, but instead of the usual token, it will contain a project_api_key. The created_by_user_id and project_id will be extracted from this key after decoding it and retrieving the associated data from the database. The GET request will continue to use the standard authorization method.

**POST request payload**: The payload for the POST request will consist of records grouped by date, each containing the author's name, email, and commits_number.

**Contributor and git_email creation logic**: For each git_email that does not already exist, a new contributor record must be created first. The contributor's name will be derived from the author's email. Afterward, the git_email record will be created, as it requires a valid contributor_id.

**Add GET endpoint**: This endpoint will allow retrieving activity records, including the associated contributor data.

GET:
image

API KEY
image

POST request:
image
POST response:
image

DB activity_logs:
image
DB contributors:
image
DB git_emails:
image

UPDATED LOOK:
image


sandrvvu (commented):
Hey! I have some questions:
1) Should I handle the manipulation of contributors and git emails in the service layer or the repository layer? 
2) Do we retrieve the token from the request parameters, pass it to the service layer, decode it there, check if it exists in the database, or follow a different approach?
3) Do I need to attach contributor data only for GET method or for both?

liza-veis:
> Should I handle the manipulation of contributors and git emails in the service layer or the repository layer?

It depends, if you can handle it in one query, like passing needed payload to contributor and this will create a needed git email then you can do so. But if need to handle those creations separately, then should should be controlled on the service level, because contributors repository should know nothing about git emails repository.

> Do we retrieve the token from the request parameters, pass it to the service layer, decode it there, check if it exists in the database, or follow a different approach?

We retrieve it from authorization header, the same way as we do for the usual token. Your flow is right.

> Do I need to attach contributor data only for GET method or for both?

let's attach it to both to keep the return data type the same for get/create.

apps/backend/src/db/migrations/20240904155242_add_activity_logs_table.ts
```diff
@@ -0,0 +1,61 @@
...
+} as const;
+
+const GIT_EMAILS_TABLE = "git_emails";
+const PROJECTS_TABLE = "projects";
+const USERS_TABLE = "users";
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.createTable(TABLE_NAME, (table) => {
+		table.increments(ColumnName.ID).primary();
+		table.integer(ColumnName.COMMITS_NUMBER).notNullable();
```
Fjortis, line 23: Given that commits_number should never be negative, i think it would be better to use unsigned() for this column.

packages/shared/src/modules/activity-logs/libs/types/activity-log-create-item-response-dto.type.ts
GvoFor, line 1: This type is the same as `ActivityLogGetAllItemResponseDto`. We decided, that for now there is no need in such type, and you should use `ActivityLogGetAllItemResponseDto` instead of `ActivityLogCreateItemResponseDto`. You can find same approach in other modules.

packages/shared/src/libs/enums/exception-message.enum.ts
```diff
@@ -1,5 +1,9 @@
 const ExceptionMessage = {
-	CREATE_PROJECT_GROUP_FAILED: "Failed to create project group",
+	CONTRIBUTOR_NOT_FOUND: "Contributor not found.",
+	CREATE_ACTIVITY_LOG_FAILED: "Failed to save activity logs.",
+	CREATE_PROJECT_GROUP_FAILED: "Failed to create project group.",
+	EMAIL_USED: "Email address is already in use.",
+	GIT_EMAIL_NOT_FOUND: "Git email not found.",
```
GvoFor, line 6: We decided to stick to the format: `SOMETHING_ACTION_RESULT` -- for example `USER_CREATE_SUCCESS`, so all the messages will be "grouped" (place near each other) by the object, they relate to
```suggestion
	CONTRIBUTOR_NOT_FOUND: "Contributor not found.",
	ACTIVITY_LOG_CREATE_FAILED: "Failed to save activity logs.",
	PROJECT_GROUP_CREATE_FAILED: "Failed to create project group.",
	EMAIL_USED: "Email address is already in use.",
	GIT_EMAIL_NOT_FOUND: "Git email not found.",
```

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -0,0 +1,119 @@
...
+							.insert({ contributorId: contributor.id, email: authorEmail })
+							.returning("*");
+					}
+				} catch {
+					throw new ActivityLogError({
+						message: ExceptionMessage.GIT_EMAIL_NOT_FOUND,
+						status: HTTPCode.NOT_FOUND,
+					});
+				}
+
```
GvoFor, line 79: This logic should be moved into separate `Contributor`/`GitEmail` services or repositories idealy

apps/backend/src/modules/activity-logs/activity-log.repository.ts
```diff
@@ -0,0 +1,97 @@
...
+				contributor,
+			};
+
+			return ActivityLogEntity.initialize(activityLogWithContributorData);
+		} catch {
+			throw new ActivityLogError({
+				message: ExceptionMessage.CREATE_ACTIVITY_LOG_FAILED,
+				status: HTTPCode.INTERNAL_SERVER_ERROR,
+			});
+		}
```
GvoFor, line 54: It seems to me, that error handling should be performed on the service level

apps/backend/src/modules/activity-logs/activity-log.service.ts
liza-veis: agree, we need separate services for git emails / contributors

sandrvvu (commented):
Question 1:
According to the task description, the created_by_user_id and project_id should be extracted from this key.

However, the decision was made in "feat: Add an ability to generate and retrieve project API key #180" that the API key will be bound only to the project, not to a pair of project and user. This means that there will be a single userId for each project.

Given this, the created_by_user_id field in the activity log no longer seems relevant. Do we still need this field at all? Currently, I am passing updatedByUser from the project API key. Is this approach correct, or should we consider another solution?

Question 2:
Where or how can I access the authorization header in the controller to retrieve api key token?

liza-veis:
> **Question 1:**
> According to the task description, the created_by_user_id and project_id should be extracted from api key.
> 
> However, the decision was made in "feat: Add an ability to generate and retrieve project API key #180" that the API key will be bound only to the project, not to a pair of project and user. This means that there will be a single userId for each project.
> 
> Given this, the created_by_user_id field in the activity log no longer seems relevant. Do we still need this field at all? Currently, I am passing updatedByUserId from the project API key. Is this approach correct, or should we consider another solution?

We still will have user id, but we will pass it with the request payload to create analytics logs.

> **Question 2:**
> Where or how can I access the authorization header in the activity log controller to retrieve api key token?

You need to retrieve it on a controller level from authorization header and pass as apiKey to payload to service.

packages/shared/src/modules/activity-logs/libs/enums/activity-logs-api-path.enum.ts
```diff
@@ -0,0 +1,6 @@
+const ActivityLogsApiPath = {
+	ROOT: "/",
+	SAVE: "/save",
```
s1rserg, line 3:
```suggestion
```

Why do we need this save at all? Can't we post to ROOT?

sandrvvu: We have a different type of authorization for the POST method. So I add SAVE, so I can define this as a white route for plugin.

s1rserg: Oh, I see. Just seems odd. Not really sure if it's needed or we can just stay with this save but we could add allowed methods to the white routes and make some minor adjustments to the authorization plugin for this. However, it's best to discuss this with coaches to avoid potential unnecessary work.

what1s1ove: @liza-veis  take a look please, i also do not like the name of this route

what1s1ove (commented):
do not forget to change a db schema in the readme

apps/backend/src/modules/contributors/contributors.service.ts
```diff
@@ -0,0 +1,78 @@
...
+
+		const existingContributor =
+			await this.contributorRepository.findByName(name);
+
+		if (existingContributor) {
+			throw new ContributorError({
+				message: ExceptionMessage.CONTRIBUTOR_NAME_USED,
+				status: HTTPCode.CONFLICT,
+			});
+		}
```
liza-veis, line 33: Contributor name can be non unique, we will identify different contributors by different git emails
```suggestion
```

apps/backend/src/modules/contributors/contributors.service.ts
```diff
@@ -0,0 +1,78 @@
...
+	private contributorRepository: ContributorRepository;
+
+	public constructor(contributorRepository: ContributorRepository) {
+		this.contributorRepository = contributorRepository;
+	}
+
+	public async create(
+		payload: ContributorCreateRequestDto,
+	): Promise<ContributorGetAllItemResponseDto> {
+		const { name } = payload;
```
liza-veis, line 23: can we create git_emails through relation inside contributors service?

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -0,0 +1,115 @@
...
+				let gitEmail = await this.gitEmailService.findByEmail(authorEmail);
+
+				if (!gitEmail) {
+					gitEmail = await this.gitEmailService.create({
+						contributorId: contributor.id,
+						email: authorEmail,
+					});
+				}
```
liza-veis, line 74: we don't need to look for contributor by name here, we need to look if we have an existing git email and if not, then create a contributor with git email

liza-veis (changes requested):

apps/backend/src/modules/activity-logs/activity-log.controller.ts
```diff
@@ -0,0 +1,165 @@
...
+ *         createdByUserId:
+ *           type: number
+ *           minimum: 1
+ *         date:
+ *           type: string
+ *           format: date-time
+ *         gitEmailId:
+ *           type: number
+ *           minimum: 1
+ *         ptojectId:
```
Fjortis, line 37: Very minor typo
```suggestion
 *         projectId :
```

what1s1ove (approved):
lgtm after Lisa's comments

Fjortis:
Just checking, is it meant for activity logs and git emails to be deleted if a contributor or project gets removed? Maybe we should avoid using `CASCADE `to keep the logs and emails intact?

apps/backend/src/modules/contributors/contributors.service.ts
sandrvvu: I chose to stick with the old method of using gitEmailService.create. Just because updating the model or types in the contributors service create to handle sending correct format data(gitEmail with the nested contributor?) or finding again already created gitEmail in the activity log service would have made things pretty messier. It seemed cleaner for me to keep things as they are for now.

liza-veis:
> Just checking, is it meant for activity logs and git emails to be deleted if a contributor or project gets removed? Maybe we should avoid using CASCADE to keep the logs and emails intact?

Good question. I think this won't be a problem because activity logs are only linked to the git emails. And we will never delete git emails, even if we delete a contributor, the git email linked to it should be transferred to another contributor before.

apps/backend/src/modules/contributors/contributors.service.ts
liza-veis: good, I'm ok with this approach

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -0,0 +1,109 @@
...
+		const { projectId } = existingProjectApiKey;
+
+		const createdActivityLogs: ActivityLogGetAllResponseDto = {
+			items: [],
+		};
+
+		for (const record of items) {
+			const { date, items } = record;
+
+			for (const logItem of items) {
```
liza-veis, line 55: Let's create a separate helper method for this, to reduce the complexity. Something like createActivityLog.

liza-veis (changes requested):

apps/backend/src/modules/git-emails/git-email.repository.ts
```diff
@@ -0,0 +1,58 @@
...
+		this.gitEmailModel = gitEmailModel;
+	}
+
+	public async create(entity: GitEmailEntity): Promise<GitEmailEntity> {
+		const { contributor, email } = entity.toNewObject();
+
+		const gitEmail = await this.gitEmailModel
+			.query()
+			.insert({
+				contributor,
```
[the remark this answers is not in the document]
sandrvvu: I made this change to avoid redundant data (contributor and contributorId) in the entity (previous comment), and I pass in the contributor in the insert method to add contributor's id and name values to git-emails, because without it it sends `underfined`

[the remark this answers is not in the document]
liza-veis: I didn't quite get what's the issue 😅 As I see you create contributor inside the log-activity service separately, and only after that you create a git_email, so the contributor already exists and we only need contributor_id. It's ok if we have both contributor and contributorId inside the entity, it sounds better to me to have it like this instead of passing unnecessary entity to the query

apps/backend/src/modules/activity-logs/libs/types/activity-log-creation-data.type.ts
```diff
@@ -0,0 +1,12 @@
+type ActivityLogCreationData = {
```
liza-veis, line 1: don't forget about file name
```suggestion
type ActivityLogCreateItemResponseDto = {
```

yevhenii-hladkov (approved):

liza-veis (approved):

PR #222, opened 2024-09-08
feat: add a character counter for the project description gf-188

AnutkaGn:
Add a character counter below the "Description" text area in the project creation form, displaying the current character count in the format "X/1000".
image
Add an error message that will appear immediately when there are more than 1000 characters and the submit button is not pressed


apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
```diff
@@ -45,6 +55,17 @@ const ProjectCreateForm = ({ onSubmit }: Properties): JSX.Element => {
...
+				<span className={styles["char-counter"]}>
+					{charCount}/{ProjectValidationRule.DESCRIPTION_MAXIMUM_LENGTH}
+				</span>
+			)}
+			{charCount > ProjectValidationRule.DESCRIPTION_MAXIMUM_LENGTH &&
+				!isSubmitted && (
+					<span className={styles["input-error"]}>
+						{ProjectValidationMessage.DESCRIPTION_TOO_LONG}
+					</span>
+				)}
```
yevhenii-hladkov, line 68:
```suggestion
			{isCharCountValid ? (
				<span className={styles["char-counter"]}>
					{charCount}/{ProjectValidationRule.DESCRIPTION_MAXIMUM_LENGTH}
				</span>
			) : !isSubmitted && (
				<span className={styles["input-error"]}>
					{ProjectValidationMessage.DESCRIPTION_TOO_LONG}
				</span>
			)}
```

yevhenii-hladkov: const isCharCountValid  = charCount <= ProjectValidationRule.DESCRIPTION_MAXIMUM_LENGTH

apps/frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
```diff
@@ -43,13 +45,15 @@ const useAppForm = <T extends FieldValues = FieldValues>({
 		control,
 		formState: { errors, isDirty },
 		handleSubmit,
+		watch,
```
liza-veis, line 48: we are already using useWatch in the project, let's stick to one approach
```suggestion
```

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
liza-veis, line 68: we should already have an error under the input if we exceed the description length by the project-create validation 
```suggestion
```

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
```diff
@@ -1,29 +1,39 @@
...
 			defaultValues: DEFAULT_PROJECT_CREATE_PAYLOAD,
 			validationSchema: projectCreateValidationSchema,
-		},
-	);
+		});
+
+	const [isSubmitted, setIsSubmitted] = useState<boolean>(false);
+
+	const descriptionValue = watch("description", "");
+	const charCount = descriptionValue.length;
```
liza-veis, line 29: it would be better to rename charCount to description length, but there is no point in this because we use it only once, so let's remove it
```suggestion
	const description = watch("description", "");
```

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
```diff
@@ -45,6 +55,17 @@ const ProjectCreateForm = ({ onSubmit }: Properties): JSX.Element => {
 				name="description"
 				rowsCount={4}
 			/>
+			{charCount <= ProjectValidationRule.DESCRIPTION_MAXIMUM_LENGTH && (
```
liza-veis, line 58: create a boolean variable for this
```suggestion
			{isDescriptionCounterShown
```

apps/frontend/src/pages/projects/components/project-create-form/styles.module.css
```diff
@@ -5,6 +5,24 @@
 	align-items: flex-end;
 }
 
+.char-counter {
+	align-self: flex-start;
+	margin-top: -8px;
```
liza-veis, line 10: Why do we need this negative margin?

liza-veis (changes requested):

apps/frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
AnutkaGn: Oh, I didn't notice that, thanks

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
AnutkaGn: There is a problem with this message appearing only when the submit button is pressed (validation starts only after the button is pressed). I have an idea how to avoid doing this check. We can change the color of the counter to red when the number of entered characters exceeds 1000. And after pressing the submit button, the user will see the error text. How is this solution?

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
AnutkaGn: This variable is used to display the length of the description text entered. I can remove it and then use descriptionValue.length. Is it better?

apps/frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
```diff
@@ -43,13 +45,15 @@ const useAppForm = <T extends FieldValues = FieldValues>({
...
 	} = useForm<T>(parameters);
 
 	return {
 		control,
 		errors,
 		handleSubmit,
 		isDirty,
+		watch,
```
what1s1ove, line 56:
```suggestion
		handleWatch: watch,
```

apps/frontend/src/pages/projects/components/project-create-form/libs/enums/enums.ts
```diff
@@ -0,0 +1,4 @@
+export {
+	ProjectValidationMessage,
+	ProjectValidationRule,
+} from "@git-fit/shared";
```
what1s1ove, line 4: It should be imported from the project module on frontend

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
liza-veis: let's try to use [trigger](https://react-hook-form.com/docs/useform/trigger) from react-hook-form, to trigger validation on this field on change

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
liza-veis: if it's used in one place it will be ok

apps/frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
```diff
@@ -43,13 +45,15 @@ const useAppForm = <T extends FieldValues = FieldValues>({
...
 		handleSubmit,
+		trigger,
 	} = useForm<T>(parameters);
 
 	return {
 		control,
 		errors,
 		handleSubmit,
 		isDirty,
+		trigger,
```
liza-veis, line 56:
```suggestion
		handleTrigger,
```

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
```diff
@@ -13,12 +19,20 @@ type Properties = {
...
+		});
+
+	const descriptionValue = useFormWatch({
+		control,
+		defaultValue: "",
+		name: "description",
+	});
+
+	const isDescriptionCounterShown =
+		descriptionValue.length <= ProjectValidationRule.DESCRIPTION_MAXIMUM_LENGTH;
```
liza-veis, line 35: Maybe it would be better to check it this way:
```suggestion
	const isDescriptionCounterShown = Boolean(errors[description]?.message)
```

liza-veis (changes requested):

what1s1ove (approved):

apps/frontend/src/pages/projects/components/project-create-form/project-create-form.tsx
```diff
@@ -13,12 +19,19 @@ type Properties = {
...
-	);
+		});
+
+	const descriptionValue = useFormWatch({
+		control,
+		defaultValue: "",
+		name: "description",
+	});
+
+	const isDescriptionCounterShown = Boolean(!errors["description"]?.message);
```
liza-veis, line 34: ! is enough here, let's remove Boolean wrapper
```suggestion
	const isDescriptionCounterShown = !errors["description"]?.message;
```

apps/frontend/src/pages/projects/components/project-create-form/styles.module.css
```diff
@@ -5,6 +5,16 @@
 	align-items: flex-end;
 }
 
+.description-counter {
+	position: relative;
+	bottom: 8px;
```
liza-veis, line 10: I don't quite understand why we need this in bottom: 8px or negative margin in the layout? shouldn't it just be at the same place where error message is usually shown?

AnutkaGn: If I remove it, there will be 16 px between the counter and the block, and I need 8 px (as well as between the error message and the block)

liza-veis: Why there will be 16 px gap? is this because of flex gap on form container? if so, it's better to wrap input and counter into wrapper to group them

liza-veis (approved):

PR #223, opened 2024-09-08
feat: Add an ability to generate and retrieve project api key gf-180

GvoFor:
## Features
- create new table `project_api_keys` with the following columns `project_id`, `created_by_user_id`, `updated_by_user_id`, `id`, `created_at`, `updated_at`
- create `project_api_keys` module (on both backend and frontend) with logic for generating api key
- adjust `encryption` module: previous `encrypt` function become `hash`, and two new functions added -- `encrypt` and `decrypt`. Encryption secret key and algorithm is setted through `.env` file
- api key is injected into project when retrieving project by id (get, patch)
- add `Setup Analytics` button. On click opens `Setup Analytics` modal, where `API Key` field rendered with `Generate` button. If API key is not present "No API key" placeholder rendered, and `Generate` button is enabled. After `Generate` button pressed API key generated and displayed instead of placeholder. When api key present `Generate` button is disabled. On successful api key generation toast notification popups

## Screenshots
image
image


liza-veis:
For key encryption/decryption let's use the crypto module. Here is an [example](https://medium.com/@tony.infisical/guide-to-nodes-crypto-module-for-encryption-decryption-65c077176980) of its usage.

apps/backend/src/modules/project-api-keys/project-api-keys.ts
GvoFor, line 1: I'm not happy with the fact I'm creating `projectRepository` here (and import `ProjectRepository` from `"~/modules/projects/project.repository.js"`), and I'd like to import it from `"~/modules/projects/projects.js"`, but inside that file I need `projectApiKeyService`, and I get cyclic dependencies and one object is undefined, while creating another.

So should I keep it as it is, or there is some better solution for this `dependency injection` problem you can suggest?

apps/frontend/src/pages/project/components/output/output.tsx
GvoFor, line 1: Maybe this can be common component? Or I can avoid it adjusting `Input` somehow?

In my initial idea, this component represent some field with readonly value, with `clipboard` icon on the right side so you can copy the value to clipboard (if value present).

apps/frontend/src/pages/project/components/setup-analytics-modal/setup-analytics-modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+		hasProjectApiKey || dataStatus === DataStatus.PENDING;
+
+	const handleGenerateSubmit = useCallback(() => {
+		void dispatch(
+			projectApiKeyActions.create({
+				projectId: project.id,
+				userId: currentUser.id,
+			}),
+		);
+	}, [dispatch, project, currentUser]);
```
GvoFor, line 42: I was thinking of creating implicit form, with default values, and no Input components to change the values. But do it this way for now. 
Should I adjust this by using `useAppForm` for submittion purpose?

apps/frontend/src/pages/project/project.tsx
```diff
@@ -31,34 +42,49 @@ const Project = (): JSX.Element => {
...
+							</h3>
+							<p className={styles["project-description"]}>
+								{project.description}
+							</p>
+						</div>
+						<Button
+							label="Setup analytics"
+							onClick={onSetupAnalyticsModalOpen}
+						/>
+					</div>
```
GvoFor, line 78: Is it looks okay? or it is better to rid of two extra identation by returning empty page layout in case there is no project (`!hasProject`)

packages/shared/src/modules/project-api-keys/libs/types/project-api-key-create-response-dto.type.ts
GvoFor, line 1: I was thinking of name this type as `ProjectApiKeyGetAllItemResponseDto` to keep consistent with other modules, but there is no `getAll` method for now (and it seems that we won't add this method at all) so I decide to name it as `ProjectApiKeyCreateResponseDto`

readme.md
```diff
@@ -124,6 +123,14 @@ erDiagram
       int project_id FK
    }
 
+   project_api_keys {
+      int id PK
+      dateTime created_at
+      dateTime updated_at
+      int user_id FK
```
GvoFor, line 130: Forget to update readme after issue update

apps/frontend/src/pages/ui/ui.tsx
GvoFor, line 1: Forget to remove this changes

GvoFor (commented):
Here is some of my thoughts. Please reply for the quesions

packages/shared/src/libs/enums/exception-message.enum.ts
```diff
@@ -7,6 +7,7 @@ const ExceptionMessage = {
 	INVALID_TOKEN_NO_USER_ID: "Token is invalid: userId missing.",
 	NO_PERMISSION: "You do not have permission to access this resource.",
 	NO_TOKEN_PROVIDED: "No token provided.",
+	PROJECT_API_KEY_ALREADY_HAS: "Project already has its API key.",
```
Fjortis, line 10: I think, we should adjust it to fit better with format 
`API key for the project not found`

GvoFor: How about  `API key already exists for the project`?

Fjortis: Even better!

packages/shared/src/modules/project-api-keys/libs/types/project-api-key-create-response-dto.type.ts
liza-veis: I'm ok with this

apps/frontend/src/pages/project/project.tsx
liza-veis: I would say the second option. And from the logic we have there shouldn't be a case where it's null

apps/frontend/src/pages/project/components/setup-analytics-modal/setup-analytics-modal.tsx
liza-veis: yes, we need to handle it using useAppForm

apps/frontend/src/pages/project/components/output/output.tsx
liza-veis: We don't need this component, we should go with our usual input and form

apps/backend/src/modules/project-api-keys/project-api-key.entity.ts
```diff
@@ -0,0 +1,104 @@
...
+			encryptedKey,
+			id: null,
+			projectId,
+			updatedBy,
+		});
+	}
+
+	public toNewObject(): { encryptedKey: string } & Omit<
+		ProjectApiKeyCreateResponseDto,
+		"apiKey" | "id"
```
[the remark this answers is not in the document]
liza-veis: looks ok to me

apps/backend/src/modules/project-api-keys/project-api-keys.ts
liza-veis: we don't need either the project repository or user repository inside projectApiKeyService. 
1. we don't need to check for project existence, we can be sure that we have only real project ids in project_api_keys table and in case the key doesn't match we just return not found

2. we need to take user id not from payload but from the current user injected into the request. We can be sure it's always present for this request

apps/backend/src/db/migrations/20240904130147_add_project_api_keys_table.ts
```diff
@@ -0,0 +1,59 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "project_api_keys";
+
+const ColumnName = {
+	CREATED_AT: "created_at",
+	CREATED_BY: "created_by",
```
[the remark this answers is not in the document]
GvoFor: Renamed

apps/backend/src/modules/project-api-keys/project-api-keys.ts
GvoFor: > we can be sure that we have only real project ids in project_api_keys table and in case the key doesn't match we just return not found

First, If I delegate ids existence checking to the database (so try to insert using project/user id which is not present in DB), then id counter for `project_api_keys` table will be incremeted, no matter insertion was success or not. I'm not sure if this is a big deal, but in my opinion this is not how it should work -- so this is basicly the reason I add project/user existence checking.
Second, what return message then should be in thrown not found error? should I craete new message like `Project or user not found`?

apps/frontend/src/pages/project/components/output/output.tsx
GvoFor: There is an issue with using `Input`: API key field shouldn't be either controlled, or have errors checks, or have name, but `Input` component forces to pass `control`, `errors` and `name` props. I think this should be another component, and not `Input` one. By the way, if I start to adjusting `Input` it seems to me, that it will become overcomplicated.

liza-veis: It should be controlled, just disabled

apps/backend/src/modules/project-api-keys/project-api-key.entity.ts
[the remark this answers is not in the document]
GvoFor: I left it as it is for now

GvoFor:
For now I use `Input` with `isDisabled` prop for displaying API key field, and cause of this, it has `opacity: 0.8` and `cursor: not-allowed`. I decide not to adjust `Input` component introducing `isEditable` prop or something else, in order to rid of this "`disabled` side effects", cause this PR already have lots of changes and I don't want to overcompicate it.

So I think the adjustment (if needed) to `Input` component should be added by the developer, who will add "copy to clipboard" logic.

Or maybe I should adjust it in this PR?

apps/backend/src/libs/modules/token/base-token.module.ts
```diff
@@ -19,12 +19,19 @@ class BaseToken implements Token {
 		this.algorithm = algorithm;
 	}
 
-	public async createToken(payload: JWTPayload): Promise<string> {
-		return await new SignJWT(payload)
+	public async createToken(
+		payload: JWTPayload,
+		hasExpirationTime: boolean = true,
```
liza-veis, line 24: Let's create a separate instance of BaseToken for projectApiKey, then we won't need to pass `hasExpirationTime`, we just won't pass the expiration time during the instance creation.
```suggestion
```

apps/backend/src/modules/project-api-keys/project-api-key.service.ts
```diff
@@ -0,0 +1,123 @@
...
+		projectApiKeyRepository,
+		token,
+	}: Constructor) {
+		this.projectApiKeyRepository = projectApiKeyRepository;
+		this.token = token;
+		this.encryption = encryption;
+	}
+
+	public async create(
+		payload: { userId: number } & ProjectApiKeyCreateRequestDto,
```
liza-veis, line 39: your approach is ok, but let's do this as in other places
```suggestion
         userId: number,
		payload: ProjectApiKeyCreateRequestDto,
```

apps/backend/src/modules/project-api-keys/project-api-key.service.ts
```diff
@@ -0,0 +1,123 @@
...
+	): Promise<ProjectApiKeyCreateResponseDto> {
+		const { projectId, userId } = payload;
+
+		const isProjectApiKeyExist = Boolean(
+			await this.projectApiKeyRepository.findByProjectId(projectId),
+		);
```
liza-veis, line 45: let's do this as in other places
```suggestion
		const existingProjectApiKey = await this.projectApiKeyRepository.findByProjectId(projectId);
```

packages/shared/src/modules/project-api-keys/libs/types/project-api-key-create-request-dto.type copy.ts
liza-veis, line 1: remove `copy` in file name

apps/frontend/src/pages/project/project.tsx
```diff
@@ -31,34 +40,45 @@ const Project = (): JSX.Element => {
...
 		return <NotFound />;
 	}
 
+	if (!hasProject) {
+		return (
+			<PageLayout isLoading={isLoading}>
+				<></>
+			</PageLayout>
+		);
+	}
```
liza-veis, line 55: looks not good, we need to either remove this condition at all if this case not possible, or use `!hasProject && !isLoading` to return nothing (without layout)
```suggestion
```

apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+	return (
+		<Modal isOpened={isOpened} onClose={onClose} title="Setup Analytics">
+			<div className={styles["content"]}>
+				<form
+					className={styles["api-key-container"]}
+					onSubmit={handleGenerateSubmit}
+				>
+					<div className={styles["api-key-output"]}>
+						<Input
```
liza-veis, line 55: Not sure if we need this wrapper, can't we just use flex on container?
```suggestion
			<div className={styles["content"]}>
				<form
					className={styles["api-key-container"]}
					onSubmit={handleGenerateSubmit}
				>
						<Input
```

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -14,10 +15,30 @@ import { ProjectEntity } from "./project.entity.js";
...
+		const projectApiKey = await this.projectApiKeyService.findByProjectId(
+			project.id,
+		);
+		const apiKey = projectApiKey ? projectApiKey.apiKey : null;
+
+		return {
+			...project,
+			apiKey,
+		};
 	}
```
liza-veis, line 42: let's do this separately for each method. We could move it to a separate method if there was more then one call, but it definitely wouldn't be something like inject, it would be just - getProjectApiKey
```suggestion
```

liza-veis (changes requested):

apps/backend/src/libs/modules/encryption/base-encryption.module.ts
```diff
@@ -1,24 +1,58 @@
...
 	public async compare(data: string, hash: string): Promise<boolean> {
 		return await bcrypt.compare(data, hash);
 	}
 
-	public async encrypt(data: string): Promise<EncryptResult> {
+	public decrypt(encryptedData: string): string {
+		const decipher = crypto.createDecipheriv(
+			this.algorithm,
+			this.secret,
+			INITIAL_VECTOR,
```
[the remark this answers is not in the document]
GvoFor: No we can't. All available algorithms expected initial vector to be not null or undefined, so without it (or then it's null/undefined) the following error appears:
```ts
TypeError: Invalid initialization vector
  at Cipheriv.createCipherBase (node:internal/crypto/cipher:121:19)
...
code: 'ERR_CRYPTO_INVALID_IV'
```

apps/backend/src/modules/project-api-keys/project-api-key.service.ts
GvoFor: I can't have it separate cause `Service` interfacte says, that `create` method should accept only one argument, which is  `payload: unknown`

apps/frontend/src/pages/project/project.tsx
GvoFor: Both variants did not work -- first because this component don't know that insides of PageLayout won't be rendered, while loading is finished and project loaded successfuly, and second -- because if `hasProject` is false, `isLoading` might be true that will prevent returning, so type sysyem cannot infer that our layout insides will be rendered only when project is not null.

Of course I can use `as ProjectGetAllItemResponse` to solve the issue with types (cause we surely now that here project cannot be null), but with this code looks a bit weird (there are 4 places, there project used). So I decided to bring back initial variant with those `{hasProject && ...` check and 2 extra tabulations for now.

apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
GvoFor: You are right

liza-veis:
> For now I use Input with isDisabled prop for displaying API key field, and cause of this, it has opacity: 0.8 and cursor: not->allowed. I decide not to adjust Input component introducing isEditable prop or something else, in order to rid of this >"disabled side effects", cause this PR already has lots of changes and I don't want to overcompicate it.
>
>So I think the adjustment (if needed) to Input component should be added by the developer, who will add "copy to clipboard" logic.
>
>Or maybe I should adjust it in this PR?
>@liza-veis what do you say?

I think it will be ok if it will look like disabled inputs, if this will be critical we will add readonly mode later

apps/backend/src/modules/project-api-keys/project-api-key.service.ts
```diff
@@ -0,0 +1,122 @@
...
+				id: createdApiKey.id,
+				projectId: createdApiKey.projectId,
+				updatedByUserId: createdApiKey.updatedByUserId,
+			};
+		} catch {
+			throw new ProjectApiKeyError({
+				message: ExceptionMessage.PROJECT_NOT_FOUND,
+				status: HTTPCode.NOT_FOUND,
+			});
+		}
```
liza-veis, line 80: oh now I see what this was for. As we don't have a proper container for di, let's create another instance of projects service and pass it here and use to check if project exists

liza-veis (changes requested):

apps/backend/.env.example
```diff
@@ -24,3 +24,5 @@ JWT_ALGORITHM="HS256"
 # ENCRYPTION
 #
 ENCRYPTION_SALT_ROUNDS=10
+ENCRYPTION_ALGORITHM=aes-128-gcm
+ENCRYPTION_SECRET=123456789abcdefg
```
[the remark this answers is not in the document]
GvoFor: In case we have `aes-128-gcm` algorithm, key should consist of `128` bits, which is `16` bytes, so key length should be 16. If I use `secret_key`, crypto package throws error `Invalid key length.

I will use `secret_key_12345` then for now. Also it looks like I should leave a comment about key length

apps/backend/src/libs/modules/encryption/base-encryption.module.ts
```diff
@@ -1,24 +1,67 @@
...
+			this.outputEncoding,
+			this.inputEncoding,
+		);
+	}
+
+	public encrypt(data: string): string {
+		const chipher = crypto.createCipheriv(
+			this.algorithm,
+			this.secret,
+			this.initialVector,
```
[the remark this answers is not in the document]
GvoFor: Didn't notice this algorithm. Thanks

apps/backend/src/modules/project-api-keys/project-api-key.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+	public constructor({
+		encryption,
+		projectApiKeyRepository,
+		projectRepository,
+		token,
+	}: Constructor) {
+		this.projectApiKeyRepository = projectApiKeyRepository;
+		this.token = token;
+		this.projectRepository = projectRepository;
```
liza-veis, line 40: we need to use project service, we shouldn't access other repositories inside this service
```suggestion
		this.projectService = projectService;
```

packages/shared/src/modules/projects/libs/types/project-get-response-dto.type.ts
```diff
@@ -0,0 +1,8 @@
+type ProjectGetResponseDto = {
```
liza-veis, line 1:
```suggestion
type ProjectGetByIdResponseDto = {
```

apps/backend/.env.example
```diff
@@ -24,3 +24,7 @@ JWT_ALGORITHM="HS256"
 # ENCRYPTION
 #
 ENCRYPTION_SALT_ROUNDS=10
+ENCRYPTION_ALGORITHM=aes-128-ecb
+# Secret length should suit the algorithm:
+# if algorithm expects it to be 128 bits it should consist of 16 chars
```
liza-veis, line 29:
```suggestion
# Number of chars in the secret must meet the requirements of the algorithm.
```

liza-veis (changes requested):

apps/backend/src/modules/project-api-keys/project-api-key.service.ts
GvoFor: Sure

liza-veis: oh I see, than let's leave it

liza-veis (approved):

what1s1ove (approved):

PR #226, opened 2024-09-09
feat: select cursor pointer gf-204

perekotypole:
- fix select control hover style and add cursor pointer
- fix option hover style and add cursor pointer
- add `isSearchable` parameter to select components

Hover/focus on select control as on input elements:
Hover on select control

Hover/focus on option as on buttons:
Hover on option


apps/frontend/src/libs/components/select/select.tsx
```diff
@@ -18,6 +18,7 @@ type Properties<TFieldValues extends FieldValues, TOptionValue> = {
 	control: Control<TFieldValues, null>;
 	isLabelHidden?: boolean;
 	isMulti?: boolean;
+	isSearchable?: boolean;
```
liza-veis, line 21: Do we need to add this property here?

liza-veis (approved):

PR #237, opened 2024-09-09
feat: add date input to analytics page gf-182

Vitaliistf:
Screenshot 2024-09-09 at 21 23 20

Created a common date input component using react-calendar.
Added an analytics page and place date input there.
On the analytics page, the user should is able to select only past dates and no more than 1 month period.

Styling is done using usual css, because the library requires BEM notations for styling.


apps/frontend/src/libs/components/date-input/libs/constants/milis-in-day.constant.ts
```diff
@@ -0,0 +1,3 @@
+const MILIS_IN_DAY = 86_400_000;
```
GvoFor, line 1:
```suggestion
const MILLIS_IN_DAY = 86_400_000;
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -0,0 +1,50 @@
...
+			<section>
+				<form
+					className={styles["filters-form-wrapper"]}
+					onSubmit={handleFormSubmit}
+				>
+					<DateInput
+						control={control}
+						maxDate={new Date()}
+						maxRange={30}
+						minDate={new Date("01-01-1999")}
```
GvoFor, line 41: Not sure if this matters but have you agreed min date with QAs?

apps/frontend/src/libs/helpers/get-short-weekday.helper.ts
GvoFor, line 1: Not sure if this helper will be used anywhere else except `DateInput` component. Then it makes sense to move it to local libs of that component

apps/frontend/src/pages/analytics/analytics.tsx
Vitaliistf: Yeah, I was asking about it and Liza said that we will think about this restriction after we implement the main logic with tracking activity

apps/frontend/src/libs/helpers/get-short-weekday.helper.ts
Vitaliistf: Oh, yeah, nice notice, thanks

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,102 @@
...
+	name,
+}: Properties<T>): JSX.Element => {
+	const { field } = useFormController({
+		control,
+		name,
+	});
+
+	const { isOpened, onClose, onOpen } = usePopover();
+
+	const getSelectedDateRange = (): string => {
```
what1s1ove, line 43: a10

what1s1ove (approved):

packages/shared/src/libs/helpers/format-date/format-date.helper.ts
```diff
@@ -1,8 +1,11 @@
 import { format } from "date-fns";
 
-type DateFormat = "d MMM yyyy HH:mm";
+type DateFormat = "d MMM yyyy HH:mm" | "MMM d, yyyy";
 
-const formatDate = (date: Date, dateFormat: DateFormat): string => {
+const formatDate = (
+	date: Date,
+	dateFormat: DateFormat = "d MMM yyyy HH:mm",
```
liza-veis, line 7: let's always pass date format explicitly
```suggestion
	dateFormat: DateFormat,
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -0,0 +1,50 @@
...
+		},
+		[handleSubmit],
+	);
+
+	return (
+		<PageLayout>
+			<h1 className={styles["title"]}>Analytics</h1>
+			<section>
+				<form
+					className={styles["filters-form-wrapper"]}
```
liza-veis, line 34:
```suggestion
					className={styles["filters-form"]}
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useAppForm, useCallback } from "~/libs/hooks/hooks.js";
+
+import styles from "./styles.module.css";
+
+const getDefaultDateRange = (): Date[] => {
+	const today = new Date();
+	const startOfMonth = new Date(today.getFullYear(), today.getMonth());
+
+	return [startOfMonth, today];
+};
```
liza-veis, line 11: let's use last 30 days as default range, also we have a library for working with dates and we need to use it in case we need manipulations with the dates
```suggestion
```

apps/frontend/src/libs/components/date-input/styles.css
liza-veis, line 1: make it .module.css and use :global() for overriding library styles

apps/frontend/src/assets/images/icons/calendar.svg
liza-veis, line 1: remove fill from the icon

liza-veis (changes requested):

apps/frontend/src/libs/components/date-input/styles.css
liza-veis: let's add the following to the rules in `stylelint.config.js` (for this case it's ok, because we want to allow using global overall in the app)

```
 "selector-pseudo-class-no-unknown": [
      true,
      {
        "ignorePseudoClasses": ["global"]
      }
    ]
```

Vitaliistf: Done.

packages/shared/src/libs/helpers/subtract-days/subtract-days.helper.ts
```diff
@@ -0,0 +1 @@
+export { subDays as subtractDays } from "date-fns";
```
liza-veis, line 1: create a function wrapper for this, like for debounce or formatDate

apps/frontend/src/pages/analytics/libs/constants/thirty-days.constant.ts
```diff
@@ -0,0 +1,3 @@
+const THIRTY_DAYS = 30;
```
liza-veis, line 1: there is no point in this name 😅 , it should describe what this is used for
```suggestion
const ANALYTICS_DATE_MAX_RANGE = 30;
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -0,0 +1,42 @@
...
+	return (
+		<PageLayout>
+			<h1 className={styles["title"]}>Analytics</h1>
+			<section>
+				<form className={styles["filters-form"]} onSubmit={handleFormSubmit}>
+					<DateInput
+						control={control}
+						maxDate={new Date()}
+						maxRange={30}
+						minDate={new Date("01-01-1999")}
```
liza-veis, line 33: we don't need to set it for now
```suggestion
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+	return (
+		<PageLayout>
+			<h1 className={styles["title"]}>Analytics</h1>
+			<section>
+				<form className={styles["filters-form"]} onSubmit={handleFormSubmit}>
+					<DateInput
+						control={control}
+						maxDate={new Date()}
+						maxRange={30}
```
liza-veis, line 32:
```suggestion
						maxRange={ANALYTICS_DATE_MAX_RANGE}
```

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,104 @@
...
+					/>
+				}
+				isOpened={isOpened}
+				onClose={onClose}
+			>
+				<button className={styles["date-input-trigger"]} onClick={onOpen}>
+					<span className={styles["calendar-icon-wrapper"]}>
+						<Icon height={20} name="calendar" width={20} />
+					</span>
+					{getSelectedDateRange() || "Choose date"}
```
perekotypole, line 97:
```suggestion
					{getSelectedDateRange() ?? "Choose date"}
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useAppForm, useCallback } from "~/libs/hooks/hooks.js";
+
+import { THIRTY_DAYS } from "./libs/constants/constants.js";
+import { subtractDays } from "./libs/helpers/helpers.js";
+import styles from "./styles.module.css";
+
+const Analytics = (): JSX.Element => {
+	const { control, handleSubmit } = useAppForm({
+		defaultValues: {
+			dateRange: [subtractDays(new Date(), THIRTY_DAYS), new Date()],
```
liza-veis, line 11: move todayDate to a separate constant
```suggestion
			dateRange: [subtractDays(todayDate, THIRTY_DAYS), todayDate],
```

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,104 @@
...
+				}
+			}
+
+			field.onChange(dates);
+		},
+		[field, maxRange],
+	);
+
+	return (
+		<div>
```
liza-veis, line 71: We need to use <input /> tag  with type date for this component, and just override it's handlers

apps/frontend/src/libs/components/date-input/libs/helpers/get-short-weekday.helper.ts
```diff
@@ -0,0 +1,10 @@
+const START_OF_WORD = 0;
+const END_OF_WORD = 2;
+
+const getShortWeekday = (locale: string | undefined, date: Date): string => {
```
liza-veis, line 4: remove this helper, we need to use formatDate helper to format dates

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,104 @@
...
+			<Popover
+				content={
+					<Calendar
+						calendarType="gregory"
+						formatShortWeekday={getShortWeekday}
+						locale="en"
+						maxDate={maxDate ?? new Date()}
+						minDate={minDate ?? new Date("01-01-1999")}
```
liza-veis, line 79: we don't need default values here, by default they should be undefined if not passed

apps/frontend/src/libs/components/date-input/styles.module.css
```diff
@@ -0,0 +1,124 @@
+:global(.react-calendar) {
```
liza-veis, line 1: assign a class name to the calendar wrapper, then we should be able to use it this way, to make it less global
```suggestion
.calendar :global(.react-calendar) {
```

liza-veis (changes requested):

apps/frontend/src/libs/components/date-input/date-input.tsx
Vitaliistf: I'm not quite sure, how to do it properly, but I tried, and there will be a lot of issues if we do it like this, for example we'll have only clickable icon, also when we click on input (even if it'll be readonly) it will get focus and characters there will become selected, and the main thing is that we will not be able to fill the input with our custom date output, because it'll throw an error. I understand that my approach is not as good for semantics, but it might be more convenient like this.

Vitaliistf: Done. I'm not sure if this is what you meant, because there are not a lot of handlers to override if it is text😅. Also I'm not sure about leaving a button as a wrapper for this input and icon, but I think it will be more convenient for user to click on a whole element, and not only on that tiny calendar.

yevhenii-hladkov (approved):

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,109 @@
...
+			}
+
+			field.onChange(dates);
+		},
+		[field, maxRange],
+	);
+
+	const formatShortWeekday = useCallback(
+		(_locale: string | undefined, date: Date): string => {
+			return formatDate(date, "EEEEEE");
```
perekotypole, line 66: Maybe we can store date formats somewhere in enums, as we do with DataStatus for instance?

Vitaliistf: You're right, I'll try to change it, thanks

Vitaliistf: Okay, now I'm not so sure about it, the naming for these formats is quite complicated, because it either will need a lot of words, or if we will use something like "short date"/"long date" it won't be enough, because we might have a lot of similar formats. I guess that this type is enough for now.

apps/frontend/src/libs/components/date-input/styles.module.css
```diff
@@ -0,0 +1,144 @@
+.calendar:global(.react-calendar) {
+	display: flex;
+	flex-direction: column;
+	padding: 12px;
+}
+
+:global(.react-calendar__navigation) {
```
liza-veis, line 7: with the past comment I meant to add .calendar to all places with use :global selector
```suggestion
.calendar:global(.react-calendar__navigation) {
```

apps/frontend/src/libs/components/date-input/libs/helpers/helpers.ts
liza-veis, line 1: we need to re-export common helpers from srs/libs/helpers, not component helpers

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,109 @@
...
+	useCallback,
+	useFormController,
+	usePopover,
+} from "~/libs/hooks/hooks.js";
+
+import { addDays, getDifferenceInDays } from "./libs/helpers/helpers.js";
+import { type DateValue } from "./libs/types/types.js";
+import styles from "./styles.module.css";
+
+type Properties<T extends FieldValues> = {
```
liza-veis, line 20: we need to pass maxDate as today, I meant we don't need to pass minDate for now, because we didn't decide on restriction 😅

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,109 @@
...
+	name,
+}: Properties<T>): JSX.Element => {
+	const { field } = useFormController({
+		control,
+		name,
+	});
+
+	const { isOpened, onClose, onOpen } = usePopover();
+
+	const getSelectedDateRange = (): string => {
```
liza-veis, line 38: What if we don't have one of the values for example we selected only start date? is this case somehow prevented or handled?

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,109 @@
...
+					return;
+				}
+			}
+
+			field.onChange(dates);
+		},
+		[field, maxRange],
+	);
+
+	const formatShortWeekday = useCallback(
```
liza-veis, line 64:
```suggestion
	const handleFormatShortWeekday = useCallback(
```

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,109 @@
...
+			if (maxRange && startDate && endDate) {
+				const diffInDays = getDifferenceInDays(endDate, startDate);
+
+				if (diffInDays > maxRange) {
+					const adjustedEndDate = addDays(startDate, maxRange);
+					field.onChange([startDate, adjustedEndDate]);
+
+					return;
+				}
+			}
```
liza-veis, line 57: instead of this logic, can't we do something like this, and adjust maxDate instead?  Also is there a visual indication of the max date to be selected? for example, dates in the future are shown as disabled (grey)?

```tsx
const adjustedMaxDate = useMemo((): Date | undefined => {
  if(!startDate) {
    return maxDate;
  }

  const rangeMaxDate = addDays(startDate, maxRange);
  
  if(!maxDate) {
    return rangeMaxDate;
  }

  return maxDate?.getTime() > rangeMaxDate.getTime() ? rangeMaxDate : maxDate;
}, [startDate, maxDate, maxRange])
```

liza-veis (changes requested):

apps/frontend/src/libs/components/date-input/date-input.tsx
Vitaliistf: This is prevented by library. We cannot select only start date. However, there is an issue when we don't pass default values, or pass only one to the useAppForm hook. I'll try to do something about it.

apps/frontend/src/libs/components/date-input/date-input.tsx
Vitaliistf: Dates in future are not visible (as in the design) and we cannot navigate to months that are in future after max date.

apps/frontend/src/libs/components/date-input/libs/types/date-value.ts
```diff
@@ -0,0 +1,5 @@
+type DateValuePiece = Date | null;
+
+type DateValue = [DateValuePiece, DateValuePiece] | DateValuePiece;
```
liza-veis, line 3:
```suggestion
type DateInputValue = [DateValuePiece, DateValuePiece] | DateValuePiece;
```

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -0,0 +1,119 @@
...
+	return (
+		<div>
+			<Popover
+				content={
+					<Calendar
+						calendarType="gregory"
+						className={styles["calendar"]}
+						formatShortWeekday={handleFormatShortWeekday}
+						locale="en"
+						maxDate={maxDate ?? new Date()}
```
liza-veis, line 89: By default there shouldn't be any limitation for the future date. we need to pass this through props and include this property here, only if it's passed

liza-veis (changes requested):

apps/frontend/src/libs/components/date-input/date-input.tsx
Vitaliistf: Changed everything. Thanks!

liza-veis (approved):

PR #239, opened 2024-09-10
feat: make buttons 100% width gf-232

HoroshkoMykhailo:
Summary:

- Made the Button component 100% width by default.
- Removed the unnecessary properties from button wrappers across the app.
- Everything stayed the same.

image

image

image

image

image

image

image

image


apps/frontend/src/pages/projects/styles.module.css
```diff
@@ -25,3 +25,7 @@
 	line-height: 150%;
 	color: var(--color-text-secondary);
 }
+
+.button-wrapper {
```
liza-veis, line 29: not sure we need min-width here, on figma this button just takes the size of it's label + paddings

apps/frontend/src/pages/profile/libs/components/edit-user-form/styles.module.css
```diff
@@ -2,18 +2,17 @@
...
+	align-items: flex-start;
 }
 
 .inputs-wrapper {
 	display: flex;
 	flex-direction: column;
 	gap: 16px;
 	width: 100%;
-	max-width: 500px;
+	min-width: 500px;
```
liza-veis, line 13: I think we should keep max-width here, the problem why it's not full width is that one of the containers lacks width 100%

apps/frontend/src/pages/profile/libs/components/delete-account/styles.module.css
```diff
@@ -18,3 +18,7 @@
 	font-size: 16px;
 	line-height: 150%;
 }
+
+.button-wrapper {
+	min-width: 190px;
```
liza-veis, line 23: the same, on figma this button just takes it's content width, we need to use flex here to align item to one of the sides

liza-veis (changes requested):

apps/frontend/src/libs/components/button/styles.module.css
```diff
@@ -1,12 +1,14 @@
 .button {
+	display: block;
 	align-content: center;
+	width: 100%;
 	min-height: 44px;
 	padding: 10px 16px;
-	font-family: Inter, sans-serif;
 	font-size: 16px;
```
liza-veis, line 7: why did you remove font-family here?

what1s1ove (approved):
lgtm after Lisa's comments

PR #251, opened 2024-09-10
fix: name field doesn`t accept hyphen on Sign-Up page gf-236

kolibri753:
Changes:

- added hyphen to NAME_PATTERN (UserValidationRule)


liza-veis (approved):

packages/shared/src/modules/users/libs/enums/user-validation-rule.enum.ts
```diff
@@ -3,7 +3,7 @@ const UserValidationRule = {
 	EMAIL_MINIMUM_LENGTH: 6,
 	NAME_MAXIMUM_LENGTH: 50,
 	NAME_MINIMUM_LENGTH: 3,
-	NAME_PATTERN: /^[\sA-Za-z]*$/,
+	NAME_PATTERN: /^[\sA-Za-z-]*$/,
```
kolibri753, line 6: should we also ensure that hyphen is surrounded by characters?
`NAME_PATTERN: /^(?!.*--)[A-Za-z]+(-[A-Za-z]+)*$/,`

kolibri753: Okay, I see. We don’t want to complicate it. Got it :+1:

what1s1ove (approved):

PR #253, opened 2024-09-10
feat: add view all projects permission gf-230

sandrvvu:
+ Add view_all_projects permission (name: "View All Projects") to the permissions table.
+ Only users with this permission should be able to see projects item in the sidebar and access GET /projects and GET /projects/:id routes.
+ Add description counter to the update project form.

User with "View All Projects" and "Access Managment" permission
image

User with one "Access Managment" permission (is redirected here after login)
image

User without permissions (placeholder home page). Maybe we can use this page for some imformation about the app etc.
image


what1s1ove (approved):

apps/frontend/src/pages/auth/auth.tsx
```diff
@@ -24,13 +27,21 @@ const Auth = (): JSX.Element => {
...
+	}, [authenticatedUser]);
+
 	const navigate = useNavigate();
 
 	useEffect(() => {
 		if (authenticatedUser) {
-			navigate(AppRoute.ROOT);
+			navigate(
+				getFirstAvailableRoute(userPermissions, SIDEBAR_ITEMS, AppRoute.ROOT),
+			);
```
liza-veis, line 42: let's navigate to the projects here, and handle redirect to the first available item inside ProtectedRoute
```suggestion
			navigate(
				navigate(AppRoute.PROJECTS);
			);
```

apps/frontend/src/libs/helpers/get-first-available-route.ts
```diff
@@ -0,0 +1,26 @@
+import {
+	type NavigationItem,
+	type PermissionGetAllItemResponseDto,
+} from "~/libs/types/types.js";
+
+const getFirstAvailableRoute = (
```
liza-veis, line 6: instead, let's have another helper and then use it inside the sidebar and protected route
```ts
const getPermittedNavigationItems = (items: NavigationItem[], userPermissions: PermissionGetAllItemResponseDto[]) => {
   return items.filter(({ pagePermissions = [] }) => checkHasPermission(pagePermissions, userPermissions),
   );
}
```

liza-veis (changes requested):

apps/frontend/src/libs/helpers/get-first-available-route.ts
liza-veis, line 1: file should be in `libs/helpers/get-permitted-navigation-items/get-permitted-navigation-items.helper.ts`

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -41,11 +50,18 @@ const ProtectedRoute = ({
...
 	if (!hasRequiredPermission) {
-		return <NotFound />;
+		const FIRST_ITEM_INDEX = 0;
+
+		const firstAvailableRoute =
+			getPermittedNavigationItems(SIDEBAR_ITEMS, userPermissions)[
+				FIRST_ITEM_INDEX
+			]?.href || AppRoute.ROOT;
+
+		return <Navigate replace to={firstAvailableRoute} />;
```
liza-veis, line 64: we can simplify this much and make more readable
```suggestion
		const [navigationItem] = getPermittedNavigationItems(SIDEBAR_ITEMS, userPermissions);
		const redirectLink = navigationItem?.href ?? AppRoute.ROOT;

		return <Navigate replace to={redirectLink} />;
```

liza-veis (changes requested):

liza-veis (changes requested):
We also need to redirect to AppRoute.PROJECTS instead of root on NotFound page and inside breadcrumbs on project details page

PR #256, opened 2024-09-10
fix: resctict user manage group gf-254

AnutkaGn:
Fix the issue with group creation. Only users with "manage_user_access" permission can create a group.


apps/backend/src/libs/hooks/check-user-permissions.hook.ts
```diff
@@ -6,27 +6,39 @@ import { type APIPreHandler } from "~/libs/modules/controller/controller.js";
 import { HTTPCode, HTTPError } from "~/libs/modules/http/http.js";
 
 const checkUserPermissions = (routePermissions: string[]): APIPreHandler => {
-	return (request: FastifyRequest): void => {
+	return (request: FastifyRequest): Promise<void> => {
```
liza-veis, line 9: I think there is no need for this promise wrapper. As I remember we should receive `(request, response, done)` as parameters and we just need to call done at the end

liza-veis (changes requested):

liza-veis (approved):

PR #273, opened 2024-09-11
fix: inconsistent layout of the users table gf-267

s1rserg:
* Set fixed column sizes in users table according to mockup.
image


perekotypole:
It won't work properly on screens smaller than about 1360px.
image

s1rserg:
> It won't work properly on screens smaller than about 1360px. image

Can you explain what's wrong a bit more, please?

liza-veis (approved):

perekotypole:
> Can you explain what's wrong a bit more, please?

The design is not responsive for smaller screens.

PR #278, opened 2024-09-11
fix: The Table should not be able to scroll when the Options Popover is open gf-272

kolibri753:
Changes:

- removed table scroll when the popover menu is opened;

### Screenshots:
image

image


apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -44,8 +48,19 @@ const Table = <T extends object>({
...
 
+	const tableContainerReference = useRef<HTMLDivElement | null>(null);
+	const [isScrollPresent, setIsScrollPresent] = useState<boolean>(false);
+
+	useEffect(() => {
+		const container = tableContainerReference.current;
+
+		if (container) {
+			setIsScrollPresent(container.scrollHeight >= TABLE_MAX_HEIGHT);
+		}
```
kolibri753, line 59: Maybe instead of TABLE_MAX_HEIGHT, I should get it dynamically

apps/frontend/src/libs/components/table/styles.module.css
```diff
@@ -51,6 +51,11 @@
 	overflow-y: auto;
 }
 
+.table .table-body--no-scroll {
+	padding-right: 10px;
+	overflow-y: hidden;
+}
```
kolibri753, line 57: I considered adding a line using `::after`, but we can't apply `position: relative` to the table as it would disrupt the popover. I can set the width to 101% to account for the scrollbar space, but this would result in a slight movement, which I think is better to avoid

apps/frontend/src/pages/access-management/libs/components/groups-table/groups-table.tsx
```diff
@@ -28,20 +29,48 @@ const GroupsTable = ({
...
-		onDelete,
-		onEdit: handleEdit,
-	});
+	const handleEdit = useCallback(
+		(groupId: number): void => {
+			const group = groups.find(({ id }) => id === groupId);
+			const hasGroup = Boolean(group);
+
+			if (hasGroup) {
+				onEdit(group as GroupGetAllItemResponseDto);
```
Vitaliistf, line 38: I think we can omit this constant, can't we?
```suggestion
			if (group) {
				onEdit(group as GroupGetAllItemResponseDto);
```

kolibri753: I thought that the variable name provided more clarity. But yes, you're right, I think it’s even better to write it like this:
```
if (group) {
	onEdit(group);
}
```

what1s1ove (approved):

apps/frontend/src/libs/components/table/libs/constants/table-max-height.constant.ts
```diff
@@ -0,0 +1,3 @@
+const TABLE_MAX_HEIGHT = 490;
```
GvoFor, line 1: Where this number comes from?

kolibri753: The max width of table's body + height of header

kolibri753: I think to change this to just table's body (450) and pass it to the style also

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -44,6 +48,17 @@ const Table = <T extends object>({
...
+	const tableBodyReference = useRef<HTMLTableSectionElement | null>(null);
+	const [isScrollPresent, setIsScrollPresent] = useState<boolean>(false);
+
+	useEffect(() => {
+		const tableBody = tableBodyReference.current;
+
+		if (tableBody) {
+			setIsScrollPresent(tableBody.scrollHeight > TABLE_BODY_MAX_HEIGHT);
+		}
+	}, [data, isMenuOpened]);
```
liza-veis, line 60: I think we don't need to calculate all of this, also the table doesn't know about menus or anything specific. Instead, let's just pass isScrollDisabled to the table, also there is no need to check if the scroll was present before or not. This complicates the logic to much, it's ok if we remove padding in case we implement this option, I think it's not that critical.

liza-veis (changes requested):

apps/frontend/src/libs/components/table/styles.module.css
```diff
@@ -51,6 +51,10 @@
 	overflow-y: auto;
 }
 
+.table .table-body--no-scroll {
```
liza-veis, line 54: we don't use bem in project
```suggestion
.table .table-body-no-scroll {
```

yevhenii-hladkov (approved):

liza-veis:
@GvoFor Do we still need this fix with your changes?

GvoFor:
I haven't add table scroll hidding yet, so It whould be nice, if we merge this, so I can only adjust `...--no-scroll` style and not rewrite everything.
However I'm not sure for now. I'll write later, when I will be sure

GvoFor:
@liza-veis I think I will need these changes, so can you merge them please

liza-veis (approved):

PR #279, opened 2024-09-11
feat: not return description on get all request gf-259

AnutkaGn:
Do not return project description from the backend on GET /projects endpoint
Changed getting the project in the modal edit window

### DIFF
--- a/apps/backend/src/modules/projects/project.entity.ts
+++ b/apps/backend/src/modules/projects/project.entity.ts
@@ -1,6 +1,6 @@
 import { type Entity } from "~/libs/types/types.js";
 
-import { type ProjectGetAllItemResponseDto } from "./libs/types/types.js";
+import { type ProjectGetByIdResponseDto } from "./libs/types/types.js";
 
 class ProjectEntity implements Entity {
 	private description!: string;
@@ -63,7 +63,10 @@ class ProjectEntity implements Entity {
 		};
 	}
 
-	public toObject(): { description: string } & ProjectGetAllItemResponseDto {
+	public toObject(): Pick<
+		ProjectGetByIdResponseDto,
+		"description" | "id" | "name"
+	> {
 		return {
 			description: this.description,
 			id: this.id as number,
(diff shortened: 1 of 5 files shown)


apps/backend/src/modules/projects/project.entity.ts
```diff
@@ -63,7 +63,7 @@ class ProjectEntity implements Entity {
 		};
 	}
 
-	public toObject(): ProjectGetAllItemResponseDto {
+	public toObject(): { description: string } & ProjectGetAllItemResponseDto {
```
perekotypole, line 66: How about creating a separate type for a full project entity and use `Pick` for `ProjectGetAllItemResponseDto`?

AnutkaGn: I think this is not a good solution, because ProjectGetAllItemResponseDto is used in other places where the description of the project is not needed

liza-veis: I think we need to use ProjectGetByIdResponseDto we should already have it

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -168,10 +180,10 @@ const Projects = (): JSX.Element => {
 				onClose={handleEditModalClose}
 				title="Update project"
 			>
-				{selectedProject && (
+				{project && projectStatus === DataStatus.FULFILLED && (
```
liza-veis, line 183:
```suggestion
				{isUpdateFormShown && (
```

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -39,8 +39,20 @@ const Projects = (): JSX.Element => {
...
+		dataStatus,
+		project,
+		projectCreateStatus,
+		projectPatchStatus,
+		projects,
+		projectStatus,
+	} = useAppSelector(({ projects }) => projects);
+
+	useEffect(() => {
+		if (selectedProject !== null) {
```
liza-veis, line 52: Maybe we need to save only project id in this case? and let's rename it to projectToModifyId (it's ok if we will use it for deletion too). That's how we named these variables in similar cases
``` suggestion
		if (projectToModifyId !== null) {
```

liza-veis (changes requested):

what1s1ove (approved):

liza-veis (approved):

PR #280, opened 2024-09-11
feat: Add infinite scroll to the projects list gf-276

perekotypole:
- implemented "load more" functionality for projects
- added observer hook to trigger loading data

image

image


apps/frontend/src/libs/hooks/use-intersection-observer/use-intersection-observer.hook.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useRef } from "react";
```
AnutkaGn, line 1: Import this from `~/libs/hooks/hooks.js`

apps/frontend/src/libs/hooks/use-intersection-observer/use-intersection-observer.hook.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const FIRST_ITEM = 0;
```
AnutkaGn, line 3: Maybe it's better to put it in a separate file

perekotypole: I thought about it too, but I don't see any reason for it. We already used a similar local constant in the avatar component, so I decided to do the same.
```
const FIRST_LETTER_INDEX = 0;

const Avatar = ({ imageUrl, name }: Properties): JSX.Element => {
	const firstLetter = name[FIRST_LETTER_INDEX];
	const hasImage = Boolean(imageUrl);
```

apps/frontend/src/modules/projects/libs/constants/default-start.constant.ts
```diff
@@ -0,0 +1,3 @@
+const DEFAULT_START: number = 0;
```
yevhenii-hladkov, line 1:
```suggestion
const DEFAULT_START = 0;
```

apps/frontend/src/modules/projects/libs/constants/infinite-scroll-load-count.constant.ts
```diff
@@ -0,0 +1,3 @@
+const INFINITE_SCROLL_LOAD_COUNT: number = 10;
```
yevhenii-hladkov, line 1:
```suggestion
const INFINITE_SCROLL_LOAD_COUNT = 10;
```

what1s1ove (approved):

yevhenii-hladkov (approved):

apps/frontend/src/libs/hooks/use-intersection-observer/use-intersection-observer.hook.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useRef } from "~/libs/hooks/hooks.js";
+
+const FIRST_ITEM = 0;
+
+type Properties = {
```
liza-veis, line 5: remove unused properties

apps/frontend/src/libs/hooks/use-intersection-observer/use-intersection-observer.hook.ts
```diff
@@ -0,0 +1,46 @@
...
+	onIntersect: () => void;
+	root?: Element | null;
+	rootMargin?: string;
+	threshold?: number;
+};
+
+const useIntersectionObserver = <T extends HTMLElement>({
+	isEnabled = true,
+	onIntersect,
+}: Properties): React.MutableRefObject<null | T> => {
```
liza-veis, line 16: let's return an object here with reference property

apps/frontend/src/libs/hooks/use-intersection-observer/use-intersection-observer.hook.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useRef } from "~/libs/hooks/hooks.js";
+
+const FIRST_ITEM = 0;
+
+type Properties = {
+	isEnabled?: boolean;
```
liza-veis, line 6: let's switch it to isDisabled, to make it more clear
```suggestion
	isDisabled?: boolean;
```

apps/frontend/src/libs/hooks/use-intersection-observer/use-intersection-observer.hook.ts
```diff
@@ -0,0 +1,46 @@
...
+	const sentinelReference = useRef<null | T>(null);
+
+	useEffect(() => {
+		if (!isEnabled || !sentinelReference.current) {
+			return;
+		}
+
+		const observer = new IntersectionObserver(
+			(entries: IntersectionObserverEntry[]): void => {
+				const firstEntry = entries[FIRST_ITEM];
```
liza-veis, line 26:
```suggestion
			(entries: IntersectionObserverEntry[]): void => {
				const [entry] = entries;
```

apps/frontend/src/modules/projects/slices/actions.ts
```diff
@@ -30,6 +35,20 @@ const loadAll = createAsyncThunk<
 >(`${sliceName}/load-all`, async (query, { extra }) => {
 	const { projectApi } = extra;
 
+	return await projectApi.getAll({
+		limit: INFINITE_SCROLL_LOAD_COUNT,
+		name: query ?? "",
+		start: DEFAULT_START,
```
liza-veis, line 41: We need to pass page, pageSize from the component. Let's create an additional hook, like useInfiniteScroll, which will do something similar to usePagination, but much simpler - without query params, and without an ability to change pageSize

liza-veis (changes requested):

packages/shared/src/modules/projects/libs/types/project-get-all-request-dto.type.ts
```diff
@@ -1,5 +1,7 @@
+import { type InfiniteScrollQueryParameters } from "src/libs/types/types.js";
+
 type ProjectGetAllRequestDto = {
 	name: string;
-};
+} & InfiniteScrollQueryParameters;
```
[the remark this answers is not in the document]
perekotypole: > we should decide how to display it on the frontend
What exactly is the problem?

apps/frontend/src/pages/projects/projects.tsx
```diff
@@ -47,27 +50,53 @@ const Projects = (): JSX.Element => {
...
-		void dispatch(projectActions.loadAll(search));
-	}, [dispatch, search]);
+		void dispatch(
+			projectActions.loadAll({
+				name: search,
+				page,
+				pageSize,
+			}),
+		);
+	}, [dispatch, search, page, pageSize]);
```
perekotypole, line 85: There is a problem with extra requests on search changing before page has changed. I have to remove search dependency to make it work properly, but I can't. So, I need some help with this

apps/frontend/src/modules/projects/slices/actions.ts
```diff
@@ -24,13 +25,18 @@ const getById = createAsyncThunk<
...
 	AsyncThunkConfig
 >(`${sliceName}/load-all`, async (query, { extra }) => {
 	const { projectApi } = extra;
 
-	return await projectApi.getAll(query);
+	const result = await projectApi.getAll(query);
+
+	return {
+		...result,
+		page: query.page,
```
liza-veis, line 38: looks like we don't need to return this as result of action, we can access page as action arguments inside reducer
```suggestion
```

apps/frontend/src/pages/projects/projects.tsx
liza-veis: Why can't you remove it? is it a linter error or something else?

packages/shared/src/modules/projects/libs/types/project-get-all-request-dto.type.ts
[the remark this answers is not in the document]
liza-veis: I meant that it's better to have common pagination on the backend, and on the frontend we will decide if we want to display it as infinite scroll or usual pagination. You already implemented this.

apps/frontend/src/pages/projects/projects.tsx
perekotypole: Yes, I get warning
```
React Hook useEffect has missing dependencies: 'dispatch', 'page', 'pageSize', and 'search'. Either include them or remove the dependency array
```

apps/frontend/src/modules/projects/slices/actions.ts
perekotypole: like this?
```
const { page } = action.meta.arg
```

apps/frontend/src/modules/projects/slices/project.slice.ts
```diff
@@ -41,17 +44,24 @@ const { actions, name, reducer } = createSlice({
...
 		builder.addCase(loadAll.pending, (state) => {
 			state.dataStatus = DataStatus.PENDING;
 		});
 		builder.addCase(loadAll.fulfilled, (state, action) => {
-			state.projects = action.payload.items;
+			const { items, totalItems } = action.payload;
+			const { page } = action.meta.arg;
+
+			state.projects =
+				page === EMPTY_LENGTH ? items : [...state.projects, ...items];
```
GvoFor, line 56: I think you wanted to use your `const FIRST_PAGE = 0` instead. This is more correct semantically
```suggestion
				page === FIRST_PAGE ? items : [...state.projects, ...items];
```

apps/frontend/src/libs/hooks/use-infinity-scroll/use-infinity-scroll.hook.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+const useInfiniteScroll: UseInfiniteScroll = ({
+	pageSize,
+	totalItemsCount,
+}) => {
+	const [page, setPage] = useState<number>(FIRST_PAGE);
+
+	const hasNextPage = useMemo(
+		() => (page + PAGE_INCREMENT) * pageSize < totalItemsCount,
```
GvoFor, line 23: Not sure if this "formula" is correct.

For example, let assume that we have 15 items in total. Page size is 10. Let the page be FIRST_PAGE (i.e. 0).
Then hasNextPage will be true (that's correct): (0+1)*10 < 15.
But if we load first page, then it wil be 1, and hasNextPage now will be false: (1+1)*10 >= 15 -- that's seems to be not right, cause we have 5 extra items, which should be on the next page, so hasNextPage should be true.

Maybe I don't fully understand the flow, but pay attention to this.

apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -51,10 +61,11 @@ class ProjectRepository implements Repository {
 			query.whereILike("name", `%${name}%`);
 		}
 
-		const projects = await query.execute();
+		const { results, total } = await query.page(page, pageSize);
```
GvoFor, line 64: Maybe lets add `.execute()` to the end?

apps/frontend/src/libs/hooks/use-infinity-scroll/use-infinity-scroll.hook.ts
perekotypole: The numbering starts from 0, so page 1 will be the second. On page 0 it will load 10 items, on page 1 - the next 5

apps/backend/src/modules/projects/project.repository.ts
perekotypole: The query is executed when you call .page(). Why do we need an additional process?

liza-veis (approved):

apps/frontend/src/pages/projects/projects.tsx
liza-veis: let's use eslint disable next line, only for this case 🥲

liza-veis: hm, got this, then looks like there should be another solution to this issue, maybe we need to return isIntersect from the intersection hook and then pass it to this useEffect?

perekotypole: It works without the extra request if to change order in code

image

perekotypole: I have solution with adding `const [isLoaded, setIsLoaded] = useState<boolean>(false);` into `useInfiniteScroll` and set it `false` on page changing and `true` in loading dispatch:
```
	useEffect(() => {
		if (!isLoaded) {
			void dispatch(
				projectActions.loadAll({
					name: search,
					page,
					pageSize,
				}),
			);
			
			setIsLoaded(true)
		}
	}, [dispatch, page, pageSize, search]);
```

liza-veis: Maybe we can just use isLoading from the statuses we have in slice?

perekotypole: or pass the loading function into `useInfiniteScroll` as parameter on page changing:
```
	const loadProjects = useCallback((page: number, pageSize: number) => {
		void dispatch(
			projectActions.loadAll({
				name: search,
				page,
				pageSize,
			}),
		);
	}, [dispatch, search]);

	const { hasNextPage, onPageChange, onPageReset, isLoaded,
		setIsLoaded, page, pageSize } =
		useInfiniteScroll({
			pageSize: PROJECTS_PAGE_SIZE,
			totalItemsCount: projectsTotalCount,
			handleLoading: loadProjects,
		});
```

apps/frontend/src/libs/hooks/use-infinity-scroll/use-infinity-scroll.hook.ts
GvoFor: Oh, yeah, I took it into account. I just got confused while writing example. Your formula is correct :)

apps/frontend/src/pages/projects/projects.tsx
liza-veis: then let's implement your suggestion above

perekotypole: already did it

PR #291, opened 2024-09-12
fix: Upon a click on the Logo element in the header user is redirecter to the Homepage gf-285

perekotypole:
- added `isAvailableWithoutPermission` property for protected routes to access only for users without permission
- notification about a lack of permissions

image

### DIFF
--- a/apps/frontend/src/index.tsx
+++ b/apps/frontend/src/index.tsx
@@ -55,7 +55,7 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 							},
 							{
 								element: (
-									<ProtectedRoute isEnableWithoutPermission>
+									<ProtectedRoute isAvailableWithoutPermission>
 										<Home />
 									</ProtectedRoute>
 								),
--- a/apps/frontend/src/index.tsx
+++ b/apps/frontend/src/index.tsx
@@ -55,7 +55,7 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 							},
 							{
 								element: (
-									<ProtectedRoute>
+									<ProtectedRoute isEnableWithoutPermission>
 										<Home />
 									</ProtectedRoute>
 								),
--- a/apps/frontend/src/libs/enums/notification-message.enum.ts
+++ b/apps/frontend/src/libs/enums/notification-message.enum.ts
@@ -1,4 +1,6 @@
 const NotificationMessage = {
+	ACCESS_DENIED:
+		"You don't have access to any pages for now. Please contact an admin to enable access for you.",
 	GROUP_CREATE_SUCCESS: "Group was successfully created.",
 	GROUP_DELETE_SUCCESS: "Group was successfully deleted.",
 	GROUP_UPDATE_SUCCESS: "Group was successfully updated.",
--- a/apps/frontend/src/libs/components/protected-route/protected-route.tsx
+++ b/apps/frontend/src/libs/components/protected-route/protected-route.tsx
@@ -20,13 +20,13 @@ import styles from "./styles.module.css";
 
 type Properties = {
 	children: React.ReactNode;
-	isEnableWithoutPermission?: boolean;
+	isAvailableWithoutPermission?: boolean;
 	routePermissions?: ValueOf<typeof PermissionKey>[];
 };
 
 const ProtectedRoute = ({
 	children,
-	isEnableWithoutPermission = false,
+	isAvailableWithoutPermission = false,
 	routePermissions = [],
 }: Properties): JSX.Element => {
 	const { authenticatedUser, dataStatus } = useAppSelector(({ auth }) => auth);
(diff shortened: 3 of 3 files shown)


apps/frontend/src/libs/enums/notification-message.enum.ts
```diff
@@ -1,4 +1,6 @@
 const NotificationMessage = {
+	ACCESS_DENIED:
+		"You don't have access to any pages for now. Please contact an admin to enable access for you.",
```
Fjortis, line 3:
```suggestion
		"You do not have permission to access this page. Please contact an administrator to request access.",
```

perekotypole: We need to coordinate this with QA since Veronika wrote this text for the notification.

apps/frontend/src/index.tsx
```diff
@@ -55,7 +55,7 @@ createRoot(document.querySelector("#root") as HTMLElement).render(
 							},
 							{
 								element: (
-									<ProtectedRoute>
+									<ProtectedRoute isAvailableWithoutPermission>
```
what1s1ove, line 58:
```suggestion
									<ProtectedRoute>
```
ProtectedRoute cannot be available without permissions

perekotypole: However, this page must be available to authorised users who have not been assigned additional permissions, but not to others.

apps/frontend/src/libs/enums/notification-message.enum.ts
VeronikaHordus: Oh this is totally fine, I have written the basic logic. You can take the one that Joel have suggested

liza-veis:
I believe the message should be placed on the page itself, not like a notification. Also, I don't like the way it's handled now. Let's do the following:
1. For the root route instead of the home page, let's return Navigate to the "/projects" page.
2. Rename the home page, to something like - no access page.
3. Place the message about no access somewhere in the center of a page. This page should look similar to Not Found page, but without `back to home` button

Fjortis:
not a code issue, but title is misspelled, 

> redirecter 

should be 

> redirected

.

apps/frontend/src/index.tsx
Fjortis: we can simplify the logic by checking if routePermissions are empty. this would ensure that authenticated users can access the page without requiring additional permissions while still redirecting unauthenticated users to the sign-in page.

apps/frontend/src/pages/no-access/no-access.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const NoAccess = (): JSX.Element => {
+	return (
+		<PageLayout>
+			<div className={styles["container"]}>
+				<h1 className={styles["title"]}>
+					You do not have permission to access this page.
+				</h1>
+				<p className={styles["text"]}>
+					Please contact an administrator to request access.
```
liza-veis, line 13: let's slightly rename it. Also please set max-width and align text to the center, so it looks pretty and not too wide 
```suggestion
					Pending access
				</h1>
				<p className={styles["text"]}>
					You don't have access to any pages for now. Please contact an admin to enable access for you.
```

liza-veis (changes requested):

yevhenii-hladkov (approved):

apps/frontend/src/index.tsx
```diff
@@ -1,5 +1,6 @@
 import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
+import { Navigate } from "react-router-dom";
```
liza-veis, line 3: let's re-export it from common components
```suggestion
import { Navigate } from "src/libs/components/components";
```

apps/frontend/src/pages/no-access/no-access.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import styles from "./styles.module.css";
+
+const NoAccess = (): JSX.Element => {
+	return (
+		<PageLayout>
+			<div className={styles["container"]}>
+				<h1 className={styles["title"]}>Pending access</h1>
+				<p className={styles["text"]}>
+					You do not have access to any pages for now. Please contact an admin
```
liza-veis, line 11: I don't see any styles added that will restrict the text to have fixed container and not going full page width 🤔

liza-veis (changes requested):

apps/frontend/src/index.tsx
perekotypole: Should I add it there?

apps/frontend/src/pages/no-access/no-access.tsx
perekotypole: Do I have to add something like `max-width: 800px;`?

liza-veis: yes, I think 800 is too much, 400-450 should be enough

apps/frontend/src/index.tsx
liza-veis: yes, let's re-export it from common components, the same as we do for NavLink

what1s1ove (approved):

liza-veis (approved):

PR #292, opened 2024-09-12
feat: Add an ability to copy project api key gf-284

kolibri753:
Added an ability to copy a project api key.

image

image


apps/frontend/src/libs/components/input/input.tsx
```diff
@@ -16,12 +16,14 @@ type Properties<T extends FieldValues> = {
...
 	isDisabled?: boolean;
 	isLabelHidden?: boolean;
+	isReadOnly?: boolean;
 	label: string;
 	leftIcon?: JSX.Element;
 	name: FieldPath<T>;
 	placeholder?: string;
 	rightIcon?: JSX.Element;
 	rowsCount?: number;
+	textOverflow?: "clip" | "ellipsis";
```
liza-veis, line 26: let's remove this, I believe we don't want to use ellipsis for inputs, because we want to be able to copy the value inside it (even without copy button).
```suggestion
```

apps/frontend/src/modules/project-api-keys/slices/actions.ts
```diff
@@ -27,4 +27,16 @@ const create = createAsyncThunk<
 	return response;
 });
 
-export { create };
+const copy = createAsyncThunk<string, string, AsyncThunkConfig>(
```
liza-veis, line 30: let's make it more clear
```suggestion
const copyToClipboard = createAsyncThunk<string, string, AsyncThunkConfig>(
```

apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
```diff
@@ -44,6 +49,12 @@ const SetupAnalyticsModal = ({
 		[handleSubmit, dispatch],
 	);
 
+	const handleCopyClick = useCallback(() => {
+		if (hasProjectApiKey) {
+			void dispatch(projectApiKeyActions.copy(project.apiKey as string));
+		}
```
liza-veis, line 55: let's add isDisabled to icon button instead
```suggestion
		void dispatch(projectApiKeyActions.copy(project.apiKey as string));
```

liza-veis (changes requested):

apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
```diff
@@ -54,10 +65,19 @@ const SetupAnalyticsModal = ({
 					<Input
 						control={control}
 						errors={errors}
-						isDisabled
+						isDisabled={!hasProjectApiKey}
+						isReadOnly={hasProjectApiKey}
```
liza-veis, line 69: let's use readonly for both cases
```suggestion
						isReadOnly={true}
```

apps/frontend/src/libs/components/input/input.tsx
kolibri753: If I'm not mistaken ellipsis also allow users to copy text, but yeah, let's do this to avoid confusion

what1s1ove (approved):

liza-veis (approved):

PR #293, opened 2024-09-12
fix: The page reloads after generating the API key gf-289

GvoFor:
Make the page not reload while/after API key generated

### DIFF
--- a/apps/frontend/src/modules/project-api-keys/slices/actions.ts
+++ b/apps/frontend/src/modules/project-api-keys/slices/actions.ts
@@ -16,13 +16,12 @@ const create = createAsyncThunk<
 	AsyncThunkConfig
 >(`${sliceName}/create`, async (payload, { dispatch, extra }) => {
 	const { projectApiKeysApi, toastNotifier } = extra;
-	const { projectId } = payload;
 
 	const response = await projectApiKeysApi.create(payload);
 
 	toastNotifier.showSuccess(NotificationMessage.PROJECT_API_KEY_CREATE_SUCCESS);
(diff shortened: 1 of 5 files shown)


apps/frontend/src/modules/project-api-keys/slices/actions.ts
```diff
@@ -16,13 +16,12 @@ const create = createAsyncThunk<
...
 >(`${sliceName}/create`, async (payload, { dispatch, extra }) => {
 	const { projectApiKeysApi, toastNotifier } = extra;
-	const { projectId } = payload;
 
 	const response = await projectApiKeysApi.create(payload);
 
 	toastNotifier.showSuccess(NotificationMessage.PROJECT_API_KEY_CREATE_SUCCESS);
 
-	void dispatch(projectActions.getById({ id: String(projectId) }));
+	void dispatch(projectActions.updateApiKey(response.apiKey));
```
perekotypole, line 24: What is the point of `updateApiKey` action if it just triggers `state.project.apiKey = action.payload;`  which can be included to `create`? Or am I misunderstanding something?

liza-veis, line 24: I think we don't need a separate action for this, we return response in this action, so we can access it inside create.success case in reducer. It's ok if we use action from another module inside projects, just import it the same way as we do for components.

liza-veis (changes requested):

what1s1ove (approved):

liza-veis (approved):

PR #294, opened 2024-09-12
fix: The users are not sorted correctly at the "Users" table in the "Access Managment" page gf-290

perekotypole:
- add users list sorting by createdAt

### DIFF
--- a/apps/backend/src/modules/users/user.repository.ts
+++ b/apps/backend/src/modules/users/user.repository.ts
@@ -1,3 +1,4 @@
+import { SortType } from "~/libs/enums/enums.js";
 import {
 	type PaginationQueryParameters,
 	type PaginationResponseDto,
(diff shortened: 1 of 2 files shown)


apps/backend/src/modules/users/user.repository.ts
```diff
@@ -64,6 +64,7 @@ class UserRepository implements Repository {
 	}: PaginationQueryParameters): Promise<PaginationResponseDto<UserEntity>> {
 		const { results, total } = await this.userModel
 			.query()
+			.orderBy("createdAt", "desc")
```
GvoFor, line 67: We have `SortType.DESCENDING`
```suggestion
			.orderBy("createdAt", SortType.DESCENDING)
```

liza-veis (approved):

PR #300, opened 2024-09-12
fix: Group and permission names are broken in the middle of a word gf-299

s1rserg:
* Refactored table styles to break word only if it can't fit in entire line. Before it broke words even if they could fit perfectly on the next line.
## Before
image
## After 
image
image


liza-veis:
Maybe we can make the permissions column bigger and created at column smaller?

s1rserg:
> Maybe we can make the permissions column bigger and created at column smaller?

This is done here #304. But if it's needed I can get it done here

what1s1ove (approved):

PR #303, opened 2024-09-12
feat: add users search in group forms gf-270

sandrvvu:
Add search input for users table to create / edit group modal.
image
image
image


what1s1ove (approved):

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -48,12 +54,19 @@ const GroupUsersTable = ({
 		totalItemsCount: usersTotalCount,
 	});
 
+	const handleSearchChange = useCallback(
```
Fjortis, line 57: Should we add a debounce mechanism in the onChange handler to delay search queries until the user has stopped typing for a short period?

packages/shared/src/modules/users/libs/types/user-get-all-query-parameters-required-name.type.ts
liza-veis, line 1: can't we use one of the types instead of these two separate?

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -62,33 +75,47 @@ const GroupUsersTable = ({
...
+								onPageSizeChange={onPageSizeChange}
+								page={page}
+								pageSize={pageSize}
+								totalItemsCount={usersTotalCount}
+							/>
+						</>
+					) : (
+						<p className={styles["empty-placeholder"]}>
+							No users found matching your search criteria.
+						</p>
```
liza-veis, line 115: we already have empty placeholder for table, maybe instead we can pass this text to it?

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -62,33 +75,47 @@ const GroupUsersTable = ({
 	const error = errors["userIds"]?.message;
 	const hasError = Boolean(error);
 
-	if (
+	const hasUser = userData.length !== EMPTY_LENGTH;
```
liza-veis, line 78:
```suggestion
	const hasUserRows = userData.length !== EMPTY_LENGTH;
```

liza-veis (changes requested):

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
AnnKasian: @Fjortis We already have debounce in Search component

Fjortis: You're right, apologies for any confusion

Fjortis:
I believe the Search icon needs to be adjusted, it is not aligned like in figma 
image

apps/frontend/src/libs/components/input/styles.module.css
```diff
@@ -64,6 +64,7 @@
 
 .input-icon-left {
 	left: 10px;
+	display: flex;
```
GvoFor, line 67: If this really help, then I think you should add this to `.input-icon-right` as well

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -29,6 +30,7 @@ type Properties<T> =
 const Table = <T extends object>({
 	columns,
 	data,
+	noDataMessage = "There is nothing yet.",
```
liza-veis, line 33:
```suggestion
	emptyPlaceholder = "There is nothing yet.",
```

yevhenii-hladkov (approved):

liza-veis (approved):

PR #304, opened 2024-09-12
feat: Add users column to the groups table gf-301

GvoFor:
## Screenshots
image

### DIFF
--- a/apps/frontend/src/pages/access-management/libs/types/group-row.type.ts
+++ b/apps/frontend/src/pages/access-management/libs/types/group-row.type.ts
@@ -3,6 +3,7 @@ type GroupRow = {
 	id: number;
 	name: string;
 	permissions: string[];
+	users: number;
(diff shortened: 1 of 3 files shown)


apps/frontend/src/pages/access-management/libs/types/group-row.type.ts
```diff
@@ -3,6 +3,7 @@ type GroupRow = {
 	id: number;
 	name: string;
 	permissions: string[];
+	users: number;
```
liza-veis, line 6: let's name it userCount for internal fields to make it more clear from the code, but the title for the column will remain "Users"
```suggestion
	userCount: number;
```

liza-veis (changes requested):

what1s1ove (approved):

liza-veis (approved):

PR #307, opened 2024-09-13
feat: add an ability to regenerate project api key gf-288

AnutkaGn:
Add an ability to regenerate the project API key.
When the key is already present, the button should have a "Regenerate" label.
After the button is clicked, the API key is updated in the database.
image


apps/frontend/src/modules/project-api-keys/project-api-keys-api.ts
```diff
@@ -35,6 +37,22 @@ class ProjectApiKeysApi extends BaseHTTPApi {
 
 		return await response.json<ProjectApiKeyCreateResponseDto>();
 	}
+
+	public async patch(
```
liza-veis, line 41: We need to delete the previous key and create a new one, not to update it. We also will need to use existing post method for this, not patch.

liza-veis (changes requested):

packages/shared/src/libs/enums/exception-message.enum.ts
```diff
@@ -8,6 +8,8 @@ const ExceptionMessage = {
 	NO_PERMISSION: "You do not have permission to access this resource.",
 	NO_TOKEN_PROVIDED: "No token provided.",
 	PROJECT_API_KEY_ALREADY_EXISTS: "API key already exists for the project.",
+	PROJECT_API_KEY_NOT_FOUND: "API key not found.",
+	PROJECT_API_KEY_UPDATE_FAILED: "Failed to update project API key.",
```
Fjortis, line 12: I think the message doesn't align with the format of the other messages, perhaps something like this would confirm to the format better?


```suggestion
	PROJECT_API_KEY_UPDATE_FAILED: "Project API key update failed.",
```

apps/backend/src/modules/project-api-keys/project-api-key.controller.ts
```diff
@@ -69,6 +69,17 @@ class ProjectApiKeyController extends BaseController {
...
+		this.addRoute({
+			handler: (options) =>
+				this.delete(
+					options as APIHandlerOptions<{
+						params: { projectId: string };
+					}>,
+				),
+			method: "DELETE",
+			path: ProjectApiKeysApiPath.$ID,
+		});
```
liza-veis, line 82: We don't need a route for deletion either, this should happen on the service level in the create method because it's a part of a creation logic:
- if we already have API key for project - then remove it
- then create a new API key
From the frontend perspective, everything is the same, except that we no longer block the button.

liza-veis (changes requested):

apps/backend/src/modules/project-api-keys/project-api-key.controller.ts
AnutkaGn: This looks like a much simpler solution than the one I have been using

apps/backend/src/modules/project-api-keys/project-api-key.repository.ts
```diff
@@ -30,8 +30,13 @@ class ProjectApiKeyRepository implements Repository {
 		return ProjectApiKeyEntity.initialize(projectApiKey);
 	}
 
-	public delete(): ReturnType<Repository["delete"]> {
-		return Promise.resolve(false);
+	public async delete(projectId: number): Promise<boolean> {
+		const deletedRowsCount = await this.projectApiKeyModel
+			.query()
+			.delete()
+			.where("project_id", projectId);
```
GvoFor, line 37: Let's end the query with `.execute()`

apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
```diff
@@ -87,8 +85,7 @@ const SetupAnalyticsModal = ({
 					/>
 					<div className={styles["button-wrapper"]}>
 						<Button
-							isDisabled={isGenerateButtonDisabled}
```
GvoFor, line 90: I think we should make the button disabled, while request is proccessing, so there is no chance user send request multiple times (if this possible)

liza-veis (approved):

yevhenii-hladkov (approved):

PR #320, opened 2024-09-14
feat: improve table navigation add a loader to tables gf-311

AnutkaGn:
The page does not scroll up while navigating the Users and Groups tables.
A loader is displayed within the table to indicate loading.

### DIFF
--- a/apps/frontend/src/libs/components/table/table.tsx
+++ b/apps/frontend/src/libs/components/table/table.tsx
@@ -4,8 +4,10 @@ import {
 	useReactTable,
 } from "@tanstack/react-table";
 
+import { Loader } from "~/libs/components/components.js";
 import { EMPTY_LENGTH } from "~/libs/constants/constants.js";
 import { getValidClassNames } from "~/libs/helpers/helpers.js";
+import { useEffect, useRef, useState } from "~/libs/hooks/hooks.js";
 import { type TableColumn } from "~/libs/types/types.js";
 
 import { SelectRowCell } from "./libs/components/components.js";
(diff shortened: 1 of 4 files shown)


apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -29,12 +32,22 @@ type Properties<T> =
...
 		| SelectableProperties<T>;
 
+	const tbodyReference = useRef<HTMLTableSectionElement>(null);
+	const [tableBodyHeight, setTableBodyHeight] = useState<null | number>(null);
+
+	useEffect(() => {
+		if (!isLoading && tbodyReference.current) {
+			setTableBodyHeight(tbodyReference.current.clientHeight);
+		}
+	}, [isLoading]);
```
liza-veis, line 49: This logic looks too complicated for its gain. Instead, let's just set min-height for table-data instead of max to make it fixed. I think even if we have fewer rows (for example, we are on the last page) we still want the table to have the same height so that pagination doesn't jump

liza-veis (changes requested):

apps/frontend/src/libs/components/table/table.tsx
AnutkaGn: But I can't set a fixed height, then the table will just jump. Because if you select 5 elements in the pagination, it will be one height, if 10 is another. Because of this I dynamically get the height of the table so that the loader has the same height.
I still don't understand how the min-height will make the table not jump. What exactly should be the min-height?

liza-veis: I mean I see two possible ways of implementation:

1. Set a fixed min-size for the table during loading, for example, min-height 300px. And this width doesn't depend on the table page size.
2. Set a fixed min-height for the table overall instead of max-height. This means if we don't have enough rows on a page we will display empty space, but the height will remain the same. 

take a look at this example, on the last table page
https://mui.com/base-ui/react-table-pagination/#customized-look-and-feel

Issue #181, opened 2024-09-04
feat: Add an ability to save and retrieve activity


liza-veis:
### What feature?

We will have a script that automatically sends an array of statistic records. It will include git emails, date, and number of commits. We need to add an ability to save this statistic to the backend and retrieve it.

- [x] Add a contributors table (columns: id, created_at, updated_at, and name). 
- [x] Add git_emails table (columns: id, created_at, updated_at, contributor_id, email). 
- [x] Add activity_logs table (columns: id, created_at, updated_at, git_email_id, project_id, created_by_user_id, date, commits_number)
- [x] We will have a different type of authorization for the POST method. It will be in Bearer format, but instead of a usual token, we will get project_api_key #180. `created_by_user_id` and `project_id` should be taken from this key after encoding it and finding in the db. For GET request, we will have the usual authorization.
- [x] We will receive records grouped by date with author name, email, and commits_number as payload for POST request.
- [x] For each git_email, that doesn't exist yet, we need to create a contributor record first and then use it to create a record in the git_emails table. Git_email record cannot exist without contributor_id. A contributor's name can be taken from the author's email.
- [x] Add an endpoint to retrieve activity records. They should contain contributor data attached.

Example of payload (can be changed if needed or if another format is more convenient)
```json
[
  {
    "date": "2024-09-01",
    "items": [{
      "authorName": "John Doe",
      "authorEmail": "john.doe@gmail.com",
      "commitsNumber": 1,
    }],
  }
]
```

VeronikaHordus:
User Story: 
As a manager, I want the application to automatically save and retrieve contributors statistics from Git, so that I can analyze developer activity and project progress over time.

Acceptance Criteria:
1. The script runs in the background, collects repository statistics, and periodically sends them to the backend.

sandrvvu:
@liza-veis Should this feature be implemented as a separate activity module with its own controller, service, and models, right?

liza-veis:
> Should this feature be implemented as a separate activity module with its own controller, service, and models, right?

yes

VeronikaHordus:
All good!

PR #334, opened 2024-09-16
feat: add contributors page gf-250

Vitaliistf:
- Added contributors item to the sidebar.
- Added contributors table with all the contributors. It contains Name, Git Emails and Projects columns.
- Projects - are all projects associated with the contributor through git emails and activity logs

(Branched from #181) 
  
Screenshot 2024-09-16 at 16 58 43


apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -75,6 +75,22 @@ class ProjectRepository implements Repository {
 		return item ? ProjectEntity.initialize(item) : null;
 	}
 
+	public async findProjectsByContributorId(
```
liza-veis, line 78:
```suggestion
	public async findByContributorId(
```

apps/backend/src/modules/contributors/contributor.service.ts
```diff
@@ -45,19 +52,53 @@ class ContributorService implements Service {
...
-	public findAll(): ReturnType<Service["findAll"]> {
-		return Promise.resolve({ items: [] });
+	public async findAll(): Promise<ContributorGetAllResponseDto> {
+		const contributors = await this.contributorRepository.findAll();
+
+		const contributorsWithProjects = await Promise.all(
+			contributors.items.map(async (item) => {
+				const contributor = item.toObject();
+
+				const projects = await this.projectService.findProjectsByContributorId(
```
liza-veis, line 67: Maybe we can join projects inside contributors repository findAll method?

liza-veis (changes requested):

apps/backend/src/modules/contributors/contributor.service.ts
Vitaliistf: Oh, I thought that such logic should be through services. You mean that I should remove it completely from projectRepository, and create such method in contributorRepository, or make contributorRepository dependent of projectRepository?

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -15,35 +15,52 @@ class ContributorRepository implements Repository {
...
 	public delete(): ReturnType<Repository["delete"]> {
 		return Promise.resolve(true);
 	}
 
 	public async find(id: number): Promise<ContributorEntity | null> {
-		const item = await this.contributorModel.query().findById(id).execute();
+		const contributor = await this.contributorModel
+			.query()
+			.findById(id)
+			.withGraphFetched("gitEmails");
```
GvoFor, line 32: I think you can modify `relationMappings` method of `ContributorModel` and use additionally `.withGraphFetched("projects")` here and in other places, so you don't need to "inject" projects using `ProjectService`'s `getProjectsByContributorId`

apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -75,6 +75,22 @@ class ProjectRepository implements Repository {
...
+		contributorId: number,
+	): Promise<{ items: ProjectEntity[] }> {
+		const projects = await this.projectModel
+			.query()
+			.select("projects.*")
+			.join("activity_logs", "activity_logs.project_id", "projects.id")
+			.join("git_emails", "git_emails.id", "activity_logs.git_email_id")
+			.where("git_emails.contributor_id", contributorId)
+			.distinct();
```
GvoFor, line 87: so you won't need this scary query ;)

apps/backend/src/modules/contributors/contributor.service.ts
liza-veis: No, I was thinking about calculating it as a custom field somehow inside the query in contributors repository, but looks like it will be complicated, so let's leave it as it is

liza-veis: Then I think we don't need the projects field inside contributors entity, looks like we don't use it

apps/backend/src/modules/projects/project.repository.ts
GvoFor: It'll be handled by objection's `.withGraphFetched()` method. At least I think so

apps/backend/src/modules/contributors/contributor.service.ts
liza-veis: so the problem is that it's not really good to make a separate query for each item. Also, we need to use for of, instead of Promise all, because this may make too many simultaneous db requests, which is bad. And the concern here is that with for of it will work slow

apps/backend/src/modules/contributors/contributor.repository.ts
Vitaliistf: I tried to do it, also discussed this with Liza, and seems like the relation is too complicated, and we cannot describe it in the model.

apps/backend/src/modules/contributors/contributor.service.ts
liza-veis: and if there is a possibility to compute or somehow join this inside the contributors query, this would be better, otherwise we need to use for of here, which can be slow

apps/backend/src/modules/contributors/contributor.repository.ts
GvoFor: Oh, got it

apps/backend/src/modules/contributors/contributor.service.ts
Vitaliistf: Oh, alright, I'll try to figure out what we could do with it tomorrow. Resolving a lot of conflicts today got me feeling fried today😂

what1s1ove (approved):
lgtm after Lisa's comments

apps/backend/src/modules/contributors/contributor.service.ts
```diff
@@ -2,19 +2,26 @@ import { ExceptionMessage } from "~/libs/enums/enums.js";
...
 	private contributorRepository: ContributorRepository;
+	private projectService: ProjectService;
 
-	public constructor(contributorRepository: ContributorRepository) {
+	public constructor(
+		contributorRepository: ContributorRepository,
+		projectService: ProjectService,
+	) {
 		this.contributorRepository = contributorRepository;
+		this.projectService = projectService;
```
liza-veis, line 24: Looks like we no longer need this
```suggestion
```

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -15,35 +17,68 @@ class ContributorRepository implements Repository {
...
+			.select(
+				raw(
+					"COALESCE(ARRAY_AGG(DISTINCT jsonb_build_object('id', projects.id, 'name', projects.name)) FILTER (WHERE projects.id IS NOT NULL), '{}') AS projects",
+				),
+			)
+			.select(
+				raw(
+					"COALESCE(ARRAY_AGG(DISTINCT jsonb_build_object('id', git_emails.id, 'email', git_emails.email)) FILTER (WHERE git_emails.id IS NOT NULL), '{}') AS git_emails",
+				),
+			)
```
liza-veis, line 56: I think we can join git emails the usual way, with relation mapping, because we will probably need this in another place

liza-veis (changes requested):

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -15,35 +17,68 @@ class ContributorRepository implements Repository {
...
 		const contributor = await this.contributorModel
 			.query()
-			.insert({
-				name,
-			})
+			.insert({ name })
 			.execute();
 
-		return ContributorEntity.initialize(contributor);
+		return ContributorEntity.initialize({ ...contributor, projects: [] });
```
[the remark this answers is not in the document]
Vitaliistf: Yeah, now that we have one query for that, we can, I forgot to change it. Thanks!

liza-veis (approved):

PR #354, opened 2024-09-17
feat: add project last activity label to project list gf-274

kolibri753:
- added a Migration: add `last_activity_date` to the projects table, defaulting to null;
- updated Projects module: add new property lastActivityDate and update method updateLastActivityDate;
- updated Activity Logs module: update last_activity_date when creating activity logs;
- adjust frontend: add 'last activity' label with appropriate color indicators based on last_activity_date.

### DIFF
--- a/apps/backend/src/modules/activity-logs/activity-log.service.ts
+++ b/apps/backend/src/modules/activity-logs/activity-log.service.ts
@@ -2,6 +2,7 @@ import { type Service } from "~/libs/types/types.js";
 import { type ContributorService } from "~/modules/contributors/contributors.js";
 import { type GitEmailService } from "~/modules/git-emails/git-emails.js";
 import { type ProjectApiKeyService } from "~/modules/project-api-keys/project-api-keys.js";
+import { type ProjectService } from "~/modules/projects/project.service.js";
 
 import { ActivityLogEntity } from "./activity-log.entity.js";
 import { type ActivityLogRepository } from "./activity-log.repository.js";
@@ -16,6 +17,7 @@ type Constructor = {
 	contributorService: ContributorService;
 	gitEmailService: GitEmailService;
 	projectApiKeyService: ProjectApiKeyService;
+	projectService: ProjectService;
 };
 
 class ActivityLogService implements Service {
@@ -23,17 +25,20 @@ class ActivityLogService implements Service {
 	private contributorService: ContributorService;
 	private gitEmailService: GitEmailService;
 	private projectApiKeyService: ProjectApiKeyService;
+	private projectService: ProjectService;
 
 	public constructor({
 		activityLogRepository,
 		contributorService,
 		gitEmailService,
 		projectApiKeyService,
+		projectService,
 	}: Constructor) {
 		this.activityLogRepository = activityLogRepository;
 		this.contributorService = contributorService;
 		this.gitEmailService = gitEmailService;
 		this.projectApiKeyService = projectApiKeyService;
+		this.projectService = projectService;
 	}
 
 	private async createActivityLog({
@@ -56,6 +61,8 @@ class ActivityLogService implements Service {
 				email: authorEmail,
 			});
 		}
+
+		await this.projectService.updateLastActivityDate(projectId, date);
 
 		return await this.activityLogRepository.create(
 			ActivityLogEntity.initializeNew({
--- a/apps/backend/src/modules/activity-logs/activity-logs.ts
+++ b/apps/backend/src/modules/activity-logs/activity-logs.ts
@@ -2,11 +2,20 @@ import { logger } from "~/libs/modules/logger/logger.js";
 import { contributorService } from "~/modules/contributors/contributors.js";
 import { gitEmailService } from "~/modules/git-emails/git-emails.js";
 import { projectApiKeyService } from "~/modules/project-api-keys/project-api-keys.js";
+import { ProjectModel } from "~/modules/projects/project.model.js";
+import { ProjectRepository } from "~/modules/projects/project.repository.js";
+import { ProjectService } from "~/modules/projects/project.service.js";
 
 import { ActivityLogController } from "./activity-log.controller.js";
 import { ActivityLogModel } from "./activity-log.model.js";
 import { ActivityLogRepository } from "./activity-log.repository.js";
 import { ActivityLogService } from "./activity-log.service.js";
+
+const projectRepository = new ProjectRepository(ProjectModel);
+const projectService = new ProjectService(
+	projectRepository,
+	projectApiKeyService,
+);
 
 const activityLogRepository = new ActivityLogRepository(ActivityLogModel);
 const activityLogService = new ActivityLogService({
@@ -14,6 +23,7 @@ const activityLogService = new ActivityLogService({
 	contributorService,
 	gitEmailService,
 	projectApiKeyService,
+	projectService,
 });
 const activityLogController = new ActivityLogController(
 	logger,
--- a/apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
+++ b/apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
@@ -1,10 +1,11 @@
 import { NavLink } from "react-router-dom";
 
 import { AppRoute } from "~/libs/enums/enums.js";
-import { configureString } from "~/libs/helpers/helpers.js";
+import { configureString, getValidClassNames } from "~/libs/helpers/helpers.js";
 import { useCallback } from "~/libs/hooks/hooks.js";
 import { type ProjectGetAllItemResponseDto } from "~/modules/projects/projects.js";
 
+import { getRelativeDateLabel } from "../../helpers/helpers.js";
 import { ProjectMenu } from "../components.js";
 import styles from "./styles.module.css";
 
@@ -31,11 +32,25 @@ const ProjectCard = ({
 		onDelete(project);
 	}, [onDelete, project]);
 
+	const lastUpdate = getRelativeDateLabel(project.lastActivityDate);
+
 	return (
 		<div className={styles["project-container"]}>
 			<NavLink className={styles["project"] as string} to={projectRoute}>
 				{project.name}
 			</NavLink>
+			{lastUpdate && (
+				<div className={styles["last-activity-wrapper"]}>
+					<span
+						className={getValidClassNames(
+							styles["last-activity"],
+							styles[lastUpdate.colorClass],
+						)}
+					>
+						{lastUpdate.label}
+					</span>
+				</div>
+			)}
 			<ProjectMenu onDelete={handleDeleteClick} onEdit={handleEditClick} />
 		</div>
 	);
--- a/apps/frontend/src/pages/projects/libs/components/project-card/styles.module.css
+++ b/apps/frontend/src/pages/projects/libs/components/project-card/styles.module.css
@@ -25,3 +25,35 @@
 	text-decoration: none;
 	background: transparent;
 }
+
+.last-activity-wrapper {
+	width: 100%;
+	padding: 0 33%;
+}
+
+.last-activity {
+	display: flex;
+	align-items: center;
+	justify-content: flex-start;
+}
+
+.last-activity::before {
+	display: inline-block;
+	width: 12px;
+	height: 12px;
+	margin-right: 8px;
+	content: "";
+	border-radius: 50%;
+}
+
+.green::before {
+	background-color: var(--color-success);
+}
+
+.yellow::before {
+	background-color: var(--color-warning);
+}
+
+.red::before {
+	background-color: var(--color-danger);
+}
--- a/apps/frontend/src/pages/projects/libs/enums/color-class.enum.ts
+++ b/apps/frontend/src/pages/projects/libs/enums/color-class.enum.ts
@@ -1,0 +1,7 @@
+const ColorClass = {
+	GREEN: "green",
+	RED: "red",
+	YELLOW: "yellow",
+} as const;
+
+export { ColorClass };
--- a/apps/frontend/src/pages/projects/libs/enums/color-threshold.enum.ts
+++ b/apps/frontend/src/pages/projects/libs/enums/color-threshold.enum.ts
@@ -1,0 +1,6 @@
+const ColorThreshold = {
+	GREEN: 2,
+	YELLOW: 5,
+} as const;
+
+export { ColorThreshold };
--- a/apps/frontend/src/pages/projects/libs/helpers/get-relative-date-label.helper.tsx
+++ b/apps/frontend/src/pages/projects/libs/helpers/get-relative-date-label.helper.tsx
@@ -1,0 +1,38 @@
+import { getDifferenceInDays } from "~/libs/helpers/helpers.js";
+
+import { ColorClass, ColorThreshold } from "../enums/enums.js";
+
+type RelativeDateLabel = {
+	colorClass: string;
+	label: string;
+} | null;
+
+const getRelativeDateLabel = (
+	lastActivityDate: null | string,
+): RelativeDateLabel => {
+	if (!lastActivityDate) {
+		return null;
+	}
+
+	const currentDate = new Date();
+	const lastActivity = new Date(lastActivityDate);
+	const diffDays = getDifferenceInDays(currentDate, lastActivity);
+
+	if (diffDays < ColorThreshold.GREEN) {
+		return { colorClass: ColorClass.GREEN, label: "Updated today" };
+	}
+
+	if (diffDays < ColorThreshold.YELLOW) {
+		return {
+			colorClass: ColorClass.YELLOW,
+			label: `Updated ${diffDays.toString()} days ago`,
+		};
+	}
+
+	return {
+		colorClass: ColorClass.RED,
+		label: `Updated ${diffDays.toString()} days ago`,
+	};
+};
+
+export { getRelativeDateLabel };
--- a/apps/backend/src/db/migrations/20240916154357_add_last_activity_date_to_projects_table.ts
+++ b/apps/backend/src/db/migrations/20240916154357_add_last_activity_date_to_projects_table.ts
@@ -1,0 +1,21 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "projects";
+
+const ColumnName = {
+	LAST_ACTIVITY_DATE: "last_activity_date",
+} as const;
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.dateTime(ColumnName.LAST_ACTIVITY_DATE).nullable();
+	});
+}
+
+function down(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table.dropColumn(ColumnName.LAST_ACTIVITY_DATE);
+	});
+}
+
+export { down, up };
--- a/apps/backend/src/modules/projects/project.entity.ts
+++ b/apps/backend/src/modules/projects/project.entity.ts
@@ -7,34 +7,42 @@ class ProjectEntity implements Entity {
 
 	private id: null | number;
 
+	private lastActivityDate!: null | string;
+
 	private name!: string;
 
 	private constructor({
 		description,
 		id,
+		lastActivityDate,
 		name,
 	}: {
 		description: string;
 		id: null | number;
+		lastActivityDate: null | string;
 		name: string;
 	}) {
 		this.id = id;
 		this.description = description;
+		this.lastActivityDate = lastActivityDate;
 		this.name = name;
 	}
 
 	public static initialize({
 		description,
 		id,
+		lastActivityDate,
 		name,
 	}: {
 		description: string;
 		id: number;
+		lastActivityDate: string;
 		name: string;
 	}): ProjectEntity {
 		return new ProjectEntity({
 			description,
 			id,
+			lastActivityDate,
 			name,
 		});
 	}
@@ -49,6 +57,7 @@ class ProjectEntity implements Entity {
 		return new ProjectEntity({
 			description,
 			id: null,
+			lastActivityDate: null,
 			name,
 		});
 	}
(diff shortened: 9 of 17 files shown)


apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
```diff
@@ -31,11 +32,25 @@ const ProjectCard = ({
...
+					<span
+						className={getValidClassNames(
+							styles["last-activity"],
+							styles[lastUpdate.colorClass],
+						)}
+					>
+						{lastUpdate.label}
+					</span>
+				</div>
+			)}
```
kolibri753, line 53: Maybe I should put this label inside the NavLink (not sure, since NavLink is positioned absolutely), and find another way (besides using `0 33% padding`) to achieve this layout

apps/frontend/src/pages/projects/libs/enums/color-class.enum.ts
```diff
@@ -0,0 +1,7 @@
+const ColorClass = {
+	GREEN: "green",
+	RED: "red",
+	YELLOW: "yellow",
+} as const;
```
kolibri753, line 5: not sure about naming

apps/backend/src/modules/activity-logs/activity-logs.ts
```diff
@@ -2,18 +2,28 @@ import { logger } from "~/libs/modules/logger/logger.js";
...
 import { ActivityLogModel } from "./activity-log.model.js";
 import { ActivityLogRepository } from "./activity-log.repository.js";
 import { ActivityLogService } from "./activity-log.service.js";
 
+const projectRepository = new ProjectRepository(ProjectModel);
+const projectService = new ProjectService(
+	projectRepository,
+	projectApiKeyService,
+);
+
```
GvoFor, line 19: Let's import `projectService` as other services instead of creating new one. The fact it (or repo only) was created inside `project-api-keys.ts` is that `project-api-keys.ts` was imported inside `projects.ts` in order to inject `projectApiKeyService` into `ProjectService`, so there was an issue with cyclic dependencies.

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
GvoFor: Hm. I think you should redesign markup and styles for this component, so every inner elements (name, activity label, and menu button) are layouted using flex-box.
And in this case instead of `NavLink` we can add `onClick` prop to `project-container` div and use `useNavigate` hook for navigating user to the project.
But this is just my point of view. So let our coaches dicide

apps/frontend/src/pages/projects/libs/enums/color-class.enum.ts
GvoFor: What about `LastActivityIndicatorColor`?

apps/frontend/src/pages/projects/libs/enums/color-threshold.enum.ts
GvoFor, line 1: Do not forget about this enum, after you pick the name for previous ;)

apps/frontend/src/pages/projects/libs/helpers/get-relative-date-label.helper.tsx
```diff
@@ -0,0 +1,38 @@
+import { getDifferenceInDays } from "~/libs/helpers/helpers.js";
+
+import { ColorClass, ColorThreshold } from "../enums/enums.js";
+
+type RelativeDateLabel = {
+	colorClass: string;
```
GvoFor, line 6:
```suggestion
	colorClass: ColorClass;
```

apps/frontend/src/pages/projects/libs/enums/color-class.enum.ts
Vitaliistf: Maybe it'd be better to name this enum as just Color, or ColorName?

Vitaliistf: Oh, I see, it's for class names, now I'm not sure too, but think that ColorClassName would be pretty self-explanatory

apps/backend/src/modules/activity-logs/activity-logs.ts
kolibri753: Oh, thanks for noticing

apps/frontend/src/pages/projects/libs/helpers/get-relative-date-label.helper.tsx
```diff
@@ -0,0 +1,38 @@
...
+	label: string;
+} | null;
+
+const getRelativeDateLabel = (
```
liza-veis, line 10: We need to use date fns for calculating relative date. This is the example from the main page:

```
formatDistance(subDays(new Date(), 3), new Date(), { addSuffix: true })
//=> "3 days ago"
```

Also we need to calculate date and color separately.

apps/frontend/src/pages/projects/libs/helpers/get-relative-date-label.helper.tsx
kolibri753: Hm.. not sure here `colorClass: typeof ColorClass[keyof typeof ColorClass];`. Are we trying to enforce the colorClass to only accept values from the `ColorClass` object?

liza-veis (changes requested):

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
```diff
@@ -31,11 +32,25 @@ const ProjectCard = ({
...
 
+	const lastUpdate = getRelativeDateLabel(project.lastActivityDate);
+
 	return (
 		<div className={styles["project-container"]}>
 			<NavLink className={styles["project"] as string} to={projectRoute}>
 				{project.name}
 			</NavLink>
+			{lastUpdate && (
+				<div className={styles["last-activity-wrapper"]}>
```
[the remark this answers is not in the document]
kolibri753: good point, thanks :)

kolibri753:
Should the result look like this (image below) or like the description ([image above](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/354#issue-2530203296)) ?

image

apps/frontend/src/pages/projects/libs/helpers/get-color.helper.tsx
```diff
@@ -0,0 +1,26 @@
+import { getDifferenceInDays } from "~/libs/helpers/helpers.js";
+
+import { ColorThreshold } from "../enums/enums.js";
+
+const getColor = (
+	currentDate: Date,
+	lastActivityDate: Date | null,
+): "green" | "red" | "yellow" | null => {
```
liza-veis, line 8:
```suggestion
): ComponentProps<typeof ActivityIndicator>["status"] => {
```

apps/frontend/src/pages/projects/libs/helpers/get-color.helper.tsx
```diff
@@ -0,0 +1,26 @@
...
+	if (!lastActivityDate) {
+		return null;
+	}
+
+	const diffDays = getDifferenceInDays(currentDate, lastActivityDate);
+
+	if (diffDays < ColorThreshold.GREEN) {
```
liza-veis, line 15:
```suggestion
	if (diffDays < ActivityIndicatorThreshold.GREEN) {
```

liza-veis (changes requested):

liza-veis:
> Should the result look like this (image below) or like the description (https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/354#issue-2530203296) ?

Similar to description, should be like "Updated today" or "Last activity today"

what1s1ove (approved):
lgtm after Lisa's comments

apps/frontend/src/pages/projects/libs/components/activity-indicator/activity-indicator.tsx
Vitaliistf, line 1: Not sure if you need to change this, but this indicator will be reused across different components, so it might be better to move it to common components folder.

apps/frontend/src/pages/projects/libs/components/project-card/styles.module.css
```diff
@@ -25,3 +25,7 @@
 	text-decoration: none;
 	background: transparent;
 }
+
+.project-name {
+	flex: 0 0 33%;
```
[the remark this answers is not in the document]
kolibri753: hmm.. it is a way to make layout
image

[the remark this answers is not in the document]
liza-veis: It looks like when we resize the screen, we expect the space before options (after the activity label) to be reduced first. So I think we need to set either max-width: max(33%, 300px) or just max-width: 300px as on figma layout

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
```diff
@@ -0,0 +1,15 @@
+import { differenceInDays, formatDistance } from "date-fns";
+
+const getRelativeDate = (date: Date, baseDate: Date): string => {
+	const LESS_THAN_DAY = 0;
+
+	const daysDifference = differenceInDays(baseDate, date);
+
+	if (daysDifference === LESS_THAN_DAY) {
+		return "Updated today";
+	}
```
liza-veis, line 10: 1. we should always have Updated keyword, like `Updated 2 days ago`
2. this helper should only return relative date - today,  2 days ago etc. and we need to add `Updated ` where its used.
3. I think formatDistance should handle `today` somehow too, so that we don't need to calculate it manually
```suggestion
```

liza-veis (changes requested):

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
liza-veis: or maybe there are other methods in date-fns, like formatRelative that will fit more

apps/frontend/src/libs/enums/activity-indicator-threshold.enum.ts
```diff
@@ -0,0 +1,6 @@
+const ActivityIndicatorThreshold = {
```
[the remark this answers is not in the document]
kolibri753: Oh, I thought about it, okay, thanks :)

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
kolibri753: already tried to find any, formatDistance seems the most appropriate, but not sure how to make it only date-based (without `15 hours ago`)

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
```diff
@@ -31,11 +37,31 @@ const ProjectCard = ({
...
+	const daysDifference = lastActivityDate
+		? getDifferenceInDays(currentDate, lastActivityDate)
+		: null;
+
+	const lastUpdateLabel = lastActivityDate
+		? getRelativeDate(lastActivityDate, currentDate)
+		: null;
+
+	const colorStatus =
+		daysDifference === null ? null : getActivityIndicatorStatus(daysDifference);
```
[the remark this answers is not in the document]
liza-veis: We can check if it's number then, to prevent this

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
liza-veis: We should be able to do this with formatRelative, checkout this example https://stackoverflow.com/a/49139748. In case this doesn't work, the second option is to remove time from today and use only date for comparison

apps/frontend/src/pages/projects/libs/components/project-card/styles.module.css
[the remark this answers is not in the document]
kolibri753: not sure how `max-width` can help in this case :thinking: 
maybe better leave `flex-basis: 300px` (if not percentage), so space between activity label and options will be reduced first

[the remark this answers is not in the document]
kolibri753: i just believe it will be better in percentage, to not cause problems in future (1024px layout)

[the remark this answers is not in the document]
liza-veis: flex-basis will be ok

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
kolibri753: image
Sorry need to clarify: is this option alright? Or are we trying to get `2 days ago`, `a month ago`, `a year ago`?

liza-veis: looks ok to me

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
kolibri753, line 1: I'll look into it more in the morning. :cat2:

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
kolibri753: thank a lot for helping :)

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const locale = {
+	...enUS,
+	formatRelative: (token: string): string =>
+		formatRelativeLocale[token as keyof typeof formatRelativeLocale],
+};
+
+const getRelativeDate = (date: Date, baseDate: Date): string => {
+	return formatRelative(date, baseDate, { locale });
+};
```
kolibri753, line 19: Haven't found anything better for now (I prefer to change it later if requested). Should I also move `formatRelativeLocale` and `locale` to different files in this folder (create libs/enums/format-relative-locale.map.ts and relative-locale.helper.ts), or even outside this folder (not sure it is reusable, so I prefer to leave it as it is).

kolibri753: But if we prefer dates, I can add a little complexity like this `"'Updated' " + daysDifference.toString() + " 'days ago"`

```
const formatRelativeLocale = {
	lastWeek: "'Updated last' eeee",
	other: (date: Date, baseDate: Date): string => {
		const daysDifference = differenceInDays(baseDate, date);

		return "'Updated' " + daysDifference.toString() + " 'days ago";
	},
	today: "'Updated today'",
	yesterday: "'Updated yesterday'",
};

const locale = {
	...enUS,
	formatRelative: (token: string, date: Date, baseDate: Date): string => {
		const formatter =
			formatRelativeLocale[token as keyof typeof formatRelativeLocale];

		return typeof formatter === "function"
			? formatter(date, baseDate)
			: formatter;
	},
};
```

It will show `X days ago`, instead of date

kolibri753: Or make it simpler, but with hours

```
import { formatDistance } from "date-fns";

const getRelativeDate = (
	date: Date | number | string,
	baseDate: Date | number | string,
): string => {
	return "Updated " + formatDistance(date, baseDate);
};

export { getRelativeDate };
```
image

kolibri753: Or a little more complicated, but with today and yersterday

```
import { differenceInDays, formatDistance, isSameDay } from "date-fns";

const getRelativeDate = (
	date: Date | number | string,
	baseDate: Date | number | string,
): string => {
	const parsedDate = new Date(date);
	const parsedBaseDate = new Date(baseDate);
	const DIFF_IN_A_DAY = 1;

	if (isSameDay(parsedDate, parsedBaseDate)) {
		return "Updated today";
	}

	if (differenceInDays(parsedBaseDate, parsedDate) === DIFF_IN_A_DAY) {
		return "Updated yesterday";
	}

	return "Updated " + formatDistance(parsedDate, parsedBaseDate) + " ago";
};

export { getRelativeDate };
```
image

liza-veis: Also as we are checking for last activity (not last statistics update), we also need to check that there is at least one commit for this project for the date before updating last_activity_date

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
```diff
@@ -31,11 +37,33 @@ const ProjectCard = ({
...
+		: null;
+
+	const lastUpdateLabel = lastActivityDate
+		? getRelativeDate(lastActivityDate, currentDate)
+		: null;
+
+	const colorStatus =
+		typeof daysDifference === "number"
+			? getActivityIndicatorStatus(daysDifference)
+			: null;
```
perekotypole, line 56: Can't we move all the calculations to a some sort of helper and just return the object of required variables?

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
kolibri753: if I understand correctly, even with reset (`startOfDay(new Date())`) we still will get them in answer (but i'll check it once  battery charge)

or you mean to reset, but also add this
```
if (isSameDay(parsedDate, parsedBaseDate)) {
	return "Updated today";
}

if (differenceInDays(parsedBaseDate, parsedDate) === DIFF_IN_A_DAY) {
	return "Updated yesterday";
}
```

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -104,6 +111,10 @@ class ActivityLogService implements Service {
 				});
 
 				createdActivityLogs.items.push(activityLog.toObject());
+
+				if (hasCommits) {
+					await this.projectService.updateLastActivityDate(projectId, date);
+				}
```
kolibri753, line 117: not sure if this is a needed check here... we can't add an activity log without commits, or can we?

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
kolibri753: @liza-veis should I create `calculateActivityData` helper in projects module?

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
```diff
@@ -31,11 +37,33 @@ const ProjectCard = ({
 		onDelete(project);
 	}, [onDelete, project]);
 
+	const currentDate = new Date();
+	const lastActivityDate = project.lastActivityDate
+		? new Date(project.lastActivityDate)
+		: null;
```
kolibri753, line 43: should I create `startOfDay` helper (just wrapper around existing startOfDay method in `date-fns`) for setting date `Wed Sep 18 2024 00:00:00`

packages/shared/src/libs/helpers/get-relative-date/get-relative-date.helper.ts
liza-veis: Let's leave the current approach, looks ok to me

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
liza-veis: let's leave this, this doesn't look like a logic that we should group as for me. All of this can be re-used separately later by the component

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
liza-veis: Do we need it with the current approach? I don't see where it's currently used

kolibri753: just checking, maybe we still don't want hours/minutes/seconds be passed and prefer  `... 00:00:00`

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
```diff
@@ -31,11 +37,33 @@ const ProjectCard = ({
...
+	const lastUpdateLabel = lastActivityDate
+		? `Updated ${getRelativeDate(lastActivityDate, currentDate)}`
+		: null;
+
+	const colorStatus =
+		typeof daysDifference === "number"
+			? getActivityIndicatorStatus(daysDifference)
+			: null;
+
+	const hasActivityIndicator = lastUpdateLabel && colorStatus;
```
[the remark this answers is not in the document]
kolibri753: hmm.. now we need to provide a fallback value when `lastUpdateLabel && colorStatus` are null. Is it okay?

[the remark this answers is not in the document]
GvoFor: Or maybe we can write this like this:
```suggestion
	const hasActivityIndicator = 
		lastUpdateLabel !== null && 
		colorStatus !== null;
```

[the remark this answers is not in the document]
liza-veis: `Boolean(lastUpdateLabel && colorStatus)` cannot return true for the null value. We need to convert it to boolean for sure and we don't want to write this condition inside the component render, because this was the whole point in moving it out. Either use the approach as @GvoFor suggested above, or cast the types, because typescript doesn't properly handle Boolean call, but we don't want to write additional checks because of typescript

apps/frontend/src/pages/projects/libs/components/project-card/project-card.tsx
liza-veis: If it works as it is, then we don't

liza-veis (approved):

PR #364, opened 2024-09-17
feat: add analytics table gf-327

sandrvvu:
Add analytics table to the analytics page.
The user should be able to select a date range that is less than 1 month and the max date is today.
When user changes the date range, the columns in the table should match a new range and data should be updated.
image


apps/frontend/src/pages/analytics/libs/types/analytics-row.type.ts
```diff
@@ -0,0 +1,7 @@
+type AnalyticsRow = {
+	[key: string]: number | string;
```
what1s1ove, line 2: We need to specify all keys explicitly. do not use the index signature
```suggestion
```

apps/frontend/src/pages/analytics/libs/helpers/get-date-range/get-date-range.helper.ts
```diff
@@ -0,0 +1,22 @@
+import { formatDate } from "~/libs/helpers/helpers.js";
+
+const getDateRange = (startDay: Date, endDay: Date): string[] => {
+	const dates: Date[] = [];
+
+	const currentDate = new Date(startDay);
+
+	const DAYS_TO_ADVANCE = 1;
```
what1s1ove, line 8: b28

apps/frontend/src/libs/components/date-input/date-input.tsx
```diff
@@ -26,13 +26,15 @@ type Properties<T extends FieldValues> = {
 	maxDate?: Date;
 	maxRange?: number;
 	name: FieldPath<T>;
+	onPopoverClose?: (selectedDateRange: [Date, Date]) => void;
```
what1s1ove, line 29: i am not sure that we can do it. we should use only the `use-form` API to handle form values

apps/frontend/src/libs/components/table/libs/helper/helper.ts
s1rserg, line 1: Folder should be named helpers, this file should be named helpers.ts and helpers should have .helper in the name. Also not sure if each helper should have its folder. For instance, in `apps\frontend\src\pages\access-management` and `apps\frontend\src\libs\hooks\use-pagination` they are all in one helpers folder

apps/frontend/src/libs/components/date-input/date-input.tsx
sandrvvu: Oh, the reason I added that was because the form only submits when the button _(image)_ is clicked again. This isn't very clear from a user experience perspective, as the values change but the data isn't updated until the button is clicked. So I wanted to capture the state when the popover is closed by clicking outside the area.

Should I remove it?
image

what1s1ove: @liza-veis take a look please

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -1,28 +1,72 @@
...
 
+	const handleLoadActivityLogs = useCallback(() => {
+		void dispatch(activityLogActions.loadAll());
+	}, [dispatch]);
+
+	useEffect(() => {
+		handleLoadActivityLogs();
+	}, [handleLoadActivityLogs]);
+
+	const handlePopoverClose = useCallback(
```
Vitaliistf, line 48: Agree with Vlad, we use useForm for submission, no need to create additional functions, data will be already present in values of form. There will also be other inputs, so use of this is more convenient.

apps/frontend/src/libs/components/table/table.tsx
```diff
@@ -129,7 +130,10 @@ const Table = <T extends object>({
 								)}
 								{row.getVisibleCells().map((cell) => (
 									<td
-										className={styles["table-data"]}
+										className={getValidClassNames(
+											styles["table-data"],
+											styles[getCellColor(cell.getValue())],
```
Vitaliistf, line 135: I think that we might as well omit creation of this helper. `GetValidClassNames` provides us conditional classNames adding. See how it's done in other places, for example our `Select` component.

apps/frontend/src/libs/components/table/libs/helper/helper.ts
sandrvvu: thanks for noticing. idk [here](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/253#discussion_r1753242793) I was asked to have a separate folder for helper, so I did it here as well

apps/frontend/src/pages/analytics/libs/components/analytics-table/analytics-table.tsx
```diff
@@ -0,0 +1,37 @@
...
+	getDateRange,
+} from "../../helpers/helpers.js";
+import { type AnalyticsRow } from "../../types/types.js";
+import styles from "./styles.module.css";
+
+type Properties = {
+	activityLogs: ActivityLogGetAllItemResponseDto[];
+	dateRange: [Date, Date];
+};
+
```
Vitaliistf, line 16: I might be wrong, but I think it might be convenient to put the table on the page directly, as we won't have any other tables there.

apps/frontend/src/pages/analytics/libs/helpers/get-date-range/get-date-range.helper.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const DAYS_TO_ADVANCE = 1;
+
+const getDateRange = (startDay: Date, endDay: Date): string[] => {
+	const dates: Date[] = [];
+	const currentDate = new Date(startDay);
+
+	while (currentDate <= endDay) {
+		dates.push(new Date(currentDate));
+		currentDate.setDate(currentDate.getDate() + DAYS_TO_ADVANCE);
```
Vitaliistf, line 11: Not critical, but we have a library for dealing with dates (date-fns), I think you might get good use of that here.

liza-veis: Agree

apps/frontend/src/libs/components/date-input/date-input.tsx
liza-veis: I think instead we need to wrap the filters block on the analytics page with a form, and whenever its value changes, submit this form. We can track it by isDirty property, or something else

liza-veis (changes requested):

apps/frontend/src/libs/components/table/table.tsx
liza-veis: We shouldn't have this logic in table component at all, if we need to change color, we need to wrap the item with the span and apply the color to it and pass as a cell to columns config.

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -1,28 +1,73 @@
...
+	const { control, handleSubmit, isDirty, watch } = useAppForm({
 		defaultValues: {
-			dateRange: [subtractDays(todayDate, ANALYTICS_DATE_MAX_RANGE), todayDate],
+			dateRange: initialDateRange,
 		},
 	});
 
+	const handleLoadActivityLogs = useCallback(() => {
+		void dispatch(activityLogActions.loadAll());
+	}, [dispatch]);
```
Vitaliistf, line 42: I think we can call `dispatch` in `useEffect` directly, can't we? We won't reuse this function anywhere else and it is just an extra code.

apps/frontend/src/pages/analytics/libs/helpers/get-analytics-rows/get-analytics-rows.helper.ts
```diff
@@ -0,0 +1,43 @@
+import { formatDate } from "~/libs/helpers/helpers.js";
+import { type ActivityLogGetAllItemResponseDto } from "~/modules/activity-logs/activity-logs.js";
+
+import { type AnalyticsRow } from "../../types/types.js";
+
+const getAnalyticsRows = (
```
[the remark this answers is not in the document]
sandrvvu: Hey, I’ve tried using Objection.js, but it seems limited for complex grouping and aggregation. The only feasible approach I discovered so far involves using more raw Knex methods for grouping and summing data. But I am not sure about it.
Should I use Knex directly? Or alternatively, maybe to fetch the data and then transform it with a helper function?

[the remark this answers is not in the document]
GvoFor: In contributors repository knex was used directly (there also was an issue with objection limitation), so I think you can use it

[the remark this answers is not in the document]
liza-veis: > The only feasible approach I discovered so far involves using more raw Knex methods for grouping and summing data. But I am not sure about it.

It depends on how complicated it will be. The approach with raw queries is preferable here because it will take fewer resources, than grouping on the backend side, also in case we add pagination or additional filters, this may be hard to handle with js grouping. If this will take a lot of time, then let's just move on with the js grouping approach, but do this on the backend side. But if you can implement this with raw queries, this will be better.

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -1,28 +1,69 @@
...
 		[handleSubmit],
 	);
 
+	const dateRangeValue = watch("dateRange");
+
+	useEffect(() => {
+		if (isDirty) {
+			handleFormSubmit();
+		}
+	}, [dateRangeValue, isDirty, handleFormSubmit]);
```
perekotypole, line 59: I wonder why `isDirty` isn't enough and you use `watch` additionally?

apps/frontend/src/pages/analytics/libs/helpers/get-analytics-columns/get-analytics-columns.helper.tsx
```diff
@@ -0,0 +1,35 @@
...
+		{
+			accessorKey: "contributorName",
+			header: "Name",
+			size: 220,
+		},
+	];
+
+	const dateColumns: TableColumn<AnalyticsRow>[] = dateRange.map((date) => ({
+		accessorKey: `commitsNumber_${date}.commitsNumber`,
+		cell: ({ getValue }): JSX.Element => {
```
perekotypole, line 18:
```suggestion
		cell: ({ getValue })=> {
```
just according to other code

apps/frontend/src/pages/analytics/libs/helpers/get-analytics-rows/get-analytics-rows.helper.ts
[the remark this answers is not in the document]
sandrvvu: image
Is this format okay? I made it through raw queries. 
But I don't know how to return it correctly, cause the entity doesn't match this data type now?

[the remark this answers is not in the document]
sandrvvu: image
Does this format work? I've tried several, and things are starting to get a bit confusing... I implemented this using a raw query. But now I’m wondering how to work with entities in this format because it no longer matches.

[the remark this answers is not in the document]
liza-veis: I think the previous format was more expected, let's use grouping on the js side then

apps/backend/src/modules/activity-logs/activity-log.controller.ts
```diff
@@ -149,11 +157,18 @@ class ActivityLogController extends BaseController {
...
+		options: APIHandlerOptions<{
+			query: ActivityLogQueryParameters;
+		}>,
+	): Promise<APIHandlerResponse> {
+		const { endDate, startDate } = options.query;
 
 		return {
-			payload: activityLogs,
+			payload: await this.activityLogService.findAllWithParameters({
+				endDate: endDate || new Date().toDateString(),
```
liza-veis, line 169: I think we always need to pass startDate in query params. And we also need to add validation for query for this endpoint
```suggestion
				endDate,
```

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -136,6 +139,60 @@ class ActivityLogService implements Service {
 		};
 	}
 
+	public async findAllWithParameters({
```
liza-veis, line 142: let's name just findAll
```suggestion
	public async findAll({
```

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -136,6 +139,60 @@ class ActivityLogService implements Service {
...
+		const dateRange = getDateRange(startDate, endDate);
+
+		const NO_COMMITS = 0;
+		const contributorMap: Record<string, number[]> = {};
+
+		for (const contributor of allContributors.items) {
+			const uniqueKey = `${contributor.name}_${String(contributor.id)}`;
+			contributorMap[uniqueKey] = Array.from(
+				{ length: dateRange.length },
+				() => NO_COMMITS,
```
liza-veis, line 165:
```suggestion
			contributorMap[uniqueKey] = new Array(dateRange.length).fill(INITIAL_COMMITS_NUMBER)
```

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -136,6 +139,60 @@ class ActivityLogService implements Service {
...
+		}
+
+		return {
+			items: Object.entries(contributorMap).map(([uniqueKey, commitsArray]) => {
+				const [contributorName, contributorId] = uniqueKey.split("_");
+
+				return {
+					commitsNumber: commitsArray,
+					contributorId: contributorId || "",
+					contributorName: contributorName || "",
```
liza-veis, line 190:
```suggestion
					contributorId: contributorId ?? "",
					contributorName: contributorName ?? "",
```

apps/frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
```diff
@@ -32,6 +33,7 @@ type ReturnValue<T extends FieldValues = FieldValues> = {
 	handleValueSet: UseFormSetValue<T>;
 
 	isDirty: boolean;
+	watch: UseFormWatch<T>;
```
liza-veis, line 36: please use useWatch as we already use it in the project
```suggestion
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -1,28 +1,85 @@
...
+	const initialDateRange: [Date, Date] = [
+		subtractDays(todayDate, ANALYTICS_DATE_MAX_RANGE),
+		todayDate,
+	];
+
+	const { activityLogs, dataStatus } = useAppSelector(
+		({ activityLogs }) => activityLogs,
+	);
+
+	const [dateRange, setDateRange] = useState<[Date, Date]>(initialDateRange);
```
liza-veis, line 32: we are already using watch for date range, why do we need this?
```suggestion
```

liza-veis (changes requested):

liza-veis (approved):

PR #372, opened 2024-09-17
fix: refactored search handling in search component gf-331

Fjortis:
1. Refactored search handling in modals: The search functionality in modals has been improved to reset the search state each time the modal is reopened by utilizing the form's default values.

2. Simplified state management: the form now resets search input based on form initialization

https://github.com/user-attachments/assets/2917b107-dbba-4202-86b3-09bb5306e96c


apps/frontend/src/libs/hooks/use-search/use-search.hook.ts
```diff
@@ -13,14 +14,20 @@ const useSearch = (): {
...
 		setSearchParameters(updatedSearchParameters);
 	}, [search, searchParameters, searchParameterName, setSearchParameters]);
 
+	const resetSearch = (): void => {
+		setSearch("");
+		const updatedSearchParameters = new URLSearchParams(searchParameters);
+		updatedSearchParameters.delete(searchParameterName);
+		setSearchParameters(updatedSearchParameters);
+	};
+
```
liza-veis, line 27: Not a the best idea to keep this in use-search hook. There is a reset method in useForm, I think we should use it

liza-veis (changes requested):

apps/frontend/src/libs/hooks/use-search/use-search.hook.ts
Fjortis: On first thought I agree that useForm's reset is the right way to handle resetting the form fields. However, useSearch is also managing the URL query parameters, which resetForm can't handle on its own. To fully clear the search, we need both: resetFor for the input and resetSearch in useSearch to remove the query parameter from the URL. 

Should i just create a new search hook for Modals or should i rewrite the useSearch hook to be able to handle both?

liza-veis: Shouldn't the query parameters update be triggered by a search change? As I see from the logic when the value inside the Search component is updated, onSearch (from useSearch) is triggered and it updates the query params value in useEffect, doesn't it?

Fjortis: I see what you're saying about useForm's reset. Just to clarify, are you suggesting that I manage the search input directly within useForm and update the query parameters through the form's state, rather than using a separate useSearch hook for this? That way, resetForm would handle both the input and query parameter reset together.

Fjortis: yes, I see what you mean, i will get this sorted asap

apps/frontend/src/libs/hooks/use-search/use-search.hook.ts
```diff
@@ -3,6 +3,7 @@ import { useEffect, useSearchParams, useState } from "~/libs/hooks/hooks.js";
 
 const useSearch = (): {
 	onSearch: (search: string) => void;
+	resetSearch: () => void;
```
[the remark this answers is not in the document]
Fjortis: This has been adjusted.

apps/frontend/src/libs/hooks/use-search/use-search.hook.ts
Fjortis: I have simplified it, and made use of the `UseFormReset `instead. I think this is cleaner and more reusable approach.

what1s1ove (approved):

apps/frontend/src/libs/hooks/use-search-reset/use-search-reset.hook.ts
```diff
@@ -0,0 +1,13 @@
+import { useAppForm, useCallback } from "~/libs/hooks/hooks.js";
+
+const useSearchReset = (defaultSearch: string): (() => void) => {
+	const { handleReset } = useAppForm({
```
liza-veis, line 4: We don't need to use separate useAppForm, we need to use useAppForm that controls the Search. I see it's defined in the search itself, we need to move it out of Search component and pass control and errors to it.

liza-veis (changes requested):

apps/frontend/src/libs/hooks/use-search-reset/use-search-reset.hook.ts
Fjortis: Gotcha, will fix it right away

Fjortis: I made it work, but gotta refactor a little bit for better reusability

apps/frontend/src/libs/components/search/search.tsx
```diff
@@ -1,48 +1,51 @@
...
 
-	const value = useFormWatch({ control, name: "search" });
-
 	useEffect(() => {
 		const debouncedOnChange = initDebounce(() => {
-			onChange(value);
-			onSearch(value);
+			if (typeof value === "string") {
+				onChange(value);
+			}
```
liza-veis, line 40:
```suggestion
				onChange(value ?? "");
```

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -85,7 +88,11 @@ const GroupUsersTable = ({
 		<>
 			<div>
 				<span className={styles["table-title"]}>Users</span>
-				<GroupUsersSearch onChange={handleSearchChange} />
+				<GroupUsersSearch
+					control={control}
+					errors={formErrors}
+					onChange={handleSearchChange}
+				/>{" "}
```
liza-veis, line 95:
```suggestion
				/>
```

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -85,7 +88,11 @@ const GroupUsersTable = ({
 		<>
 			<div>
 				<span className={styles["table-title"]}>Users</span>
-				<GroupUsersSearch onChange={handleSearchChange} />
+				<GroupUsersSearch
```
liza-veis, line 91: looks like we need to pass name to the search too, because inside it we don't know how we named it in form values

apps/frontend/src/pages/access-management/libs/components/group-users-table/libs/components/group-users-search/group-users-search.tsx
```diff
@@ -1,15 +1,29 @@
...
+};
+
 type Properties = {
+	control: Control<SearchForm>;
+	errors: FieldErrors<SearchForm>;
 	onChange: (search: string) => void;
 };
 
-const GroupUsersSearch = ({ onChange }: Properties): JSX.Element => {
+const GroupUsersSearch = ({
```
liza-veis, line 17: We need to infer the type using generic here

liza-veis (changes requested):

apps/frontend/src/pages/project-access-management/libs/components/project-group-users-table/libs/components/project-group-users-search/project-group-users-search.tsx
```diff
@@ -1,17 +1,35 @@
...
+	type FieldPath,
+} from "react-hook-form";
+
 import { Search } from "~/libs/components/components.js";
+import { type SearchForm } from "~/libs/types/types.js";
 
 import styles from "./styles.module.css";
 
 type Properties = {
+	control: Control<SearchForm>;
```
liza-veis, line 13: let's remove SearchForm type and use generic too

liza-veis (changes requested):

apps/frontend/src/pages/project-access-management/libs/components/project-group-users-table/libs/components/project-group-users-search/project-group-users-search.tsx
Fjortis: Now that is fixed.

liza-veis (approved):

Issue #274, opened 2024-09-11
feat: Add project last activity label to project list


liza-veis:
### What feature?

- [ ] Add project's last activity label to the project list.
- [ ] This label should show a relative date of the last activity log with more than zero commits. We don't track commit time, so instead of minutes we will display it as date units, e.g. `Updated today`.
- [ ] Time limits for the indicators are the following:
  - Green: < 2 days
  - Yellow: < 5 days
  - Red: >= 5 days

### Screenshots

![Image](https://github.com/user-attachments/assets/4445eceb-e0de-4537-8d73-8f3109884854)

AnnaRomaniuta:
**User Story:** 
As a user, I want to see the relative date of the last commit for each project in the project list. So that I can quickly understand how recently a project was updated. The color indicators help quickly identify how actively the project is being updated.

**Acceptance Criteria:**
1. A "Last Activity" label is added to each project in the project list.
2. The label displays the relative date of the last activity with more than zero commits, in a format such as "Updated today."
3. The label changes color based on the time since the last activity:
    - Green: last activity was less than 2 days ago.
    - Yellow: last activity was between 2 and 5 days ago.
    - Red: last activity was 5 or more days ago.
4. If there is no activity, no label is shown for the project.

kolibri753:
1. Should I create a new type `ProjectGetAllWithActivityResponseDto` or model `ProjectActivityModel` for this "last activity" label?
2. This is a helper - `projects/libs/helpers/get-relative-date-label.tsx`

> Green: last activity was less than 2 days ago.
> Yellow: last activity was between 2 and 5 days ago.
> Red: last activity was 5 or more days ago.

liza-veis:
>Should I create a new type ProjectGetAllWithActivityResponseDto or model ProjectActivityModel for this "last activity" >label? Or should I just modify ProjectGetAllItemResponseDto by adding lastActivityDate
>This is a helper - projects/libs/helpers/get-relative-date-label.tsx

Let's add last_activity_date column to the projects table, null by default. And we need to update it when we successfully create activity logs for this project

VeronikaHordus:
Seems to be good! Thank you

PR #379, opened 2024-09-17
feat: add notification for failed project activity update gf-336

s1rserg:
* Added notifications icon and popup on UI.
* During initial render, all notifications are loaded.
* Add notifications table and module to the backend.
image
image
image
image
## TODO
After merge of #274:
1. Get rid of mock data on frontend and use relative date helper for notifications timestamp
2. Schedule cron job for projects that have last_activity_date not null.


apps/frontend/src/assets/images/icons/notifications.svg
Vitaliistf, line 1: As I remember we must remove fill from the icons, or give them fill of `currentColor`.

packages/shared/src/modules/notifications/libs/types/notification-create-response-dto.type.ts
Vitaliistf, line 1: I see that we have same code in NotificationGetAllItemResponseDto. Can't we reuse it?

apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
```diff
@@ -0,0 +1,50 @@
...
+							timestamp="5 minutes ago"
+						/>
+					</div>
+				</div>
+			}
+			isOpened={isOpened}
+			onClose={onClose}
+		>
+			{children}
+		</Popover>
```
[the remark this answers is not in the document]
s1rserg: I believe mark as read will be implemented in other issue. 
Can you please explain more what you mean by empty state for notification item? I added empty placeholder but not sure it's what you mean.

s1rserg:
For now notifications are loaded when the page loads. I believe it is what would be needed for the notification counter on top of icon for other issue. But if it's wrong I can change it.

apps/frontend/src/assets/images/icons/notifications.svg
s1rserg: Thanks for noticing. Hope I did this right now.

apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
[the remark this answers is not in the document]
Fjortis: I see that you added the empty placeholder, that is what i was talking about.

apps/backend/src/modules/projects/projects.ts
```diff
@@ -13,4 +18,16 @@ const projectService = new ProjectService(
...
 const projectController = new ProjectController(logger, projectService);
 
+const inactiveProjectsNotifier = new InactiveProjectsNotifier({
+	logger,
+	notificationService,
+	projectService,
+	userService,
+});
+
+taskScheduler.start(
```
what1s1ove, line 28: we should add a new method to the app server. and init all jobs in this method

apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -75,6 +75,20 @@ class ProjectRepository implements Repository {
 		return item ? ProjectEntity.initialize(item) : null;
 	}
 
+	public async findInactiveProjects(
+		thresholdInDays: number,
+	): Promise<ProjectEntity[]> {
+		const currentDate = new Date();
+		currentDate.setDate(currentDate.getDate() - thresholdInDays);
```
liza-veis, line 82: use date helper to subtract dates

apps/backend/src/modules/projects/libs/enums/notification-project-permission-key.enum.ts
```diff
@@ -0,0 +1,7 @@
+const NotificationProjectPermissionKey = {
```
liza-veis, line 1: Move it to shared near the PermissionKey, we will reuse it
```suggestion
const ProjectPermissionKey = {
```

apps/backend/src/modules/projects/inactive-projects-notifier.ts
```diff
@@ -0,0 +1,105 @@
...
+					NotificationRootPermissionKey.VIEW_ALL_PROJECTS,
+				]);
+
+				const users = [
+					...projectUsers,
+					...rootUsers.filter(
+						(rootUser) =>
+							!projectUsers.some(
+								(projectUser) => projectUser.id === rootUser.id,
+							),
```
liza-veis, line 84: Maybe we can split this to some constants, or wrap it wich brackets, it's hard to understand what's happening here.

apps/backend/src/modules/projects/inactive-projects-notifier.ts
```diff
@@ -0,0 +1,105 @@
...
+			if (!project.lastActivityDate) {
+				continue;
+			}
+
+			const daysInactive = this.calculateInactiveDays(project.lastActivityDate);
+
+			if (
+				daysInactive === NotificationDayThreshold.FIRST_THRESHOLD ||
+				daysInactive % NotificationDayThreshold.SECOND_THRESHOLD ===
+					EMPTY_LENGTH
```
liza-veis, line 61: Let's store NotificationDayThreshold as array
```suggestion
const [firstThreshold, secondThreshold] = NOTIFICATION_DAY_THRESHOLD

			if (
				daysInactive === firstThreshold ||
				daysInactive % secondThreshold ===
					EMPTY_LENGTH
```

apps/backend/src/modules/projects/inactive-projects-notifier.ts
```diff
@@ -0,0 +1,105 @@
...
+
+	public async processInactiveProjects(): Promise<void> {
+		const projects = await this.projectService.findInactiveProjects(
+			NotificationDayThreshold.FIRST_THRESHOLD,
+		);
+
+		for (const project of projects) {
+			if (!project.lastActivityDate) {
+				continue;
+			}
```
liza-veis, line 54: let's return only projects where last activity not null  from repository
```suggestion
```

apps/backend/src/modules/projects/inactive-projects-notifier.ts
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	private calculateInactiveDays(lastActivityDate: string): number {
+		const lastActivity = new Date(lastActivityDate);
+		const today = new Date();
+
+		return getDifferenceInDays(today, lastActivity);
+	}
+
+	public async processInactiveProjects(): Promise<void> {
```
liza-veis, line 46: let's remove InactiveProjectsNotifier and move this logic to projects service

liza-veis (changes requested):

packages/shared/src/libs/helpers/format-relative-time/format-relative-time.helper.ts
```diff
@@ -0,0 +1,9 @@
+import { formatDistanceToNowStrict } from "date-fns";
+
+const formatRelativeTime = (date: string): string => {
```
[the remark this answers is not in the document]
s1rserg: The other helper getRelativeDate returns today and yesterday and in figma we have 5 minutes ago. Should I keep this or use the getRelativeDate?

apps/backend/src/modules/projects/inactive-projects-notifier.ts
s1rserg: They shouldn't be returned. That was just a way to avoid ts errors, but i will make it nicer.

what1s1ove (approved):
lgtm after Lisa's comments

liza-veis (changes requested):

liza-veis (approved):

PR #380, opened 2024-09-17
fix: Make project group name unique only within the scope of a project gf-361

perekotypole:
- unique project group key is fixed by adding project id
- changed checking of existing group

### DIFF
--- a/apps/backend/src/modules/project-groups/libs/helpers/generate-project-group-key.helper.ts
+++ b/apps/backend/src/modules/project-groups/libs/helpers/generate-project-group-key.helper.ts
@@ -1,0 +1,15 @@
+import { changeCase } from "~/libs/helpers/helpers.js";
+
+const generateProjectGroupKey = ({
+	name,
+	projectId,
+}: {
+	name: string;
+	projectId: number;
+}): string => {
+	const key = [projectId, name].join(" ");
+
+	return changeCase(key, "snakeCase");
+};
+
+export { generateProjectGroupKey };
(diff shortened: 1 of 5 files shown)


packages/shared/src/libs/helpers/change-case/change-case.helper.ts
```diff
@@ -1,10 +1,14 @@
 import { caseTypeToFunction } from "./libs/maps/maps.js";
 import { type CaseType } from "./libs/types/types.js";
 
-const changeCase = (input: string, caseType: CaseType): string => {
+const changeCase = (
+	input: Array<string> | string,
```
what1s1ove, line 5:
```suggestion
	input: string,
```
you need to do it before passing as an argument to the function

apps/backend/src/modules/project-groups/libs/helpers/generate-project-group-key.helper.ts
```diff
@@ -0,0 +1,15 @@
+import { changeCase } from "~/libs/helpers/helpers.js";
+
+const generateProjectGroupKey = ({
```
liza-veis, line 3: We don't need this helper, just concat the string where it's used `${projectId}_${changeCase(name, "snakeCase")}`

liza-veis (changes requested):

apps/backend/src/modules/project-groups/libs/helpers/generate-project-group-key.helper.ts
perekotypole: Okay, I just thought it would be too much to copy this three times

liza-veis: Oh, I thought we only needed this in one place. In this case yes, better to have this helper

perekotypole: I returned it

liza-veis (approved):

PR #384, opened 2024-09-18
feat: added placeholder to contributors list on project page gf-377

Vitaliistf:
Added a placeholder, which looks like this:

Screenshot 2024-09-18 at 13 36 09

P.S. had to split conditions in order to get rid of linter problems. Linter does not allow to nest ternary expression without parentheses, but formatter wants that expression without them. If there is a better solution, please suggest it.

### DIFF
--- a/apps/frontend/src/pages/project/libs/components/contributors-list/contributors-list.tsx
+++ b/apps/frontend/src/pages/project/libs/components/contributors-list/contributors-list.tsx
@@ -15,12 +15,14 @@ const ContributorsList = ({
 	isLoading,
 }: Properties): JSX.Element => {
 	const hasContributors = contributors.length > EMPTY_LENGTH;
+	const isListShown = !isLoading && hasContributors;
+	const isEmptyPlaceholderShown = !isLoading && !hasContributors;
 
 	return (
 		<div className={styles["container"]}>
 			<h2 className={styles["title"]}>Contributors</h2>
 			{isLoading && <Loader />}
-			{!isLoading && hasContributors && (
+			{isListShown && (
 				<ul className={styles["list"]}>
 					{contributors.map((contributor) => (
 						<li key={contributor.id}>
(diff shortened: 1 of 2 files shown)


apps/frontend/src/pages/project/libs/components/contributors-list/contributors-list.tsx
```diff
@@ -13,12 +14,13 @@ const ContributorsList = ({
...
+	const hasContributors = contributors.length > EMPTY_LENGTH;
+
 	return (
 		<div className={styles["container"]}>
 			<h2 className={styles["title"]}>Contributors</h2>
-			{isLoading ? (
-				<Loader />
-			) : (
+			{isLoading && <Loader />}
+			{!isLoading && hasContributors && (
```
perekotypole, line 23: How about returning <Loader/> before, and don't use so many conditions for each block?

Vitaliistf: Looks like a nice solution, let's see what will Liza say about it. The only thing is that we won't have this layout of a component at all (container and title). I just tried to do it as we've done before in table and projects page.

liza-veis: I think in this case we will miss the title, so I would leave it

liza-veis:
I think it would be better to add some space between title and placeholder

Vitaliistf:
> I think it would be better to add some space between title and placeholder

Added more space.

apps/frontend/src/pages/project/libs/components/contributors-list/contributors-list.tsx
liza-veis, line 23: let's move these conditions to separate constants - isListShown and isEmptyPlaceholderShown
```suggestion
			{isListShown && (
```

liza-veis (changes requested):

liza-veis (approved):

PR #391, opened 2024-09-18
feat: fix quality criteria issues gf-382

Vitaliistf:
Fixed quality criteria issues in mentioned files. Also looked through the whole project to find other issues like these.


Vitaliistf:
While looking through the code, found some things that might be improved, in my opinion. I'm not sure about it so please, let me know if I should change this code.

1. I think we can simplify this, remove `hasError` constant and use `error` to check if it's truthy.
Screenshot 2024-09-18 at 15 39 16

2. Menu item component is intended to be used in menu, so we could move it to menu's components and re-export it there.
Screenshot 2024-09-18 at 15 43 58

GvoFor:
1. We are trying to stick to the following approach: in the constructions like `{condition && <component>}` condition type should be boolean, so that's why `hasError` used here. I think you shouldn't remove it.

2. I'm not sure, but I'd like to suggest the following: 
- `MenuItem` should be local component relative to `Menu`
- Others components shouldn't know about `MenuItem` -- there should be some type describing `MenuItem` props, and we should pass an array of such a type to `Menu` component, and `Menu` is responsible to convert this array into array of `MenuItem` objects (which it will render). So I image menu usage as:
```ts
<Menu
  isOpened=...
  onClose=...
  onOpen=...
  items={[
    { icon: ..., label: ..., ... }
  ]}
/>
```

liza-veis:
@GvoFor I agree with your approach, but for now let's leave it as it is, at least in this ticket

apps/frontend/src/libs/components/table/styles.module.css
```diff
@@ -65,7 +65,7 @@
 	table-layout: fixed;
 }
 
-.table .table-row:last-child .table-data {
+.table-row:last-child .table-data {
```
liza-veis, line 68: Maybe we need to remove .table for all other selectors here to keep it consistent?

Vitaliistf: Yeah, makes sense. Thank you!

perekotypole:
Could you move `apps\backend\src\modules\project-groups\libs\helpers\generate-project-group-key.helper.ts` as well as you did with `getValidClassNames` if we keep helpers in folders everywhere?

Vitaliistf:
> Could you move `apps\backend\src\modules\project-groups\libs\helpers\generate-project-group-key.helper.ts` as well as you did with `getValidClassNames` if we keep helpers in folders everywhere?

Sure, good notice, Thanks!

what1s1ove (approved):

liza-veis (approved):

PR #397, opened 2024-09-18
feat: replace forever with pm2 for analytics script gf-385

s1rserg:
* Replaced forever with pm2 for analytics script.
* Added a part of auto start functionality.

## How to run
1. `npm run build:analytics`
2. `node scripts/analytics/build/index.js track apiKey userId repoPath`

## How to delete process without rebooting
1. `npm install pm2 -g`
2. `Set-ExecutionPolicy -Scope Process -ExecutionPolicy Bypass`  (optional)
3. `pm2 delete all`

## How to setup AutoStart on Unix
1. `pm2 startup`
The result in terminal will be like this:
`[PM2] You have to run this command as root. Execute the following command:
      sudo su -c "env PATH=$PATH:/home/unitech/.nvm/versions/node/v14.3/bin pm2 startup <distribution> -u <user> --hp <home-path>`
2. Take the string from previous step, populate with your values and execute it in terminal.
3. Run the script (on start it will be saved so it will respawn after reboot).


what1s1ove (approved):

scripts/analytics/src/libs/helpers/helpers.ts
```diff
@@ -1 +1,2 @@
+export { executeCommand } from "./execute-command.helper.js";
```
liza-veis, line 1:
```suggestion
export { executeCommand } from "./execute-command/execute-command.helper.js";
```

liza-veis (changes requested):

liza-veis (approved):

PR #399, opened 2024-09-19
feat: handle dates that cross into the next day but are less than 24 hours apart gf-274

kolibri753:
- created a new helper `getStartOfDay` (wrapper around `startOfDay` from "date-fns");
- wrapped creating dates in `ProjectCard` with this helper.

### Screenshots
image
image

### DIFF
--- a/packages/shared/src/libs/helpers/get-start-of-day/get-start-of-day.helper.ts
+++ b/packages/shared/src/libs/helpers/get-start-of-day/get-start-of-day.helper.ts
@@ -1,0 +1,7 @@
+import { startOfDay } from "date-fns";
+
+const getStartOfDay = (date: Date | number | string): Date => {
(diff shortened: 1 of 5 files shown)


packages/shared/src/libs/helpers/get-start-of-day/get-start-of-day.helper.ts
```diff
@@ -0,0 +1,7 @@
+import { startOfDay } from "date-fns";
+
+const getStartOfDay = (date: Date | number | string): Date => {
```
liza-veis, line 3: we need to reduce variations
```suggestion
const getStartOfDay = (date: Date): Date => {
```

liza-veis (changes requested):

liza-veis (approved):

PR #400, opened 2024-09-19
feat: add an ability to mark contributors as hidden gf-346

kolibri753:
### Changes:

- added new icons: check, circle-question;
- added is_hidden column to contributors table by migration;
- added isHidden to types, model, entity, repository;
- adjusted contributors table;
- excluded hidden contributors from project contributors page;
- excluded hidden contributors from analytics page.

### Screenshots:
image

image

image

image


apps/frontend/src/pages/contributors/libs/components/contributor-update-form/contributor-update-form.tsx
```diff
@@ -68,6 +73,21 @@ const ContributorUpdateForm = ({
...
 			/>
+			<div className={styles["checkbox-wrapper"]}>
+				<label className={styles["checkbox-label"]} htmlFor="isHidden">
+					<Checkbox
+						control={control}
+						errors={errors}
+						id="isHidden"
+						name="isHidden"
+					/>
+					<span>Is hidden</span>
```
Fjortis, line 84: I understand that you didn't choose this name, but am I the only one who thinks "Is Hidden" sounds more like developer language than something suited for a user interface?
"Do Not Track", or perhaps just "Hidden"

apps/frontend/src/assets/images/icons/check.svg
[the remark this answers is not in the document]
kolibri753: For me, `check` feels more familiar. Before naming it, I checked [FontAwesome](https://fontawesome.com/search), and the name there is also `check`. Additionally, I haven't found us using `-mark` in this project. So, I would prefer to leave it unchanged :cat2:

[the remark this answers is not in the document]
GvoFor: Reasonable. Fine

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -176,9 +176,14 @@ class ActivityLogService implements Service {
...
 
+				const contributor = allContributors.items.find(
+					(contributor) => contributor.id === Number(contributorId),
+				);
+
 				return {
 					commitsNumber: commitsArray,
 					contributorId: contributorId ?? "",
+					contributorIsHidden: contributor ? contributor.isHidden : false,
 					contributorName: contributorName ?? "",
```
liza-veis, line 187: maybe we can return contributor as an object and id, isHidden, and name as properties, or is there an issue? cc @sandrvvu

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -68,6 +68,10 @@ const Analytics = (): JSX.Element => {
 		}
 	}, [dateRangeValue, isDirty, handleFormSubmit]);
 
+	const visibleActivityLogs = activityLogs.filter(
+		(log) => !log.contributorIsHidden,
+	);
```
liza-veis, line 73: this should be filtered inside backend query

apps/frontend/src/pages/contributors/libs/components/contributor-update-form/contributor-update-form.tsx
liza-veis: good suggestion, I like do not track option (on the backend side let's keep it as it is now)

apps/frontend/src/pages/contributors/libs/components/is-hidden-header/is-hidden-header.tsx
```diff
@@ -0,0 +1,11 @@
+import { Icon } from "~/libs/components/components.js";
+
+import styles from "./styles.module.css";
+
+const IsHiddenHeader = (): JSX.Element => (
+	<div className={styles["header-with-icon"]}>
+		Is Hidden <Icon height={15} name="circleQuestion" width={15} />
```
liza-veis, line 7: In the design, this icon is meant to be with a tooltip, but if we rename the column on UI to Do Not Track, then I think we can remove this icon at all, especially taking into consideration that it's described inside edit modal

apps/frontend/src/pages/contributors/libs/helpers/get-contributor-columns/get-contributor-columns.helper.tsx
```diff
@@ -23,6 +27,15 @@ const getContributorColumns = (actions: {
 		header: "Projects",
 		size: 300,
 	},
+	{
+		cell: ({ row: { original: contributor } }) =>
```
liza-veis, line 31: let's wrap this expression with curly braces

apps/frontend/src/pages/project/project.tsx
```diff
@@ -166,6 +166,10 @@ const Project = (): JSX.Element => {
 		userPermissions,
 	);
 
+	const visibleProjectContributors = projectContributors.filter(
```
liza-veis, line 169: same, should be returned this way from the backend

liza-veis (changes requested):

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -176,9 +176,14 @@ class ActivityLogService implements Service {
...
 				const [contributorName, contributorId] = uniqueKey.split("_");
 
+				const contributor = allContributors.items.find(
+					(contributor) => contributor.id === Number(contributorId),
+				);
+
 				return {
 					commitsNumber: commitsArray,
 					contributorId: contributorId ?? "",
+					contributorIsHidden: contributor ? contributor.isHidden : false,
```
what1s1ove, line 186:
```suggestion
					contributorIsHidden: contributor?.isHidden ?? false,
```

what1s1ove (approved):
lgtm after Lisa's changes

apps/frontend/src/pages/analytics/analytics.tsx
kolibri753: Is it mandatory to modify query like this `builder.select("id", "name").where("isHidden", false);` or can it be just check in service 			`if (!contributor.isHidden)`, not sure it is needed to add isHidden to contributor on `GitEmailModel`.

(I'll push it with just a check for a better look)

what1s1ove (approved):

packages/shared/src/modules/activity-logs/libs/types/activity-log-get-all-item-analytics-response-dto.type.ts
```diff
@@ -1,6 +1,7 @@
 type ActivityLogGetAllItemAnalyticsResponseDto = {
 	commitsNumber: (number | string)[];
 	contributorId: string;
+	contributorIsHidden: boolean;
 	contributorName: string;
```
liza-veis, line 5: so what is with the question regarding moving this props inside contributor object

apps/frontend/src/pages/analytics/analytics.tsx
liza-veis: we don't need isHidden in GitEmailModel for now, it's better to use query builder

packages/shared/src/modules/activity-logs/libs/types/activity-log-get-all-item-analytics-response-dto.type.ts
kolibri753: I haven't found any issues with but I still can't figure out why I'm getting the error `[Cannot destructure property 'id' of 'gitEmail.contributor' as it is null.]`. I'm trying to understand why by using the following query:

```
const activityLogs = await this.activityLogModel
			.query()
			.withGraphFetched("[gitEmail.contributor, project, createdByUser]")
			.modifyGraph("gitEmail.contributor", (builder) => {
				builder.select("id", "name").where("isHidden", false);
			})
			.whereBetween("activity_logs.date", [startDate, endDate])
			.orderBy("date");
```

I see that contributor is coming back as `null`.

GvoFor: Oh, I also faced similar problem. I guess the problem is how graph fetcing works, and how you used it. 
You are trying to filter main query (`SELECT ... FROM activity_logs`) by the condition, that contributor is hidden, but your code just selects all the activity logs in some date range, and then fetches to those selected activity logs contibutors data  (other stuff doesn't matter right now), but you add additional `.where` filter, which just replaced contributors with `isHidden=true` to null in activity_logs where it is the case.
Try to use [withGraphJoined](https://vincit.github.io/objection.js/api/query-builder/eager-methods.html#withgraphjoined) instead of fetched, and then add `.where` filter to main query

GvoFor: Something like
```ts
const activityLogs = await this.activityLogModel
	.query()
	.withGraphJoined("[gitEmail.contributor, project, createdByUser]")
	.modifyGraph("gitEmail.contributor", (builder) => {
		builder.select("id", "name");
	})
	.where("contributor.isHidden", false)
	.whereBetween("activity_logs.date", [startDate, endDate])
	.orderBy("date");
```

GvoFor: image

GvoFor: Using this info probably the query should looks like
```ts
const activityLogs = await this.activityLogModel
	.query()
	.withGraphFetched("[project, createdByUser]")
	.withGraphJoined("gitEmail.contributor")
	.where("gitEmail:contributor.isHidden", false)
	.whereBetween("activity_logs.date", [startDate, endDate])
	.orderBy("date");
```

and also we need to add something which will do
```ts
.modifyGraph("gitEmail.contributor", (builder) => {
	builder.select("id", "name");
})
```
cause this method won't work with joining the graph. Or will be, I'm not sure.

kolibri753: Thanks a lot, I've tried smth similar to this, last time it didn't give the needed result. But hope this time it will :)

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -150,11 +150,13 @@ class ActivityLogService implements Service {
...
 
 		for (const contributor of allContributors.items) {
-			const uniqueKey = `${contributor.name}_${String(contributor.id)}`;
-			contributorMap[uniqueKey] = Array.from(
-				{ length: dateRange.length },
-				() => INITIAL_COMMITS_NUMBER,
-			);
+			if (!contributor.isHidden) {
+				const uniqueKey = `${contributor.name}_${String(contributor.id)}`;
+				contributorMap[uniqueKey] = Array.from(
```
kolibri753, line 155: will look into this further and try to remove the check

liza-veis: So what's the issue with adding it as where condition to the query?

kolibri753: The query works fine, as we only retrieve logs from contributors who are not hidden. However, in the service, we add all contributors to the contributorMap, and at the end, we include this contributors also, but 0 logs. So, I think this check should remain.

apps/frontend/src/pages/contributors/libs/helpers/get-contributor-columns/get-contributor-columns.helper.tsx
kolibri753: hmm.. in the app we are using it like this.. not sure I understand

apps/backend/src/modules/activity-logs/activity-log.service.ts
liza-veis: I think instead we need to pass option to findAll contributors, like - hasHidden, same as we did for hasDeleted

kolibri753: If I understand correctly, you’re proposing to also filter contributors in the findAll method. While this approach will work, it will also apply the filtering directly to the contributors table

liza-veis: for the contributors table we need to pass this option as true, so nothing should be changed in that functionality

kolibri753: hope I got, thanks for patience :)

packages/shared/src/modules/activity-logs/libs/types/activity-log-query-parameters.type.ts
```diff
@@ -1,5 +1,6 @@
 type ActivityLogQueryParameters = {
 	endDate: string;
+	hasHidden?: boolean;
```
liza-veis, line 3: I meant to only pass this parameter as true on activity log service level, similarly, we did for hasDeleted parameter, where we needed to find with deleted users in one place and without them in another place. And don't we need to pass this parameter to findAllByProjectId?

liza-veis (changes requested):

packages/shared/src/modules/activity-logs/libs/types/activity-log-query-parameters.type.ts
kolibri753: Oops, now I see it clearly — it’s much simpler now. Regarding `findAllByProjectId`, I added the same check as I did for `findAllWithoutPagination`, but I already had a check in place that handles this: `.andWhere("contributors.isHidden", false)`.

After you pointed out that this should be handled on the backend, I tried implementing it the way I originally thought one more time. Surprisingly (maybe Visual Studio was bugging earlier), it's now working without adding hasHidden at all, just using the `.where("contributors.isHidden", false)` condition.

However, the new solution might be more consistent with the project's structure, similar to hasDeleted, as you mentioned.

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -56,12 +56,13 @@ class ContributorRepository implements Repository {
 	}
 
 	public async findAll({
+		hasHidden = true,
 		page,
 		pageSize,
-	}: PaginationQueryParameters): Promise<
+	}: { hasHidden?: boolean } & PaginationQueryParameters): Promise<
```
Vitaliistf, line 62: Maybe it would be better to create a separate type for this parameters intersection? (and use it in service too)

apps/backend/src/db/migrations/20240918171311_add_is_hidden_to_contributors_table.ts
```diff
@@ -0,0 +1,21 @@
+import { type Knex } from "knex";
+
+const TABLE_NAME = "contributors";
+
+const ColumnName = {
+	IS_HIDDEN: "is_hidden",
```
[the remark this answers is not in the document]
kolibri753: so I'm changing it everywhere, right?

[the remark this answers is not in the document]
liza-veis: yes

apps/frontend/src/pages/contributors/libs/components/contributor-update-form/contributor-update-form.tsx
```diff
@@ -17,32 +22,53 @@ const ContributorUpdateForm = ({
...
+	useEffect(() => {
+		handleValueSet("isHidden", isHidden);
+	}, [isHidden, handleValueSet]);
+
 	const handleFormSubmit = useCallback(
 		(event_: React.BaseSyntheticEvent): void => {
-			void handleSubmit((formData: { name: string }) => {
-				const payload: ContributorPatchRequestDto = { name: formData.name };
+			void handleSubmit((formData: { isHidden: boolean; name: string }) => {
+				const hiddenAtValue = isHiddenValue ? new Date().toISOString() : null;
```
perekotypole, line 61: Why don't you use `isHidden` instead of `isHiddenValue` and `useFormWatch`?

kolibri753: isHidden (formData.isHidden) resulted in undefined, will investigate it more

apps/backend/src/modules/contributors/contributor.repository.ts
GvoFor: Like it was done with `Project`/`ProjectGroup`/`Notification` `GetAllResponseDto`

apps/backend/src/modules/contributors/contributor.service.ts
```diff
@@ -106,9 +110,11 @@ class ContributorService implements Service {
 		};
 	}
 
-	public async findAllWithoutPagination(): Promise<ContributorGetAllResponseDto> {
+	public async findAllWithoutPagination(
+		hasHidden?: boolean,
+	): Promise<ContributorGetAllResponseDto> {
```
GvoFor, line 115: It's may be not clear what boolean parameter is used for outside of this service
```suggestion
	public async findAllWithoutPagination({
		hasHidden,
	}: { hasHidden?: boolean }): Promise<ContributorGetAllResponseDto> {
```
For example, you call  `this.contributorService.findAllWithoutPagination(false));` -- when we look at this call it is hard to guess what `false` may be used for.

Same for `findAllByProjectId`

apps/frontend/src/pages/contributors/libs/components/contributor-update-form/contributor-update-form.tsx
```diff
@@ -17,32 +22,53 @@ const ContributorUpdateForm = ({
...
+
+	useEffect(() => {
+		handleValueSet("isHidden", isHidden);
+	}, [isHidden, handleValueSet]);
+
 	const handleFormSubmit = useCallback(
 		(event_: React.BaseSyntheticEvent): void => {
-			void handleSubmit((formData: { name: string }) => {
-				const payload: ContributorPatchRequestDto = { name: formData.name };
+			void handleSubmit((formData: { isHidden: boolean; name: string }) => {
```
GvoFor, line 60: Why don't we just destucture it
```suggestion
			void handleSubmit(({ isHidden: boolean; name: string }) => {
```
BTW, no matter you annotate `formData` with `{ isHidden: boolean; name: string }` type, it still contains `gitEmails`, `name` and `projects`

apps/frontend/src/pages/contributors/libs/helpers/get-contributor-columns/get-contributor-columns.helper.tsx
```diff
@@ -37,6 +37,16 @@ const getContributorColumns = (actions: {
...
+			contributor.hiddenAt ? (
+				<Icon height={18} name="check" width={18} />
+			) : null,
```
GvoFor, line 44: To render nothing we are using this approach
```suggestion
			) : <></>,
```

packages/shared/src/modules/contributors/libs/validation-schemas/contributor-patch.validation-schema.ts
```diff
@@ -8,6 +8,7 @@ import { type ContributorPatchRequestDto } from "../types/types.js";
 
 const contributorPatch: z.ZodType<ContributorPatchRequestDto> = z
 	.object({
+		hiddenAt: z.union([z.string().nullable(), z.null()]),
```
GvoFor, line 11: Can't we do it like this?
```suggestion
		hiddenAt: z.string().nullable(),
```

GvoFor: or at least
```suggestion
		hiddenAt: z.union([z.string(), z.null()]),
```

GvoFor:
Don't forget to update `readme.md`

apps/backend/src/modules/contributors/contributor.repository.ts
liza-veis: this parameter is not a part of request dto, but is passed separately from other services. So I would leave it this way

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -86,9 +93,10 @@ class ContributorRepository implements Repository {
 	}
 
 	public async findAllByProjectId(
+		hasHidden: boolean = true,
```
liza-veis, line 96: let's pass it as a second parameter, it would be more logical to me

apps/frontend/src/pages/contributors/libs/components/contributor-update-form/contributor-update-form.tsx
liza-veis: Maybe we need to pass a default value? there is no need for setValue here, usual checkbox should be handled as inputs using useForm default flow

packages/shared/src/modules/contributors/libs/types/contributor-patch-request-dto.type.ts
```diff
@@ -1,4 +1,5 @@
 type ContributorPatchRequestDto = {
+	hiddenAt: null | string;
```
liza-veis, line 2: for request it's better to pass isHidden, and set the timestamp on the backend side 
```suggestion
	isHidden: boolean;
```

liza-veis (changes requested):

apps/frontend/src/pages/contributors/libs/components/contributor-update-form/contributor-update-form.tsx
kolibri753: yep, but if I'm not mistaken `void handleSubmit(({ isHidden, name }) => {`

liza-veis: agree with this option, void handleSubmit(({ isHidden, name }) => {, from the code it looks like formData should already have an appropriate type, so we don't need to define it again

packages/shared/src/modules/contributors/libs/types/contributor-patch-response-dto.type.ts
```diff
@@ -1,5 +1,6 @@
 type ContributorPatchResponseDto = {
 	id: number;
+	isHidden?: boolean;
```
liza-veis, line 3: Don't we return hiddenAt in response?

kolibri753: yes, sorry, :facepalm:

PR #401, opened 2024-09-19
feat: Adjust project permissions gf-351

AnnKasian:
Adjusted needed project permissions. All permissions enable functionality only for specific projects. 
Rules: 

- Projects page (sidebar and page itself): View Project, Edit Project, Manage Project. If user doesn't have root permissions, only projects enabled for user by project groups should be shown.
-  Project Access Management (button and page itself): Manage Project.
- Analytics page (sidebar and page itself): View Project, Edit Project, Manage Project. If user doesn't have root permissions, only projects enabled for user by project groups should be shown.
-  Project groups (create, edit, delete): Manage Project.
-  Edit project (option): Edit Project, Manage Project.
-  Setup analytics button and api key generation: Edit Project, Manage Project.


apps/backend/src/modules/project-groups/project-group.controller.ts
```diff
@@ -81,7 +87,13 @@ class ProjectGroupController extends BaseController {
 				this.delete(options as APIHandlerOptions<{ params: { id: string } }>),
 			method: "DELETE",
 			path: ProjectGroupsApiPath.$ID,
-			preHandlers: [checkUserPermissions([PermissionKey.MANAGE_USER_ACCESS])],
+			preHandlers: [
+				checkUserPermissions([
+					PermissionKey.MANAGE_PROJECT,
+					PermissionKey.VIEW_ALL_PROJECTS,
```
GvoFor, line 93: User with only `View All Projects` permissong shouldn't be able to delete project group

apps/backend/src/modules/project-groups/project-group.controller.ts
```diff
@@ -70,7 +70,13 @@ class ProjectGroupController extends BaseController {
 				),
 			method: "POST",
 			path: ProjectGroupsApiPath.ROOT,
-			preHandlers: [checkUserPermissions([PermissionKey.MANAGE_USER_ACCESS])],
+			preHandlers: [
+				checkUserPermissions([
+					PermissionKey.MANAGE_PROJECT,
+					PermissionKey.VIEW_ALL_PROJECTS,
```
GvoFor, line 76: User with only View All Projects permission shouldn't be able to create project group

apps/backend/src/modules/project-groups/project-group.repository.ts
```diff
@@ -122,6 +122,21 @@ class ProjectGroupRepository implements Repository {
...
+		const results = await this.projectGroupModel
+			.query()
+			.orderBy("createdAt", SortType.DESCENDING)
+			.withGraphJoined("[projects, users, permissions]")
+			.where("users.id", userId);
+
+		return results.map((projectGroup) =>
+			ProjectGroupEntity.initialize({
+				...projectGroup,
+				projectId: { id: projectGroup.projects.pop()?.id as number },
```
what1s1ove, line 135:
```suggestion
				projectId: { id: (projectGroup.projects.pop() as Type).id },
```

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
GvoFor, line 1: Hm. It seems to me like another hook. Maybe let's leave this hook without changes (except those for new APIPreHandler type) and create `checkHasProjectPermissions` (not sure about name), where you put all neccessery code

apps/frontend/src/pages/project/project.tsx
```diff
@@ -240,11 +243,47 @@ const Project = (): JSX.Element => {
...
+			allPermissions,
+		) &&
+			hasProject &&
+			authenticatedUser?.projectGroups.some(
+				(group) =>
+					group.projectId.includes(project.id) &&
+					group.permissions.some(
+						(permission) => permission.key === PermissionKey.MANAGE_PROJECT,
+					),
+			),
```
GvoFor, line 258: This looks like `checkHasProjectPermission` helper

apps/backend/src/modules/activity-logs/activity-log.repository.ts
```diff
@@ -42,23 +42,26 @@ class ActivityLogRepository implements Repository {
...
 			})
-			.whereBetween("activity_logs.date", [startDate, endDate])
-			.orderBy("date");
+			.whereBetween("activity_logs.date", [startDate, endDate]);
 
-		if (projectId) {
-			query.where("project_id", projectId);
+		if (!hasRootPermission) {
+			query.whereIn("activity_logs.projectId", projectIds);
 		}
```
Vitaliistf, line 62: I'm not sure but seems like this logic should be in service. Like one method in repository to retrieve for user with root permission and another to retrieve with project id's.

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -103,8 +104,13 @@ class ContributorRepository implements Repository {
 			)
 			.leftJoin("git_emails", "contributors.id", "git_emails.contributor_id")
 			.leftJoin("activity_logs", "git_emails.id", "activity_logs.git_email_id")
-			.leftJoin("projects", "activity_logs.project_id", "projects.id")
-			.where("projects.id", projectId)
+			.leftJoin("projects", "activity_logs.project_id", "projects.id");
+
+		if (!hasRootPermission) {
+			query.whereIn("projects.id", projectIds);
+		}
```
Vitaliistf, line 111: Same as my prev comment, looks like a business logic. But again, I'm not sure, better to clarify it with coaches.

apps/frontend/src/pages/project/project.tsx
```diff
@@ -240,11 +244,38 @@ const Project = (): JSX.Element => {
...
-		userPermissions,
+	const hasManagePermission = Boolean(
+		checkHasPermission(
+			[PermissionKey.MANAGE_ALL_PROJECTS, PermissionKey.MANAGE_PROJECT],
+			allPermissions,
+		) &&
+			hasProject &&
+			checkHasProjectPermission(authenticatedUser, project.id, [
+				PermissionKey.MANAGE_PROJECT,
+			]),
```
liza-veis, line 255: Split all of this and bellow into different variables, it's hard to understand the code this way

apps/frontend/src/pages/project/project.tsx
```diff
@@ -59,6 +58,11 @@ const Project = (): JSX.Element => {
 		projectStatus,
 	} = useAppSelector(({ projects }) => projects);
 
+	const { authenticatedUser, projectUserPermissions, userPermissions } =
+		useAppSelector(({ auth }) => auth);
+
+	const allPermissions = [...userPermissions, ...projectUserPermissions];
```
liza-veis, line 64: we don't need to combine permissions here, we need to check root permissions, and then check project permissions. If one of the check is true, then display needed functionality
```suggestion
```

apps/frontend/src/pages/project/libs/helpers/check-has-project-permission.helper.ts
```diff
@@ -0,0 +1,23 @@
...
+	permissionKeys: ValueOf<typeof PermissionKey>[],
+): boolean => {
+	return Boolean(
+		authenticatedUser?.projectGroups.some(
+			(group) =>
+				group.projectId.includes(projectId) &&
+				group.permissions.some((permission) =>
+					permissionKeys.includes(
+						permission.key as ValueOf<typeof PermissionKey>,
+					),
```
liza-veis, line 17: Too many nested expressions, split this into variables or make it more readable

apps/frontend/src/pages/project/libs/helpers/check-has-project-permission.helper.ts
```diff
@@ -0,0 +1,23 @@
+import { type PermissionKey } from "~/libs/enums/enums.js";
+import { type ValueOf } from "~/libs/types/types.js";
+import { type UserAuthResponseDto } from "~/modules/auth/auth.js";
+
+const checkHasProjectPermission = (
+	authenticatedUser: null | UserAuthResponseDto,
```
liza-veis, line 6: we need to pass only permissions here, not the whole authenticated user, the same as we do for checkHasPermissions. maybe for this it will be more convenient to store permissions in the slice as an object where key is project id and value is permission array

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
```diff
@@ -1,13 +1,20 @@
...
+	type APIPreHandler,
+} from "~/libs/modules/controller/controller.js";
 import { HTTPCode, HTTPError } from "~/libs/modules/http/http.js";
+import { type UserAuthResponseDto } from "~/modules/users/users.js";
 
-const checkUserPermissions = (routePermissions: string[]): APIPreHandler => {
-	return (request: FastifyRequest, _, done): void => {
-		const { user } = request;
+const checkUserPermissions = (
+	routePermissions: string[],
```
liza-veis, line 11: let's pass it this way, also let's pass object instead of separate parameters here
```suggestion
	permissions: PermissionKey[],
	projectPermissions: ProjectPermissionKey[]
```

liza-veis (changes requested):

what1s1ove (approved):
lgtm after Lisa's comments

apps/backend/src/modules/project-groups/project-group.repository.ts
```diff
@@ -122,6 +122,21 @@ class ProjectGroupRepository implements Repository {
...
+		const results = await this.projectGroupModel
+			.query()
+			.orderBy("createdAt", SortType.DESCENDING)
+			.withGraphJoined("[projects, users, permissions]")
+			.where("users.id", userId);
+
+		return results.map((projectGroup) =>
+			ProjectGroupEntity.initialize({
+				...projectGroup,
+				projectId: { id: (projectGroup.projects.pop() as ProjectModel).id },
```
GvoFor, line 135: I'm not sure, but pay attention, that this pop mutates `projectGroup.projects` which was spreaded, so resulted object may loose some data. Also I think we should rename `projectId` field of `ProjectGroupEntity` to `project`, as having `projectId.id` is not good

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
AnnKasian: So should I separate Permission Keys? Like now on Permission Key we have all permissions. Should I separate it everywhere or just for this one hook?

AnnKasian: now i just did 
```
const checkUserPermissions = (
	permissions: ValueOf<typeof PermissionKey>[],
	getProjectId?: (options: APIHandlerOptions) => number | undefined,
```

apps/backend/src/modules/project-groups/project-group.repository.ts
GvoFor: Here is the issue I'm talking about
image
`toBeInitialized` should contain all three projects, not two

apps/backend/src/modules/projects/project.repository.ts
```diff
@@ -70,9 +71,18 @@ class ProjectRepository implements Repository {
...
-			.query()
+	public async findAllWithoutPagination({
+		userProjectIds,
+	}: {
+		userProjectIds?: number[];
+	}): Promise<ProjectEntity[]> {
+		let query = this.projectModel.query();
+
+		if (userProjectIds && userProjectIds.length !== EMPTY_LENGTH) {
+			query = query.whereIn("id", userProjectIds);
```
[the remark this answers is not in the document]
AnnKasian: If you about 
`userProjectIds.length !== EMPTY_LENGTH`
 it`s for all projects (like roop permission) it this case when we have [] we don't filter projects and all of them appear

apps/backend/src/modules/project-groups/project-group.repository.ts
AnnKasian: Well i dont really get it, because i have correct amount of projects...

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
```diff
@@ -1,13 +1,22 @@
...
-	return (request: FastifyRequest, _, done): void => {
-		const { user } = request;
+const checkUserPermissions = (
+	permissions: ValueOf<typeof PermissionKey>[],
+	getProjectId?: (options: APIHandlerOptions) => number | undefined,
+): APIPreHandler => {
+	return (options, done): void => {
+		const user = options.user as null | UserAuthResponseDto;
+		const shouldValidateProject = Boolean(getProjectId);
+		const projectId = getProjectId?.(options);
```
liza-veis, line 19: looks like this const is unnecessary, because then you check `(projectId &&` which will always result in false if getProjectId was not passed
```suggestion
		const projectId = getProjectId?.(options);
```

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
```diff
@@ -1,13 +1,22 @@
...
 
-const checkUserPermissions = (routePermissions: string[]): APIPreHandler => {
-	return (request: FastifyRequest, _, done): void => {
-		const { user } = request;
+const checkUserPermissions = (
+	permissions: ValueOf<typeof PermissionKey>[],
+	getProjectId?: (options: APIHandlerOptions) => number | undefined,
+): APIPreHandler => {
+	return (options, done): void => {
+		const user = options.user as null | UserAuthResponseDto;
```
liza-veis, line 17: we will only call this on endpoints which require user to be present
```suggestion
		const user = options.user as UserAuthResponseDto;
```

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -87,6 +87,29 @@ class ActivityLogService implements Service {
 		}
 	}
 
+	private getAllowedProjectIds(
```
liza-veis, line 90: we need to have this validation inside project service. We need to get user project permissions inside our common get all method for projects and then return only allowed projects. If we need to get a specific project this validation should still happen on project service level

liza-veis (changes requested):

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
liza-veis: yes, I mean projectPermissions and permissions should be passed separately and inside the checkPermission hook should be checked separately. So I mean I want it to be this way:
```
const hasProjectPermission = // check project permission
const hasUserPermission = // check user permission
const hasPermission =  hasProjectPermission || hasUserPermission;
```
And the same flow I want to see on frontend

apps/backend/src/modules/project-groups/project-group.repository.ts
liza-veis: agree with the @GvoFor comment, but firstly I don't understand why do we have projects array inside project-group if the project group is linked to only one project

GvoFor: >firstly I don't understand why do we have projects array inside project-group if the project group is linked to only one project

Good question

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -103,8 +103,13 @@ class ContributorRepository implements Repository {
 			)
 			.leftJoin("git_emails", "contributors.id", "git_emails.contributor_id")
 			.leftJoin("activity_logs", "git_emails.id", "activity_logs.git_email_id")
-			.leftJoin("projects", "activity_logs.project_id", "projects.id")
-			.where("projects.id", projectId)
+			.leftJoin("projects", "activity_logs.project_id", "projects.id");
+
+		if (projectIds) {
+			query.whereIn("projects.id", projectIds);
```
GvoFor, line 109: This thing contains a bug. 

The bug is that after we joined tables and this `where` query executed, we filter out all the projects, which are not in the `projectIds` list. So this function will return list of contributors but all of the them will have `projects` fields containing only projects that are in `projectIds`, so we lost information about other projects contributors may be involved into.

Check [my commit](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/516/commits/c28407ca1a6f3a71a457d5a0df366488eb5e4a96) to see how I solved it, and try to adjust it for your case.

Here is small example of what I mean.
Here is the table from what I'd like to select all the rows, containing job 'Sales'
image
So here is expected result
image
And here is what we get using `where` function
image

And here is the code for this. I used this online PostgreSQL [compiler](https://onecompiler.com/postgresql)
```
-- create
CREATE TABLE job (
  id SERIAL PRIMARY KEY,
  name VARCHAR NOT NULL UNIQUE
);

CREATE TABLE person (
  id SERIAL PRIMARY KEY,
  name VARCHAR NOT NULL UNIQUE
);

CREATE TABLE person_to_job (
  id SERIAL PRIMARY KEY,
  personId INTEGER,
  jobId INTEGER,
  FOREIGN KEY (jobId) REFERENCES job(id),
  FOREIGN KEY (personId) REFERENCES person(id),
  UNIQUE (personId, jobId)
);

-- insert
INSERT INTO job (name) VALUES ('Sales');
INSERT INTO job (name) VALUES ('Accounting');
INSERT INTO job (name) VALUES ('Freelance');

INSERT INTO person (name) VALUES ('Clark');
INSERT INTO person (name) VALUES ('Dave');
INSERT INTO person (name) VALUES ('Ava');

INSERT INTO person_to_job (personId, jobId) VALUES (1, 1);
INSERT INTO person_to_job (personId, jobId) VALUES (1, 2);
INSERT INTO person_to_job (personId, jobId) VALUES (1, 3);
INSERT INTO person_to_job (personId, jobId) VALUES (2, 2);
INSERT INTO person_to_job (personId, jobId) VALUES (3, 1);

-- fetch 
SELECT
  p.id,
  p.name,
  ARRAY_AGG(j.name) job
FROM person p
LEFT JOIN person_to_job pj ON pj.personId=p.id
LEFT JOIN job j ON j.id=pj.jobId
-- WHERE j.name = 'Sales'
GROUP BY p.id
-- HAVING 'Sales' = ANY(ARRAY_AGG(j.name))
```

AnnKasian: The main idea is to compute the amount of commits contributors have on related to user projects. We don't only have to get contributors from these projects, but also sum amounts of commits only of these projects, so I believe it's neccessary to filter exactly the way it is now. Am I missing something?

Simply saying, according to your example, when user have permission ONLY for `Sales` project, they wouldn't see the amount of commits from `Freelance` or else.

Am I missing something?

apps/backend/src/modules/project-groups/project-group.repository.ts
AnnKasian: We don't need projects here, so removing the last and only project is completely safe. We are using an array because the database design includes a projects_to_project_groups table, as mentioned in the README. The migration for this was created a long time ago, establishing a many-to-many relationship between projects and their respective groups.

I changed the logic a little, it should be a better. However, changing the array of project Ids in an entity creates a mismatch with the database.

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -135,10 +151,20 @@ class ProjectService implements Service {
 		};
 	}
 
-	public async findAllWithoutPagination(): Promise<
-		ProjectGetAllItemResponseDto[]
-	> {
-		const projects = await this.projectRepository.findAllWithoutPagination();
+	public async findAllWithoutPagination({
+		hasRootPermission,
+		userProjectIds,
```
liza-veis, line 156: let's pass userProjectIds optionally, depending on whether user has root access. We only need to pass it when we want to check userProjectIds
```suggestion
		userProjectIds,
```

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -156,6 +182,29 @@ class ProjectService implements Service {
...
+			!userProjectIds.includes(projectId)
+		) {
+			throw new ActivityLogError({
+				message: ExceptionMessage.NO_PERMISSION,
+				status: HTTPCode.FORBIDDEN,
+			});
+		}
+
+		return projectId ? [projectId] : userProjectIds;
+	}
```
liza-veis, line 206: we don't need this helper, let's handle this for each method separately

liza-veis (changes requested):

apps/frontend/src/libs/components/protected-route/protected-route.tsx
```diff
@@ -54,13 +65,23 @@ const ProtectedRoute = ({
...
 	);
 
-	const hasRequiredPermissions = hasBasePermission && hasExtraPermission;
+	const hasProjectPermissions =
+		routeProjectPermissions.length === EMPTY_LENGTH
+			? false
+			: checkHasPermission(
+					routeProjectPermissions,
+					Object.values(projectUserPermissions).flat(),
+				);
```
liza-veis, line 74: let's simplify it
```suggestion
	const hasProjectPermissions =
		routeProjectPermissions.length !== EMPTY_LENGTH 
		&& checkHasPermission(
					routeProjectPermissions,
					Object.values(projectUserPermissions).flat(),
				);
```

apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -139,9 +139,33 @@ class ActivityLogService implements Service {
...
 		startDate,
-	}: ActivityLogQueryParameters): Promise<ActivityLogGetAllAnalyticsResponseDto> {
+		userProjectIds,
+	}: {
+		hasRootPermission: boolean;
+		userProjectIds: number[];
+	} & ActivityLogQueryParameters): Promise<ActivityLogGetAllAnalyticsResponseDto> {
+		const projectIdParsed = projectId ? Number(projectId) : undefined;
+
+		let permissionedProjectIds: number[];
```
liza-veis, line 152:
```suggestion
		let permittedProjectIds: number[];
```

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
```diff
@@ -1,27 +1,47 @@
...
+): APIPreHandler => {
+	return (options, done): void => {
+		const user = options.user as UserAuthResponseDto;
+		const projectId = getProjectId?.(options);
+
+		const userPermissions = [
+			...user.groups.flatMap((group) => group.permissions),
+			...user.projectGroups
+				.filter((group) => projectId && group.projectId === projectId)
+				.flatMap((projectGroup) => projectGroup.permissions),
```
liza-veis, line 29: looks like we no longer need this as we check project permissions separately
```suggestion
```

apps/frontend/src/pages/project/libs/helpers/chask-is-project-permitted.ts
liza-veis, line 1: there is a misspelling in fileName should be `check`

apps/frontend/src/libs/helpers/get-permitted-navigation-items/get-permitted-navigation-items.helper.ts
```diff
@@ -1,16 +1,23 @@
...
+	projectPermissions: ProjectPermissionsGetAllItemResponseDto[] = [],
 ): NavigationItem[] => {
-	return items.filter(({ pagePermissions = [] }) =>
-		checkHasPermission(pagePermissions, userPermissions),
+	return items.filter(
+		({ pagePermissions = [], pageProjectPermissions = [] }) =>
+			(pagePermissions.length !== EMPTY_LENGTH &&
+				checkHasPermission(pagePermissions, userPermissions)) ||
+			(pageProjectPermissions.length !== EMPTY_LENGTH &&
+				checkHasPermission(pageProjectPermissions, projectPermissions)),
```
liza-veis, line 20: use curly braces and split this checks into constants inside it, so that it looks more readable

apps/frontend/src/libs/components/page-layout/page-layout.tsx
```diff
@@ -13,7 +13,10 @@ const PageLayout = ({
 	children,
 	isLoading = false,
 }: Properties): JSX.Element => {
-	const { userPermissions } = useAppSelector(({ auth }) => auth);
+	const { projectUserPermissions, userPermissions } = useAppSelector(
+		({ auth }) => auth,
+	);
+	const projectPermissions = Object.values(projectUserPermissions).flat();
```
liza-veis, line 19: Not sure how it works, don't we need to access project permission by project id. What if we have different permissions for different projects?

apps/backend/src/modules/project-groups/project-group.repository.ts
```diff
@@ -122,6 +122,21 @@ class ProjectGroupRepository implements Repository {
...
+		const results = await this.projectGroupModel
+			.query()
+			.orderBy("createdAt", SortType.DESCENDING)
+			.withGraphJoined("[projects, users, permissions]")
+			.where("users.id", userId);
+
+		return results.map(({ projects, ...projectGroup }) =>
+			ProjectGroupEntity.initialize({
+				...projectGroup,
+				projectId: { id: (projects.pop() as ProjectModel).id },
```
liza-veis, line 135: Avoid mutating variables
```ts
const [project] = projects;
```

apps/frontend/src/libs/components/page-layout/page-layout.tsx
AnnKasian: On Page layout we just check permissions for sidebar. We don't have any projects here.
Maybe I don't get it right

liza-veis: Got this, then this is ok

what1s1ove (approved):

PR #410, opened 2024-09-19
feat: add last activity time to contributors gf-347

Vitaliistf:
Added last activity date to contributors on project details page.
This label shows a relative date of the last activity log with more than zero commits. 

Screenshot 2024-09-19 at 13 13 19
Screenshot 2024-09-19 at 13 13 30


what1s1ove (approved):

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -73,6 +73,11 @@ class ContributorRepository implements Repository {
 					"COALESCE(ARRAY_AGG(DISTINCT jsonb_build_object('id', projects.id, 'name', projects.name)) FILTER (WHERE projects.id IS NOT NULL), '{}') AS projects",
 				),
 			)
+			.select(
+				raw(
+					"MAX(CASE WHEN activity_logs.commits_number > 0 THEN activity_logs.date ELSE NULL END) AS last_activity_date",
```
GvoFor, line 78: Why need this check?
If I'm not wrong, if there is no commits then date is null by default

Vitaliistf: Activity log has commits number field. If last activity has 0 commits then we are looking on the one before it, and so on. There is a requirement that the label should show last activity log with more than zero commits.

GvoFor: I got it. I just thought that if `activity_logs.commits_number == 0` then `activity_logs.date` should be `null`. Isn't this the case? If this is the case, then can't we simplify query as 
```suggestion
					"MAX(activity_logs.date) AS last_activity_date",
```
If it isn't the case, then all is good

Vitaliistf: Oh, alright. Maybe I'm missing something out, but as I see, in activity logs repo and service, there is no such check or anything, the date can be present even if there are 0 commits.

liza-veis: yes, let's keep this check

liza-veis (approved):

apps/frontend/src/pages/project/libs/components/contributor-card/contributor-card.tsx
```diff
@@ -10,13 +17,36 @@ type Properties = {
 };
 
 const ContributorCard = ({ contributor, onEdit }: Properties): JSX.Element => {
+	const currentDate = getStartOfDay(new Date());
+	const lastActivityDate = contributor.lastActivityDate
+		? new Date(contributor.lastActivityDate)
+		: null;
```
kolibri753, line 23: when passing the last activity date, we should also use `getStartOfDay`

PR #412, opened 2024-09-19
feat: add users column to groups table on project access management page gf-403

kolibri753:
- added users column to groups table on project access management page.

### Screenshots

image

### DIFF
--- a/apps/frontend/src/pages/project-access-management/libs/helpers/get-project-group-columns.helper.tsx
+++ b/apps/frontend/src/pages/project-access-management/libs/helpers/get-project-group-columns.helper.tsx
@@ -20,6 +20,10 @@ const getProjectGroupColumns = (actions: {
 		header: "Permissions",
 	},
 	{
+		accessorKey: "userCount",
+		header: "Users",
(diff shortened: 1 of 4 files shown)


apps/frontend/src/pages/project-access-management/libs/helpers/get-project-group-columns.helper.tsx
kolibri753, line 1: maybe I should also add the same sizes as in root access management (not sure about permissions, because it is just one for now, will look a little strange)

kolibri753, line 1: maybe I should also add the same sizes as in root access management (not sure about permissions, because it is just one for now, will look a little strange)

liza-veis: yes, please, would be better to keep it consistent

what1s1ove (approved):

liza-veis (approved):

PR #413, opened 2024-09-19
feat: Log out user when 401 error occurs gf-343

perekotypole:
- created `UnauthorizedError`
- run logout dispatch in `handleUnauthorizedError`

### DIFF
--- a/apps/frontend/src/libs/modules/api/base-http-api.ts
+++ b/apps/frontend/src/libs/modules/api/base-http-api.ts
@@ -5,6 +5,7 @@ import {
 	HTTPCode,
 	HTTPError,
 	HTTPHeader,
+	UnauthorizedError,
 } from "~/libs/modules/http/http.js";
 import { type Storage, StorageKey } from "~/libs/modules/storage/storage.js";
 import { type ServerErrorResponse, type ValueOf } from "~/libs/types/types.js";
@@ -109,7 +110,10 @@ class BaseHTTPApi implements HTTPApi {
 		const isCustomException = Boolean(parsedException.errorType);
 
 		if (response.status === HTTPCode.UNAUTHORIZED) {
-			void this.storage.drop(StorageKey.TOKEN);
+			throw new UnauthorizedError({
+				details: "details" in parsedException ? parsedException.details : [],
+				message: parsedException.message,
+			});
 		}
 
 		throw new HTTPError({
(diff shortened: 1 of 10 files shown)


apps/frontend/src/libs/modules/api/base-http-api.ts
```diff
@@ -108,6 +108,10 @@ class BaseHTTPApi implements HTTPApi {
 
 		const isCustomException = Boolean(parsedException.errorType);
 
+		if (response.status === HTTPCode.UNAUTHORIZED) {
+			void this.storage.drop(StorageKey.TOKEN);
```
liza-veis, line 112: Will this navigate the user to the sign in?

perekotypole: yes

perekotypole: our main logout function looks the same:
```
const logout = createAsyncThunk<undefined, undefined, AsyncThunkConfig>(
	`${sliceName}/logout`,
	async (_payload, { extra }): Promise<undefined> => {
		const { storage } = extra;
		await storage.drop(StorageKey.TOKEN);
	},
);
```

perekotypole: It works thought `ProtectedRoute` component:
```
if (!authenticatedUser) {
	return <Navigate replace to={AppRoute.SIGN_IN} />;
}
```

what1s1ove (approved):

apps/frontend/src/libs/modules/api/base-http-api.ts
liza-veis: but in case we drop the token, the user is still present in a store until we reload the page, isn't it?

perekotypole: I got the point.  Should I use some sort of `window.location.href` for redirection?

liza-veis: it's ok if we won't see the notification, the problem is that we can't call window inside http service, by architecture. So we need re-thror error with custom name and also to create a middleware in redux which will handle unauthorized users and log out them

yevhenii-hladkov (approved):

liza-veis (approved):

PR #414, opened 2024-09-19
fix: Block access to project access management page when user hasn't either View or Manage All Projects permission gf-350

GvoFor:
## Fixes
- adjust Project Access Management page route, so it can be accessed only if user has Manage User Access permission AND one from the View All Projects and Manage All Projects
- adjust Project Group PATCH method


packages/shared/src/libs/types/required-permission.ts
```diff
@@ -0,0 +1,8 @@
+import { type PermissionKey as PermissionKeys } from "../enums/permission-key.enum.js";
+import { type ValueOf } from "./value-of.type.js";
+
+type PermissionKey = ValueOf<typeof PermissionKeys>;
+
+type RequiredPermission = PermissionKey | PermissionKey[];
```
liza-veis, line 6: We don't need this type it's very confusing. We either need to check needed permissions separately or pass a mode as a third argument, which will have values: 'all-of' or 'one-of'
```suggestion
```

liza-veis (changes requested):

what1s1ove (approved):
lgtm after Lisa's comments

packages/shared/src/libs/types/required-permission.ts
GvoFor: Agree that this is confising. 

My initial idea was to create two separate functions `checkHasAnyPermissions` and `checkHasAllPermissions`. But the problem with this approach is that it allows only two possible variants: either user should have all of listed permission or at least one of them. But for Project Access Management page the rule is that user should have `Manage User Access` AND eighter `Manage All Projects` OR `View All Projects` permissions, and my initial idea can't handle this.

I'll try to figure out how to solve this.

GvoFor: I can't figure out how to deal the case I mentioned without this type. The only idea that comes to my mind is to define this type as followed:
image

This way it will be used like
image

and it's much clearer then previous variant, which in this case looks like
image

image

cause this (new) way we can clearly see, that user must have all of two permissions: `Manage User Access` and another permission, which could be any of `Manage All Projects`, `View All Projects`

But I'm really not sure if this solution is what you expected to see.

Fjortis: How about we rewrite checkHasPermission to something like this? 

```
const checkHasPermission = (
  requiredPermissions: string[],
  permissions: PermissionGetAllItemResponseDto[],
  alternativePermissions?: string[] // Adding an OR condition
): boolean => {
  const hasRequired = requiredPermissions.every((permission) =>
    permissions.some((userPermission) => userPermission.key === permission)
  );

  const hasAlternative = alternativePermissions
    ? alternativePermissions.some((permission) =>
        permissions.some((userPermission) => userPermission.key === permission)
      )
    : true;

  return hasRequired && hasAlternative;
};
```

Fjortis: Now it can handle AND + OR combinations

GvoFor: Your suggestion less general, but I guess I just overcomplicated the task, and your solution is perfect for now. Thanks a lot! I'll do it like you suggest for now

GvoFor: I did it. Hope now it's less confusing

Fjortis: Current version is working and is clear, waiting for update to see what you came up with

GvoFor: Here is another solution which can fix the problem mentioned in PR title:

image
image

What should I left: current one, or one that on this screenshots?

yevhenii-hladkov (approved):

packages/shared/src/libs/types/required-permission.ts
Fjortis: I think this is even simpler and required minimal changes, but what version of CheckHasPermission do you use here?

GvoFor: Oh, sorry for confusing. `checkHasPermission` looks like

image

packages/shared/src/libs/helpers/check-has-permission/check-has-permission.helper.ts
```diff
@@ -1,17 +1,24 @@
...
+	const hasAlternativePermission = alternative
+		? alternative.some((alternativePermission) =>
+				actuals.has(alternativePermission),
+			)
+		: true;
+
+	return hasRequiredPermissions && hasAlternativePermission;
 };
 
 export { checkHasPermission };
```
Fjortis, line 24: This keeps it centralized and checks the permissions in the same place, this will handle cases where it can accept both X and Y permissions as well  A or B. It keeps the logic for the check inside of the check has permissions.  

You suggested change from the comments is simpler but will fail if we ever get 2 required permissions since it will only check OR, unless you set them separately. as basePermission and additional, which doesn't make it clear.  
I do not think this is a problem for our project since we are currently not using 2 required and an additional anywhere, but for scalability and clarity, the handling for this might be better in first iteration?

packages/shared/src/libs/types/permissions.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type PermissionKey } from "../enums/enums.js";
+import { type ValueOf } from "./types.js";
+
+type Permissions = {
```
liza-veis, line 4: Maybe instead we can pass permissions in the following format: 
```
[PermissionKey, [PermissionKey,PermissionKey]]
```

liza-veis: allow either permission key or array, if it's array then one of the permission from array is required

GvoFor: but this was the type I initially add and you said it's very confusing... 🙈😅
https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/414#discussion_r1766868924

GvoFor: Maybe I didn't get the point you meant in that comment?

GvoFor: Or maybe I didn't get your current suggession..?

Initially I made it work in such a way, that you should pass an array containing elements of type `PermissinKey | PermissionKey[]` so user must have ALL of this passed permissions. And thanks to `PermissionKey[]` we can list permissions that user must have AT LEAST one of them. So for example if we pass such an array: [A, [B,C]] the function will return true if user have two permissions: A & at least one of [B, C].

Haven't you suggest the same right now?

liza-veis: I meant to pass it as array instead of an object. But now I'm think maybe we don't need this at all. Why can't we just call checkPermissions twice? like checkHasPermissions(requiredPermission) &  checkHasPermissions([permission1, permission2]). And for the backend hooks it will be just [checkHasPermissions(requiredPermission), checkHasPermissions([permission1, permission2])]

GvoFor: Yeap. I also think so. Then I'll revert all the changes and implement this easier solution

GvoFor: Thanks

liza-veis (approved):

PR #415, opened 2024-09-19
feat: Update analytics instructions with the new flow gf-406

s1rserg:
* Updated analytics instructions.
* Updated margins to not have scroll in modal on full hd displays.
* Created scripts frontend module.
image

### DIFF
--- a/apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
+++ b/apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
@@ -162,6 +162,13 @@ const SetupAnalyticsModal = ({
 							<span className={styles["list-item-title"]}>PM2 5.4</span>: The
 							script requires PM2 5.4 to be installed on your machine.
 						</li>
+						<li>
+							<span className={styles["list-item-title"]}>
+								Unix-based system
+							</span>
+							: The script requires a Unix-based operating system (e.g., Linux
+							or macOS) to run properly.
+						</li>
 					</ul>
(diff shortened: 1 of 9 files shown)


what1s1ove (approved):

apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
```diff
@@ -143,12 +158,46 @@ const SetupAnalyticsModal = ({
 							<span className={styles["list-item-title"]}>Node.js 20</span>: The
 							script requires Node.js 20 to be installed on your machine.
 						</li>
+						<li>
```
liza-veis, line 161: I think we should also mention that for now only unix based systems are supported

liza-veis (changes requested):

liza-veis:
also instead of `Copy the command below and modify with your local repository's path:`, let's say explicitly: `Copy the command below and replace <project-page> placeholder with your local repository's path:`

apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
s1rserg: should it be in prerequisites or in installation steps title in brackets 
image
image

liza-veis: Looks good to me now

liza-veis (approved):

PR #416, opened 2024-09-19
fix: inconsistent page content layout on the contributors page with different screen resolutions gf-387

Vitaliistf:
Screenshot 2024-09-19 at 15 25 43
Screenshot 2024-09-19 at 15 26 04

Fixed: When user have a long gitEmail in the Contributors table (on Contributors page) the data from the GitEmail column overlaps the Projects column for the 1024px screen resolution.
Fixed: Inconsistent table layout on the page for the big screen resolution (2560px).

Spent some time to figure out the best way to limit table height so it won't overflow the whole page, and I think this is the best that I've found so far (we also have something similar in our layout component), and it works for different heights of the screens. I'm open to suggestions though, if you know some better way to do it, I'll be glad to hear about it.


what1s1ove (approved):

apps/frontend/src/libs/components/table/styles.module.css
```diff
@@ -37,8 +41,10 @@
 
 .table-data {
 	min-height: 60px;
+	overflow: hidden;
 	font-weight: 400;
 	line-height: 150%;
+	text-overflow: ellipsis;
 	word-wrap: normal;
```
liza-veis, line 48: Now sure if we want this, we won't be able to view the text in other ways

Vitaliistf: Ok, so what should I do with it then? If we're leaving it as it is, then text will overlap.

liza-veis (approved):

PR #441, opened 2024-09-20
fix: double dots shown nearby user checkboxes in the create/edit group/project group modal window gf-382

Vitaliistf:
Screenshot 2024-09-20 at 10 26 55


liza-veis:
Don't we need to remove overflow:hidden from data too?

Vitaliistf:
> Don't we need to remove overflow:hidden from data too?

Yeah, you're right. Sorry.

liza-veis (approved):

PR #442, opened 2024-09-20
fix: The Options popover in unaccessible when opened on Contributors and Access Managment (both Root and Projects one) pages gf-436

GvoFor:
## Fixes
- Fixed popover appearance in all the tables
- Ellipsis text overflow now shouldn't be rendered on check-box column

## Screenshots
image
image

## P.s.
Ellipsis text overflow now appears only in columns holding text values
image
image
image

## Questions
Should ellipsis text overflow appear in columns holding number values?


GvoFor:
This bug was fixed in PR #441

PR #459, opened 2024-09-20
fix: useSearch change to handle modal search gf-455

Fjortis:
This change is implemented to avoid unintended double retendering when using the useSearch where search queries aren't relevant. 
Improved flexibility: Allows the search functionality to be tailored for various contexts (e.g., modal or page search) without hardcoding specific behaviors.

### DIFF
--- a/apps/frontend/src/libs/hooks/use-search/use-search.hook.ts
+++ b/apps/frontend/src/libs/hooks/use-search/use-search.hook.ts
@@ -1,24 +1,24 @@
 import { QueryParameterName } from "~/libs/enums/enums.js";
 import { useEffect, useSearchParams, useState } from "~/libs/hooks/hooks.js";
 
-type SearchOptions = {
-	shouldUseQuery?: boolean;
+type Parameters = {
+	isSavedToUrl?: boolean;
 };
 
-const useSearch = ({ shouldUseQuery = true }: SearchOptions = {}): {
+const useSearch = ({ isSavedToUrl = true }: Parameters = {}): {
 	onSearch: (search: string) => void;
 	search: string;
 } => {
 	const [searchParameters, setSearchParameters] = useSearchParams();
 	const searchParameterName = QueryParameterName.SEARCH;
 
-	const searchParameter = shouldUseQuery
+	const searchParameter = isSavedToUrl
 		? (searchParameters.get(searchParameterName) ?? "")
 		: "";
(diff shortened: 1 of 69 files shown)


apps/frontend/src/libs/hooks/use-search/use-search.hook.ts
```diff
@@ -1,20 +1,26 @@
 import { QueryParameterName } from "~/libs/enums/enums.js";
 import { useEffect, useSearchParams, useState } from "~/libs/hooks/hooks.js";
 
-const useSearch = (): {
+const useSearch = (
+	useQueryParameters: boolean = true,
+): {
```
perekotypole, line 6: I think it's better to pass an argument as an object, useSeatch(false) doesn't look very clear. 
And don't forget about B19 (now naming is the same as hooks)

Fjortis:
I am unsure about this change, perhaps it is better to make use of useState and useAppForm for modals instead of using useSearch. Do you think this is a better way of doing it?

apps/frontend/src/pages/access-management/libs/components/group-users-table/group-users-table.tsx
```diff
@@ -43,7 +43,7 @@ const GroupUsersTable = ({
 	setValue,
 }: Properties): JSX.Element => {
 	const dispatch = useAppDispatch();
-	const { onSearch, search } = useSearch();
+	const { onSearch, search } = useSearch({ shouldUseQuery: false });
```
liza-veis, line 46: let's rename it to isSavedToUrl
```suggestion
	const { onSearch, search } = useSearch({ isSavedToUrl: false });
```

apps/frontend/src/libs/hooks/use-search/use-search.hook.ts
```diff
@@ -1,20 +1,35 @@
 import { QueryParameterName } from "~/libs/enums/enums.js";
 import { useEffect, useSearchParams, useState } from "~/libs/hooks/hooks.js";
 
-const useSearch = (): {
+type SearchOptions = {
```
liza-veis, line 4:
```suggestion
type Parameters = {
```

liza-veis (changes requested):

what1s1ove (approved):
lgtm after Lisa's changes

liza-veis (approved):

PR #467, opened 2024-09-21
feat: authorize the request before starting background job for analytics gf-464

s1rserg:
* On start script makes a request with api key and user id to check if the credentials are valid.
 * In response to this request project name and id are returned.
 * In case the request was successful, the project name is used as a process name for pm2.
 
image


s1rserg:
Autorestart doesn't trigger for errors in collecting or sending statistics
image
image

packages/shared/src/libs/enums/api-path.enum.ts
```diff
@@ -1,6 +1,7 @@
 const APIPath = {
 	ACTIVITY_LOGS: "/activity-logs",
 	AUTH: "/auth",
+	AUTH_ANALYTICS: "/auth-analytics",
```
liza-veis, line 4:
```suggestion
	AUTH_ANALYTICS: "/auth/analytics",
```

what1s1ove (approved):

liza-veis (approved):

PR #468, opened 2024-09-21
feat: Add an activity chart to contributor items on project details page gf-420

GvoFor:
## Features
- created `ActivityChart` common component
- added `ActivityChart` to contributor cards on project details page, plotting contributor activity (commits number) during last month
- adjusted `getAll` activity logs method adding optional `projectId` query param, so you can get all logs relavent only to specific project

## Screenshots
image
image
image

### DIFF
--- a/apps/backend/src/modules/activity-logs/activity-log.repository.ts
+++ b/apps/backend/src/modules/activity-logs/activity-log.repository.ts
@@ -42,9 +42,10 @@ class ActivityLogRepository implements Repository {
 
 	public async findAll({
 		endDate,
+		projectId,
 		startDate,
 	}: ActivityLogQueryParameters): Promise<{ items: ActivityLogEntity[] }> {
-		const activityLogs = await this.activityLogModel
+		let query = this.activityLogModel
 			.query()
 			.withGraphFetched("[gitEmail.contributor, project, createdByUser]")
 			.modifyGraph("gitEmail.contributor", (builder) => {
@@ -52,6 +53,12 @@ class ActivityLogRepository implements Repository {
 			})
 			.whereBetween("activity_logs.date", [startDate, endDate])
 			.orderBy("date");
+
+		if (projectId) {
+			query = query.where("project_id", projectId);
+		}
+
+		const activityLogs = await query.execute();
 
 		return {
 			items: activityLogs.map((activityLog) =>
--- a/apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
+++ b/apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
@@ -1,0 +1,22 @@
+import { Line, LineChart } from "recharts";
+
+import { type ActivityChartData } from "./libs/types/types.js";
+
+type Properties = {
+	data: ActivityChartData;
+};
+
+const ActivityChart = ({ data }: Properties): JSX.Element => {
+	return (
+		<LineChart data={data} height={40} width={120}>
+			<Line
+				dataKey="commitsNumber"
+				dot={false}
+				stroke="var(--color-brand-primary)"
+				strokeWidth={3}
+			/>
+		</LineChart>
+	);
+};
+
+export { ActivityChart };
--- a/apps/frontend/src/pages/project/libs/components/contributor-card/contributor-card.tsx
+++ b/apps/frontend/src/pages/project/libs/components/contributor-card/contributor-card.tsx
@@ -9,17 +9,25 @@ import {
 	getStartOfDay,
 } from "~/libs/helpers/helpers.js";
 import { useCallback } from "~/libs/hooks/hooks.js";
+import { type ActivityLogGetAllItemAnalyticsResponseDto } from "~/modules/activity/activity.js";
 import { type ContributorGetAllItemResponseDto } from "~/pages/project/libs/types/types.js";
 
 import { ContributorMenu } from "../components.js";
 import styles from "./styles.module.css";
 
 type Properties = {
+	activityLog?:
+		| ActivityLogGetAllItemAnalyticsResponseDto["commitsNumber"]
+		| undefined;
 	contributor: ContributorGetAllItemResponseDto;
 	onEdit: (contributorId: number) => void;
 };
 
-const ContributorCard = ({ contributor, onEdit }: Properties): JSX.Element => {
+const ContributorCard = ({
+	activityLog,
+	contributor,
+	onEdit,
+}: Properties): JSX.Element => {
 	const currentDate = getStartOfDay(new Date());
 	const lastActivityDate = contributor.lastActivityDate
 		? getStartOfDay(new Date(contributor.lastActivityDate))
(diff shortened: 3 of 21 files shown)


apps/frontend/src/pages/project/project.tsx
```diff
@@ -83,6 +85,20 @@ const Project = (): JSX.Element => {
...
+			const startDate = subtractDays(
+				todayDate,
+				ANALYTICS_DATE_MAX_RANGE,
+			).toISOString();
+			void dispatch(
+				projectActions.loadAllContributorsActivityByProjectId({
+					endDate,
+					projectId,
+					startDate,
+				}),
```
GvoFor, line 100: Maybe I should dispatch this action inside `projectActions.loadAllContributorsByProjectId` action, or even rid of this action and place its code into mensioned action, cause in case when loading contributors we also want their activity being loaded? or it's better to keep them separate.

what1s1ove (approved):

apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
```diff
@@ -0,0 +1,22 @@
...
+	data: ActivityChartData;
+};
+
+const ActivityChart = ({ data }: Properties): JSX.Element => {
+	return (
+		<LineChart data={data} height={40} width={120}>
+			<Line
+				dataKey="commitsNumber"
+				dot={false}
+				stroke="var(--color-brand-primary)"
```
liza-veis, line 15: can't we pass this using css className

apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+type Properties = {
+	data: ActivityChartData;
+};
+
+const ActivityChart = ({ data }: Properties): JSX.Element => {
+	return (
+		<LineChart data={data} height={40} width={120}>
+			<Line
+				dataKey="commitsNumber"
```
liza-veis, line 13: I think we shouldn't be attached to commitsNumber inside this component, we should be able to reuse it not only for commits

apps/frontend/src/pages/project/libs/components/contributor-card/contributor-card.tsx
```diff
@@ -37,6 +48,12 @@ const ContributorCard = ({ contributor, onEdit }: Properties): JSX.Element => {
 
 	const hasActivityIndicator = lastUpdateLabel !== null && colorStatus !== null;
 
+	const hasActivityData = Boolean(activityLog);
+	const activityData =
```
liza-veis, line 52: let's use useMemo for this

apps/frontend/src/pages/project/libs/components/contributors-list/contributors-list.tsx
```diff
@@ -37,6 +40,12 @@ const ContributorsList = ({
 					{contributors.map((contributor) => (
 						<li key={contributor.id}>
 							<ContributorCard
+								activityLog={
+									activityLogs.find(
+										(activityLog) =>
+											Number(activityLog.contributorId) === contributor.id,
+									)?.commitsNumber
```
liza-veis, line 47: maybe we can group it by contributor id inside useMemo outside of jsx and then access it here by contributor id?

liza-veis (changes requested):

apps/frontend/src/pages/project/project.tsx
liza-veis: I think we need to dispatch it inside loadAllContributorsByProjectId, it's ok to keep it as a separate action

apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
GvoFor: I can do it this way
image
image
But it seems to me, that this violates B10 QC

apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
GvoFor: Okay. Should I assume that we should be able to use it not only for activity charts, and rename it to `Chart` as well?

GvoFor:
I also have question regarding contributor with no activity: in case contributor has no activity yet, then his activity list won't be null/undefined, but will be an array filled with zeros only. So the question is: should I render this activity (it will looks like a strait line in the middle of chart), or should I add tiny check for this case and do not render anything in this case?

apps/frontend/src/pages/project/libs/components/contributors-list/contributors-list.tsx
```diff
@@ -41,6 +47,20 @@ const ContributorsList = ({
...
+	const contributorActivitiesMap = useMemo(() => {
+		const activitiesMap = new Map<number, ContributorActivity | undefined>();
+
+		for (const contributor of contributors) {
+			const contributorActivity = activityLogs.find(
+				(activityLog) => Number(activityLog.contributorId) === contributor.id,
+			)?.commitsNumber;
+
+			activitiesMap.set(contributor.id, contributorActivity);
+		}
```
liza-veis, line 59: we don't need the logic with matching contributors here. We can just go through activityLogs and set activityLog.contributorId as the key. activityLogs will already contain only needed contributor ids and we don't want to search through the array for each contributor if we can avoid this.

liza-veis (changes requested):

liza-veis:
> I also have question regarding contributor with no activity: in case contributor has no activity yet, then his activity list won't be null/undefined, but will be an array filled with zeros only. So the question is: should I render this activity (it will looks like a strait line in the middle of chart), or should I add tiny check for this case and do not render anything in this case?

Let's show an empty line. If the contributor was created for the project, this means that he had at least one commit in the project, but maybe with a different timestamp, and in this case I think it's more logical to show this graph

apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
liza-veis: yes, this will be better I think

apps/frontend/src/pages/project/libs/components/contributors-list/contributors-list.tsx
GvoFor: Sure

apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
liza-veis: To override library styles it's ok

apps/backend/src/modules/activity-logs/activity-log.repository.ts
```diff
@@ -53,6 +54,12 @@ class ActivityLogRepository implements Repository {
 			.whereBetween("activity_logs.date", [startDate, endDate])
 			.orderBy("date");
 
+		if (projectId) {
+			query = query.where("project_id", projectId);
+		}
```
perekotypole, line 59: It has to work like this, doesn't it?
```
		const query = this.activityLogModel
			.query()
			.withGraphFetched("[gitEmail.contributor, project, createdByUser]")
			.modifyGraph("gitEmail.contributor", (builder) => {
				builder.select("id", "name");
			})
			.whereBetween("activity_logs.date", [startDate, endDate])
			.orderBy("date");

		if (projectId) {
			query.where("project_id", projectId);
		}
```

GvoFor: Right. Thanks. Changed

apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
GvoFor: Done

apps/frontend/src/libs/components/activity-chart/activity-chart.tsx
GvoFor: Done

PR #470, opened 2024-09-21
feat: add projects select to analytics page gf-348

AnutkaGn:
Add project selector to the analytics page
When the project is selected only contributors from this project be shown.
image


AnutkaGn:
Can I make a separate request to get all the projects, because it is a bit problematic to make it an optional query? Because I need to extract all the projects for the selector

apps/backend/src/modules/projects/project.controller.ts
```diff
@@ -227,17 +227,11 @@ class ProjectController extends BaseController {
 	 */
 	private async findAll(
 		options: APIHandlerOptions<{
-			query: ProjectGetAllRequestDto;
+			query: object | ProjectGetAllRequestDto;
```
what1s1ove, line 230:
```suggestion
			query: ProjectGetAllRequestDto;
```
specify type

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -120,7 +120,7 @@ class ProjectService implements Service {
 	}
 
 	public async findAll(
-		parameters: ProjectGetAllRequestDto,
+		parameters: object | ProjectGetAllRequestDto,
```
what1s1ove, line 123: same

liza-veis:
Pay attention to the changes in this PR https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/468, so that they are not conflicting

apps/frontend/src/pages/analytics/libs/helpers/get-project-options.helper.ts
sandrvvu, line 1: Place helpers in their separate folders

apps/frontend/src/modules/projects/slices/project.slice.ts
```diff
@@ -66,10 +66,10 @@ const { actions, name, reducer } = createSlice({
 		});
 		builder.addCase(loadAll.fulfilled, (state, action) => {
 			const { items, totalItems } = action.payload;
-			const { page } = action.meta.arg;
-
-			state.projects =
+			const page = action.meta.arg?.page;
```
what1s1ove, line 69:
```suggestion
			const { page } = action.meta.arg;
```

what1s1ove (approved):

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -76,6 +88,18 @@ const Analytics = (): JSX.Element => {
 			<h1 className={styles["title"]}>Analytics</h1>
 			<section>
 				<form className={styles["filters-form"]} onSubmit={handleFormSubmit}>
+					<div className={styles["wraper-selector"]}>
```
GvoFor, line 91:
```suggestion
					<div className={styles["select-wrapper"]}>
```

apps/backend/src/modules/projects/project.controller.ts
```diff
@@ -227,17 +227,11 @@ class ProjectController extends BaseController {
 	 */
 	private async findAll(
 		options: APIHandlerOptions<{
-			query: ProjectGetAllRequestDto;
+			query: null | ProjectGetAllRequestDto;
```
GvoFor, line 230: Why not `undefined`?
```suggestion
			query: undefined | ProjectGetAllRequestDto;
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -22,26 +24,34 @@ const Analytics = (): JSX.Element => {
 	const { activityLogs, dataStatus } = useAppSelector(
 		({ activityLogs }) => activityLogs,
 	);
+	const { projects } = useAppSelector(({ projects }) => projects);
+
+	useEffect(() => {
+		void dispatch(projectActions.loadAll());
```
liza-veis, line 30: We need to load these projects separately, specifically for analytics module, so it should be in analytics slice. Also on the backend side we need to split service method for pagination and without and the same for repository. Take a look at how it's done here #491

liza-veis (changes requested):

apps/frontend/src/pages/analytics/analytics.tsx
AnutkaGn: Did I understand correctly that I can make a separate request on the backend in order to get all the projects (without pagination)? And use it in the analytics slice? I thought it would be better that way, I asked, but I didn't get an answer then

liza-veis: Yes, let's make it as a separate request

liza-veis: probably missed your question in the chat

AnutkaGn: Everything is fine, I hope it is correct now

apps/backend/src/modules/projects/project.controller.ts
```diff
@@ -242,6 +254,33 @@ class ProjectController extends BaseController {
...
+	 *               type: array
+	 *               items:
+	 *                 $ref: "#/components/schemas/Project"
+	 */
+	private async findAllProjects(): Promise<APIHandlerResponse> {
+		const projects = await this.projectService.findAllWithoutPagination();
+
+		return {
+			payload: projects,
+			status: HTTPCode.OK,
```
GvoFor, line 280:
```suggestion
		return {
			payload: await this.projectService.findAllWithoutPagination(),
			status: HTTPCode.OK,
```

apps/backend/src/modules/projects/project.service.ts
```diff
@@ -135,6 +135,18 @@ class ProjectService implements Service {
...
+	public async findAllWithoutPagination(): Promise<
+		ProjectGetAllItemResponseDto[]
+	> {
+		const projects = await this.projectRepository.findAllWithoutPagination();
+
+		return projects.map((project) => {
+			const { id, lastActivityDate, name } = project.toObject();
+
+			return { id, lastActivityDate, name };
+		});
```
GvoFor, line 147: Can't we do it like
```suggestion
		return projects.map((project) => project.toObject());
```

apps/frontend/src/modules/activity/libs/types/types.ts
```diff
@@ -4,4 +4,5 @@ export {
 	type ActivityLogGetAllItemResponseDto,
 	type ActivityLogGetAllResponseDto,
 	type ActivityLogQueryParameters,
+	type ProjectGetAllItemResponseDto,
```
GvoFor, line 7: I'm not sure, but maybe it'll be better to import this type from `~/modules/projects/projects.ts` in places where you need it, and not reexport this here?

liza-veis (approved):

PR #478, opened 2024-09-23
fix: user profile not deleted without "Manage User Access" permission gf-449

Vitaliistf:
Added logic to handle user self-accessing activity to the check user permissions hook.


apps/backend/src/libs/hooks/check-user-permissions.hook.ts
```diff
@@ -16,6 +16,14 @@ const checkUserPermissions = (routePermissions: string[]): APIPreHandler => {
 			});
 		}
 
+		const resourceId = (params as { id?: string }).id;
+
+		if (url.includes(APIPath.USERS) && Number(resourceId) === user.id) {
```
perekotypole, line 21: How about adding a variable like `isCurrentUserRoute`,  `hasFreeAccess` or something for the condition? Just to keep semantic

Vitaliistf: Thanks, nice notice. I've changed this, however decided on a little different names.

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
```diff
@@ -16,6 +16,16 @@ const checkUserPermissions = (routePermissions: string[]): APIPreHandler => {
...
 
+		const resourceId = (params as { id?: string }).id;
+		const isUsersRoute = url.includes(APIPath.USERS);
+		const isCurrentUserRoute = isUsersRoute && Number(resourceId) === user.id;
+
+		if (isCurrentUserRoute) {
+			done();
+
+			return;
+		}
```
liza-veis, line 27: This logic shouldn't be in checkUserPermissions hook. This hook is for common permission case and should now nothing about specific routes and their logic. Instead let's use DELETE DELETE /authenticated-user and PUT /authenticated-user in auth controller for current user, and move the logic related to this routes to auth module

liza-veis (changes requested):

apps/backend/src/libs/hooks/check-user-permissions.hook.ts
liza-veis: yes

liza-veis (approved):

what1s1ove (approved):

PR #488, opened 2024-09-23
feat: the earliest date availible to be choosen on the date range on the analytics page should be 1 year from today gf-482

Vitaliistf:
Limited earliest date for date input on analytics page. It is 365 days from today now. Also set minimum view to year, so now user won't see decades and centuries view.
Screenshot 2024-09-23 at 16 07 49


liza-veis:
Maybe we can disable this month's view at all? I think users will only use 3 nearest months and in case of one year, it is not that much to scroll using the usual view
Screenshot 2024-09-23 at 17 02 37

Vitaliistf:
> Maybe we can disable this month's view at all? I think users will only use 3 nearest months and in case of one year, it is not that much to scroll using the usual view Screenshot 2024-09-23 at 17 02 37

Sure. Now there is an only view, the month view.

apps/frontend/src/pages/analytics/libs/constants/analytics-lookback-days.constant.ts
```diff
@@ -0,0 +1,3 @@
+const ANALYTICS_LOOKBACK_DAYS = 365;
```
what1s1ove, line 1:
```suggestion
const ANALYTICS_LOOKBACK_DAYS_COUNT = 365;
```

liza-veis (approved):

PR #490, opened 2024-09-23
feat: add infinite scroll and pagination to notificationspopover gf-458

Fjortis:
Implemented infinite scroll for the NotificationsPopover component.

Added pagination handling to load notifications incrementally.

https://github.com/user-attachments/assets/79a87c88-0840-4248-b055-f34eb0fab971


s1rserg:
Try `feat: add infinite scroll and pagination to notificationspopover gf-458`

Fjortis:
Do we want the notification title to be sticky? 

https://github.com/user-attachments/assets/baca22e6-86f4-43b8-b9da-82623fb5101a

liza-veis:
> Do we want the notification title to be sticky?

yes, let's make it sticky

apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
```diff
@@ -18,26 +21,46 @@ type Properties = {
 	onClose: () => void;
 };
 
+const PAGE_SIZE = 10;
```
what1s1ove, line 24: should be in a separate file

apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
```diff
@@ -18,26 +21,46 @@ type Properties = {
...
+	const { hasNextPage, onNextPage, onPageReset } = useInfiniteScroll({
+		currentItemsCount: notifications.length,
+		onLoading: handleLoadNotifications,
+		pageSize: PAGE_SIZE,
+		totalItemsCount: notificationsTotalCount,
+	});
+
+	const { reference: sentinelReference } =
+		useIntersectionObserver<HTMLDivElement>({
+			isDisabled: !hasNextPage || dataStatus === "pending",
```
what1s1ove, line 52: use an enum here

apps/backend/src/modules/notifications/notification.controller.ts
```diff
@@ -53,7 +59,22 @@ class NotificationController extends BaseController {
...
+	 *           type: integer
+	 *           minimum: 1
+	 *           default: 1
+	 *         description: The page number to retrieve
+	 *       - in: query
+	 *         name: pageSize
+	 *         schema:
+	 *           type: integer
+	 *           minimum: 1
+	 *           default: 10
```
GvoFor, line 76: Here you mentioned default values for page and pageSize, but there is no code for this ( frontend doesn't counted :0 )
So what would happen if we try to send get request without any query params?

Fjortis: I updated the Swagger documentation to match the backend code. Initially, I had set default values for page and `pageSize `in the backend, but I realized that the frontend already manages those values in the `NotificationsPopover` component and `useInfiniteScroll `hook. So, there's no need for default handling in the backend.

apps/frontend/src/libs/components/header/libs/components/notifications-popover/libs/constants/default-page-size.constant.ts
```diff
@@ -0,0 +1,3 @@
+const PAGE_SIZE = 10;
```
liza-veis, line 1: and file name should be the same
```suggestion
const NOTIFICATIONS_PAGE_SIZE = 10;
```

apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
Fjortis: This has since been adjust, ty for the review

apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
Fjortis: Now is using enum like the other places in code

apps/frontend/src/modules/notifications/slices/notification.slice.ts
```diff
@@ -22,11 +25,17 @@ const { actions, name, reducer } = createSlice({
...
+			const { page } = action.meta.arg;
+
+			state.notifications =
+				page === FIRST_PAGE ? items : [...state.notifications, ...items];
+			state.notificationsTotalCount = totalItems;
 			state.dataStatus = DataStatus.FULFILLED;
 		});
 		builder.addCase(loadAll.rejected, (state) => {
 			state.notifications = [];
+			state.notificationsTotalCount = 1;
```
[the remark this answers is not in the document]
Fjortis: Yep, that is a mistake.

apps/frontend/src/libs/components/header/libs/components/notifications-popover/styles.module.css
```diff
@@ -20,6 +25,10 @@
 	padding: 4px;
 }
 
+.sentinel {
+	height: 1px;
+}
```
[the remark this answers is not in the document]
Fjortis: Oh, that is leftover code from before 😳
I was testing the functionality and styling for the scrolling. I think it is all good now

apps/backend/src/modules/notifications/notification.controller.ts
```diff
@@ -66,19 +83,28 @@ class NotificationController extends BaseController {
...
+	}>): Promise<APIHandlerResponse> {
+		const { page, pageSize } = query;
 		const typedUser = user as { id: number };
 
 		return {
-			payload: await this.notificationService.findAll(typedUser.id),
+			payload: await this.notificationService.findAll({
+				page,
+				pageSize,
+				userId: typedUser.id,
```
liza-veis, line 104: let's avoid variables just for typescript, where we can. Also we have a specific type for user in request
```suggestion
				userId: (user as UserAuthResponseDto).id,
```

apps/frontend/src/modules/notifications/slices/notification.slice.ts
```diff
@@ -22,11 +25,17 @@ const { actions, name, reducer } = createSlice({
...
+			const { page } = action.meta.arg;
+
+			state.notifications =
+				page === FIRST_PAGE ? items : [...state.notifications, ...items];
+			state.notificationsTotalCount = totalItems;
 			state.dataStatus = DataStatus.FULFILLED;
 		});
 		builder.addCase(loadAll.rejected, (state) => {
 			state.notifications = [];
+			state.notificationsTotalCount = initialState.notificationsTotalCount;
```
GvoFor, line 38: The idea to use `initialState` for "reseting" is really good, but we don't use it anywhere, so maybe to keep it consistent  let's reset it like
```suggestion
			state.notificationsTotalCount = 0;
```

apps/backend/src/modules/notifications/notification.controller.ts
Fjortis: I resolved the conversation, but https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/490#discussion_r1771722196, recommended for this way instead.

apps/frontend/src/modules/notifications/slices/notification.slice.ts
Fjortis: I resolved the conversation, but https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/490#discussion_r1771722196, recommended for this way instead.

liza-veis (approved):

apps/frontend/src/modules/notifications/slices/notification.slice.ts
liza-veis: If we need a specific value it's better to use the initial state or any other constant we have as default. Here we don't have a constant, so initialValue is ok

PR #491, opened 2024-09-23
feat: add pagination to the contributors page gf-484

s1rserg:
* Added pagination to the contributors page.
* Contributors are sorted by created at, recent contributors should be at the top of the list.
* When contributors are merged, items total is updated.
image

### DIFF
--- a/apps/backend/src/modules/activity-logs/activity-log.service.ts
+++ b/apps/backend/src/modules/activity-logs/activity-log.service.ts
@@ -148,10 +148,7 @@ class ActivityLogService implements Service {
 
 		const allContributors = await (projectId
 			? this.contributorService.findAllByProjectId(Number(projectId))
-			: this.contributorService.findAll({
-					page: 1,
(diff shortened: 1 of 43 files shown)


apps/backend/src/modules/activity-logs/activity-log.service.ts
```diff
@@ -148,7 +148,10 @@ class ActivityLogService implements Service {
 
 		const allContributors = await (projectId
 			? this.contributorService.findAllByProjectId(Number(projectId))
-			: this.contributorService.findAll());
+			: this.contributorService.findAll({
+					page: 1,
+					pageSize: Number.MAX_SAFE_INTEGER,
+				}));
```
liza-veis, line 154: Here we either need to separate these two methods in the contributors controller and repository, to have one with pagination and one without or add pagination to the analytics table too. let's move on with the first option in this PR.

liza-veis (approved):

PR #498, opened 2024-09-23
feat: ability to mark notifications as read gf-460

sandrvvu:
Add is_read column to notifications table.
Add unread notifications count badge to the notifications icon.
When user opens notifications popup, unread notifications should have an effect we currently have on hover. Remove this hover effect on notifications (notifications are not clickable for now).
When a user opens popup we need to mark notifications as read, but the user will only see this when he opens the popup again. So on the initial render, a user should be able to see which notifications are unread.

https://github.com/user-attachments/assets/5de134d3-133c-47f0-8248-d2d7765c1d55


readme.md
```diff
@@ -168,11 +168,10 @@ erDiagram
 
    notifications {
       int id PK
+      boolean is_read
       dateTime created_at
       dateTime updated_at
       int receiver_user_id FK
-      enum status
```
what1s1ove, line 174: Having enums is much better than having flags. I would recommend moving it back
cc @liza-veis

liza-veis: agree, let's use status enum instead

apps/backend/src/db/migrations/20240924130637_add_status_to_notifications_table.ts
```diff
@@ -0,0 +1,24 @@
...
+const TABLE_NAME = "notifications";
+
+const ColumnName = {
+	STATUS: "status",
+} as const;
+
+function up(knex: Knex): Promise<void> {
+	return knex.schema.alterTable(TABLE_NAME, (table) => {
+		table
+			.enu(ColumnName.STATUS, ["unread", "read"])
```
GvoFor, line 12: It looks like `.enu` is the same as `.enum` but last one sounds better
```suggestion
			.enum(ColumnName.STATUS, ["unread", "read"])
```

apps/frontend/src/libs/components/header/header.tsx
```diff
@@ -26,11 +37,31 @@ const Header = (): JSX.Element => {
...
+
+	useEffect(() => {
+		void dispatch(notificationActions.loadAllUnread());
+	}, [dispatch]);
+
+	const markAllNotificationsAsRead = useCallback(() => {
+		for (const notification of unreadNotifications) {
+			void dispatch(notificationActions.markAsRead({ id: notification.id }));
+		}
+	}, [dispatch, unreadNotifications]);
```
liza-veis, line 52: This will be very inefficient. Instead let's use /read route to mark all notifications as read at once, and handle them as a bulk on backend side too.

liza-veis (changes requested):

what1s1ove (approved):
LGTM after Lisa's comments

apps/frontend/src/libs/hooks/use-handle-click-outside/use-handle-click-outside.hook.ts
```diff
@@ -3,9 +3,14 @@ import { useEffect } from "~/libs/hooks/hooks.js";
 const useHandleClickOutside = <T extends HTMLElement>(
 	reference: React.RefObject<T>,
 	onOutsideClick: () => void,
+	contentReference?: React.RefObject<T>,
 ): void => {
 	useEffect(() => {
 		const handleClickOutside = (event: MouseEvent): void => {
+			if (contentReference && !contentReference.current) {
+				return;
+			}
```
Vitaliistf, line 12: What is this contentReference prop for? Looks like it'll only stop function if there is no content ref passed. If so, then I don't think there is no logic we want to have here.

sandrvvu: While I was working with notification popover, I noticed that when popover is closed, and you click anywhere on the page, onPopoverClose function was always running. With these changes it only runs after the popover was open and you click outside this area.

Vitaliistf: Oh, strange. It's good that you've noticed this, thanks for explanation!

packages/shared/src/modules/notifications/libs/types/notification-bulk-mark-as-read-request-dto.type.ts
```diff
@@ -0,0 +1,5 @@
+type NotificationBulkMarkAsReadRequestDto = {
+	notificationIds: number[];
```
liza-veis, line 2: we don't need to pass anything to the backend, this request means to mark all notifications as read
```suggestion
```

apps/frontend/src/modules/notifications/slices/notification.slice.ts
```diff
@@ -38,6 +44,40 @@ const { actions, name, reducer } = createSlice({
...
+				notificationIds.includes(notification.id)
+					? { ...notification, isRead: true }
+					: notification,
+			);
+
+			state.unreadNotifications = state.unreadNotifications.filter(
+				(notification) => !notificationIds.includes(notification.id),
+			);
+
+			state.markAsReadStatus = DataStatus.FULFILLED;
```
liza-veis, line 76: Do we need this logic at all? we want user to see unread notification until he reopens the notification modal

liza-veis (changes requested):

apps/backend/src/modules/notifications/notification.repository.ts
```diff
@@ -82,6 +93,23 @@ class NotificationRepository implements Repository {
 		};
 	}
 
+	public async getUnreadCount(
```
liza-veis, line 96: We don't need to return all the unread notifications, we only need to return number of them from this method

liza-veis (changes requested):

liza-veis (approved):

PR #512, opened 2024-09-24
feat: redirect to analytics page on contributor item click gf-349

s1rserg:
* When a user clicks on a contributor item on the project details page, he is navigated to analytics page with the project preselected.
* Select query is copied to url query parameters.


packages/shared/src/libs/helpers/configure-query-string/configure-query-string.helper.ts
```diff
@@ -0,0 +1,22 @@
+import { EMPTY_LENGTH } from "../../constants/constants.js";
+
+const configureQueryString = <T extends Record<string, string>>(
```
liza-veis, line 3: we need to use url search params instead, also I wonder why we use a different approach here from pagination hook

liza-veis (changes requested):

what1s1ove (approved):
lgtm after Lisa's comments

packages/shared/src/libs/helpers/configure-query-string/configure-query-string.helper.ts
s1rserg: I made the helper use the same approach as pagination hook.

apps/frontend/src/libs/enums/query-parameter-name.enum.ts
```diff
@@ -1,6 +1,7 @@
 const QueryParameterName = {
 	PAGE: "page",
 	PAGE_SIZE: "pageSize",
+	PROJECT_SELECT: "projectId",
```
liza-veis, line 4:
```suggestion
	PROJECT_ID: "projectId",
```

apps/frontend/src/libs/hooks/use-query-parameters/use-query-parameters.hook.ts
```diff
@@ -1,16 +1,21 @@
...
 import { useEffect, useSearchParams, useState } from "~/libs/hooks/hooks.js";
+import { type ValueOf } from "~/libs/types/types.js";
 
 type Parameters = {
 	isSavedToUrl?: boolean;
+	queryParameterName?: ValueOf<typeof QueryParameterName>;
 };
 
-const useSearch = ({ isSavedToUrl = true }: Parameters = {}): {
+const useQueryParameters = ({
```
liza-veis, line 10: let's rename it because we have an option, not to save this to url, so previous name is confusing
```suggestion
const useSearchFilters = ({
```

apps/frontend/src/libs/hooks/use-query-parameters/use-query-parameters.hook.ts
```diff
@@ -1,16 +1,21 @@
...
+	isSavedToUrl = true,
+	queryParameterName,
+}: Parameters = {}): {
 	onSearch: (search: string) => void;
 	search: string;
 } => {
 	const [searchParameters, setSearchParameters] = useSearchParams();
-	const searchParameterName = QueryParameterName.SEARCH;
+	const searchParameterName = queryParameterName ?? QueryParameterName.SEARCH;
```
liza-veis, line 18: let's pass default value during properties destructuring
```suggestion
```
```
queryParameterName = QueryParameterName.SEARCH,
```

packages/shared/src/libs/helpers/configure-query-string/configure-query-string.helper.ts
liza-veis: looks a bit complicated, we have getUrl method inside base-http-api, so I believe this helper should have the same logic. If so, we also will need to update usePagination hook, and  base-http-api to use this helper instead of their implementation

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -35,18 +37,32 @@ const Analytics = (): JSX.Element => {
...
 	const { control, handleSubmit, isDirty } = useAppForm({
 		defaultValues: {
 			dateRange: [
 				subtractDays(todayDate, ANALYTICS_DATE_MAX_RANGE),
 				todayDate,
 			] as [Date, Date],
-			project: null,
+			project: searchParameters.get(QueryParameterName.PROJECT_SELECT)
+				? Number(searchParameters.get(QueryParameterName.PROJECT_SELECT))
+				: null,
```
liza-veis, line 54: let's simplify this logic
```tsx
const projectIdQueryParameter = searchParameters.get(QueryParameterName.PROJECT_ID);

//...
project: projectIdQueryParameter ? Number(projectIdQueryParameter) : null,
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -35,18 +37,32 @@ const Analytics = (): JSX.Element => {
...
 	});
 
 	const dateRangeValue = useFormWatch({ control, name: "dateRange" });
 	const projectValue = useFormWatch({ control, name: "project" });
+	const projectValueString =
+		projectValue === null ? "" : projectValue.toString();
+
+	useEffect(() => {
+		onSelect(projectValueString);
+	}, [onSelect, projectValueString]);
```
liza-veis, line 65:
```suggestion
	useEffect(() => {
		onSelect(projectValue ? String(projectValue) : null);
	}, [onSelect, projectValueString]);
```

apps/frontend/src/pages/analytics/analytics.tsx
```diff
@@ -65,13 +81,16 @@ const Analytics = (): JSX.Element => {
...
+	}, [dateRangeValue, projectValue, handleLoadLogs, projectValueString]);
 
 	const handleFormSubmit = useCallback(
 		(event_?: React.BaseSyntheticEvent): void => {
 			void handleSubmit((formData) => {
-				handleLoadLogs(formData.dateRange, formData.project);
+				handleLoadLogs(
+					formData.dateRange,
+					formData.project === null ? "" : formData.project.toString(),
+				);
```
liza-veis, line 93: let's pass it this way, we need to convert project id to string only in place where we pass it to query parameters (action). Also we don't need this check for null. We need to include this parameter conditionally inside the action 
```suggestion
			void handleSubmit(({ dateRange, projectId }) => {
				handleLoadLogs({
				  dateRange,
				  projectId
				});
```

liza-veis (changes requested):

apps/frontend/src/pages/analytics/analytics.tsx
s1rserg: There are some complications with validation schema if I change the types so action receives number instead of string. Hope this is ok
```
const handleLoadLogs = useCallback(
		([startDate, endDate]: [Date, Date], projectId?: null | number) => {
			const formattedStartDate = startDate.toISOString();
			const formattedEndDate = endDate.toISOString();

			void dispatch(
				activityLogActions.loadAll({
					endDate: formattedEndDate,
					projectId: projectId?.toString() ?? undefined,
					startDate: formattedStartDate,
				}),
			);
		},
		[dispatch],
	);
```

packages/shared/src/libs/helpers/configure-query-string/configure-query-string.helper.ts
s1rserg: I updated the helper to use the same logic as was in getUrl and updated base-http-api. But I cannot use this helper in usePagination because it doesn't create url with query, it just adds its parameters to existing. So it needs another helper which may be redundant or the configureQueryString will look more complicated.

liza-veis (approved):

PR #514, opened 2024-09-24
feat: upgrade the visibility of the project's options button on the project detail page gf-508

Vitaliistf:
Added optional prop for button class name. 
Changed appearance of project's options button to match contributors cards. (Agreed with Veronika)
Screenshot 2024-09-24 at 16 06 56


Fjortis:
I see that Figma uses text-primary color for the icon button, perhaps you can adjust that as well.

Vitaliistf:
> I see that Figma uses text-primary color for the icon button, perhaps you can adjust that as well. That will make it stand out much more

Yeah, you're right, thanks. Adjusted it.

what1s1ove (approved):

apps/frontend/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -6,6 +6,7 @@ import { ICON_SIZE } from "./libs/constants/constants.js";
 import styles from "./styles.module.css";
 
 type Properties = {
+	className?: string;
```
GvoFor, line 9: We shouldn't be able to styles common component like this. This is way it won't be COMMON component

GvoFor: You can extend `variant` property with additonal variant, I guess

liza-veis: Can't we use the already existing outlined variant here? Or adjust its styles slightly so that it fits both cases? It's better to keep as less variants as we can in the project ui overall and for common components too

Vitaliistf: I talked to Veronika, and we decided that we'll just use outlined variant then. Thanks!

apps/frontend/src/libs/components/menu/menu.tsx
```diff
@@ -3,13 +3,15 @@ import { IconButton, Popover } from "~/libs/components/components.js";
 import styles from "./styles.module.css";
 
 type Properties = {
+	buttonVariant?: "outlined" | "primary";
```
Vitaliistf, line 6: Maybe it would be right to move this type to the other file, as we're using it in two places now, but hope that it's not a big deal, as we have only two variants.

liza-veis: let's use `ComponentProps<typeof IconButton>` instead

liza-veis (approved):

PR #516, opened 2024-09-24
feat: Add project select to the contributors page gf-504

GvoFor:
## Features
Added project select above contributors table

## Screenshots
image

image

image

image

image


apps/frontend/src/modules/contributors/slices/actions.ts
```diff
@@ -27,6 +27,16 @@ const loadAll = createAsyncThunk<
 	return await contributorApi.getAll(query);
 });
 
+const loadAllByProjectId = createAsyncThunk<
+	ContributorGetAllResponseDto,
+	{ projectId: string },
```
perekotypole, line 32: Should `projectId` be a string, not a number?

GvoFor: Right. It's better to have this as number. I made a hude refactoring where change the type

apps/backend/src/modules/contributors/contributor.controller.ts
```diff
@@ -167,22 +181,22 @@ class ContributorController extends BaseController {
...
-				status: HTTPCode.OK,
-			};
-		}
+		const query = {
+			...(contributorName ? { contributorName } : {}),
+			...(hasHidden ? { hasHidden: Boolean(hasHidden) } : {}),
+			...(page ? { page: Number(page) } : {}),
+			...(pageSize ? { pageSize: Number(pageSize) } : {}),
+			...(projectId ? { projectId: Number(projectId) } : {}),
+		};
```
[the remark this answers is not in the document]
GvoFor: This will work only for `contributorName`, `hasHidden` and `projectId` since I've defined them as `field?: type | undefined`, and won't work for `pageSize` and `page` cause for them I used `Partial<..>` so their type is `field?: type` which means, that either field present in the object and have type `type` or it is absent at all -- `undefined` cannot be assigned to `page`/`pageSize`

So I can rewrite it only this way
```ts
const query = {
	contributorName,
	hasHidden: hasHidden ? Boolean(hasHidden) : undefined,
	projectId: projectId ? Number(projectId) : undefined,
	...(page ? { page: Number(page) } : {}),
	...(pageSize ? { pageSize: Number(pageSize) } : {}),
};
```
but this way it's a bit inconsistant (there are both `projectId: projectId ? Number(projectId) : undefined,` and `...(page ? { page: Number(page) } : {}),` in this object construction) and I'm not sure what looks more readable: this way or that current one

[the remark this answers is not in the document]
liza-veis: let's leave it as it is, we don't want to pass unneeded query parameters

apps/frontend/src/pages/contributors/contributors.tsx
```diff
@@ -47,15 +54,25 @@ const Contributors = (): JSX.Element => {
...
+		void dispatch(activityActions.loadAllProjects());
+	}, [dispatch]);
+
 	const { onPageChange, onPageSizeChange, page, pageSize } = usePagination({
 		queryParameterPrefix: "contributor",
 		totalItemsCount: totalCount,
 	});
 
-	useEffect(() => {
+	const loadContributors = useCallback(() => {
```
what1s1ove, line 68: a10

apps/backend/src/modules/contributors/contributor.controller.ts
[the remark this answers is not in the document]
GvoFor: Oh, wait. Sorry for confusing. This code is responsible not for sending query (frontend), but for retrieving it (on backend).
Here is the code where we choose what to send (`apps/frontend/src/modules/contributors/contributors-api.ts`)
image

So what about now. Should I change this (_backend_) code?

P.s. Noticed, that I forget to pass contributorName to queryToSend. I'll fix it

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -57,57 +58,16 @@ class ContributorRepository implements Repository {
...
 		page,
 		pageSize,
-	}: { hasHidden?: boolean } & ContributorGetAllQueryParameters): Promise<
+		projectId,
+	}: ContributorGetAllRequestDto): Promise<
 		PaginationResponseDto<ContributorEntity>
 	> {
 		const query = this.contributorModel
 			.query()
 			.orderBy("createdAt", SortType.DESCENDING)
```
Vitaliistf, line 70: As you left only one method for retrieving contributors, I think we might also need to pass option for sorting or smth. When we're retrieving all contributors, they were ordered by the time they were created, and when we're retrieving them by projectId, they were ordered by the time of their last activity. (The last thing was merged to main yesterday) Or maybe we can sort all contributors by their last activity date.

Vitaliistf: Oh, yeah, I see in the chat that you've noticed it already😅

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -135,50 +93,29 @@ class ContributorRepository implements Repository {
...
-			.leftJoin("git_emails", "contributors.id", "git_emails.contributor_id")
-			.leftJoin("activity_logs", "git_emails.id", "activity_logs.git_email_id")
-			.leftJoin("projects", "activity_logs.project_id", "projects.id")
-			.groupBy("contributors.id")
-			.withGraphFetched("gitEmails");
-
-		if (!hasHidden) {
-			query.whereNull("contributors.hiddenAt");
+		if (projectId) {
+			query.havingRaw("?? = ANY(ARRAY_AGG(projects.id))", projectId);
```
Vitaliistf, line 97: Is there a reason we can't just use `where` here?

apps/backend/src/modules/contributors/contributor.repository.ts
GvoFor: Yeah. I'm also thinking of sortBy option, cause in the table we need to sort by created_at (so rows don't swap while script updates analytics), and in the project details page we need to sort by last_activity_date.

Of course I may choose sorting depanding on `projectId` (either it was passed or not), but this is not a good solution, as in the table I also retrieve contributors by project id, and the issue will remain

>maybe we can sort all contributors by their last activity date.

Also thinking about this but not sure. I'll clarify this.

Thanks for suggession)

apps/backend/src/modules/contributors/contributor.repository.ts
GvoFor: No we can't. Check [this](https://github.com/BinaryStudioAcademy/bsa-2024-gitfit/pull/401#discussion_r1774008896) comment, where I've described the reason

kolibri753: Thanks for always providing the links :cat2:

apps/backend/src/modules/contributors/contributor.repository.ts
GvoFor: I have added `orderBy` to the query

apps/backend/src/modules/contributors/contributor.controller.ts
```diff
@@ -167,22 +188,23 @@ class ContributorController extends BaseController {
...
-			return {
-				payload: await this.contributorService.findAllByProjectId({
-					projectId,
-				}),
-				status: HTTPCode.OK,
-			};
-		}
+		const query = {
+			...(contributorName ? { contributorName } : {}),
+			...(hasHidden ? { hasHidden: Boolean(hasHidden) } : {}),
```
GvoFor, line 199: In fact this line contains a bug: if we get `hasHidden` as `"false"` then this convertion will make it equal to `true`.
I didn't add add solution cause it looks a bit weird (it looks like `...(hasHidden ? { hasHidden: JSON.parse(hasHidden as unknown as string) as boolean } : {}),`). Also if we don't want to include hidden contributors we just shouldn't include `hasHidden` param to our query, so that's why it works.

So should I add this weird convertion just to for case, when we explicitly pass `hasHidden: false` to the query?

apps/backend/src/modules/contributors/contributor.repository.ts
```diff
@@ -136,50 +94,41 @@ class ContributorRepository implements Repository {
...
+				break;
+			}
 
-		if (!hasHidden) {
-			query.whereNull("contributors.hiddenAt");
+			case ContributorOrderBy.LAST_ACTIVITY_DATE: {
+				query.orderBy("last_activity_date", SortType.DESCENDING);
+				break;
+			}
 		}
```
GvoFor, line 111: Not sure about this switch. In theory I could just write something like `query.orderBy(orderBy, SortType.DESCENDING);`
But if user pass `orderBy` param with value which is not belongs to ContributorOrderBy enum, then he'll get error with DB query details, which I think is bad

apps/backend/src/modules/contributors/contributor.controller.ts
GvoFor: That was so easy... Thanks

packages/shared/src/modules/contributors/libs/types/contributor-get-all-request-dto.type.ts
```diff
@@ -0,0 +1,14 @@
+import {
+	type PaginationQueryParameters,
+	type ValueOf,
+} from "../../../../libs/types/types.js";
+import { type ContributorOrderBy } from "../enums/contributor-order-by.enum.js";
+
+type ContributorGetAllRequestDto = {
```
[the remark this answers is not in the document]
GvoFor: You right

packages/shared/src/modules/contributors/libs/validation-schemas/contributor-get-all.validation-schema.ts
```diff
@@ -0,0 +1,19 @@
+import { z } from "zod";
+
+import {
+	ContributorOrderByKey,
+	ContributorValidationMessage,
+} from "../enums/enums.js";
+import { type ContributorGetAllQueryParameters } from "../types/types.js";
+
+const contributorGetAll: z.ZodType<ContributorGetAllQueryParameters> = z.object(
+	{
```
liza-veis, line 10: Add the rest of the query params too

apps/backend/src/modules/contributors/contributor.controller.ts
```diff
@@ -167,22 +192,23 @@ class ContributorController extends BaseController {
...
-			return {
-				payload: await this.contributorService.findAllByProjectId({
-					projectId,
-				}),
-				status: HTTPCode.OK,
-			};
-		}
+		const query = {
+			...(contributorName ? { contributorName } : {}),
+			...(hasHidden ? { hasHidden: (hasHidden as unknown) === "true" } : {}),
```
liza-veis, line 203: It should be a string, not unknown. Validation schema should only allow 'true' or 'false'

liza-veis (changes requested):

apps/backend/src/modules/contributors/contributor.controller.ts
liza-veis: And probably we need a mapper or helper to  parse boolean value

packages/shared/src/modules/contributors/libs/validation-schemas/contributor-get-all.validation-schema.ts
GvoFor: There is a problem with this. `ContributorGetAllQueryParameters` type consist of fields of different non-string types, but our incoming `query` (which we get in controller) have all fields of type string and this leads to error

GvoFor: So It means that all of the `ContributorGetAllQueryParameters` fields should be strings, which is probably not ok

apps/backend/src/modules/contributors/contributor.controller.ts
GvoFor: I tried `as string`, but typescript doesn't like it, as he think that `hasHidden` has boolean type, and I'm getting lint error that I can't convert boolean to string

packages/shared/src/modules/contributors/libs/validation-schemas/contributor-get-all.validation-schema.ts
GvoFor: image
image

apps/backend/src/modules/contributors/contributor.controller.ts
GvoFor: image

liza-veis: Ok, let's leave it with unknown then

packages/shared/src/modules/contributors/libs/validation-schemas/contributor-get-all.validation-schema.ts
liza-veis: Isn't there type for string which is number in zod?

GvoFor: No, there isn't. At least I haven't found it so far ((

Here is the only way to validate query params I found so far
image
and I have to remove generic type parameter from ZodType, cause otherwise there is mismatch with types.

Maybe I should separate `ContributorGetAllQueryParameters` into two types -- one with string types for ZodType, controller and api, and second one with actual types (booleans, numbers etc) for service/repository and some files on frontend. But this idea doen't sound good to me.

liza-veis: Would be better to separate types, but we have no so much time left, so let's leave it as it is

what1s1ove (approved):
lgtm after Lisa's comments

apps/backend/src/modules/project-groups/project-group.repository.ts
```diff
@@ -129,14 +129,16 @@ class ProjectGroupRepository implements Repository {
 			.withGraphJoined("[projects, users, permissions]")
 			.where("users.id", userId);
 
-		return results.map(({ projects, ...projectGroup }) => {
-			const [project] = projects;
+		return results
+			.filter(({ projects }) => projects.length)
```
GvoFor, line 133: This check for cases when user deletes project, which has some project groups. After such a deletion in the DB project groups are remained in `project_groups` table, but they relation with `projects` are lost (corresponding record in `projects_to_project_groups` are deleted). So without this `.filter`, `project` in `const [project] = projects;` expression is undefined which leads to `Reading property 'id' of undefined` error.

It would be better to remove this filter after we fix mentioned issue.

packages/shared/src/modules/contributors/libs/enums/contributor-order-by-key.enum.ts
```diff
@@ -0,0 +1,6 @@
+const ContributorOrderByKey = {
+	CREATED_AT: "created_at",
+	LAST_ACTIVITY_DATE: "last_activity_date",
```
Vitaliistf, line 3: So as we agreed, maybe we should use camelCase here? (I think that this might be fixed in ticket regarding quality criteria, but if you'll fix anything else, you might to fix this also).

liza-veis: @Vitaliistf can we fix this in your PR please? we need this pr to be merged to fix deployment

liza-veis (approved):

packages/shared/src/modules/contributors/libs/enums/contributor-order-by-key.enum.ts
Vitaliistf: Sure

PR #518, opened 2024-09-24
feat: decrease default date range to one week gf-503

sandrvvu:
The default analytics table date range is 1 week
image


GvoFor:
I used this constant in `loadAllContributorsByProjectId` action of `project.slice`, which is used for contributor activity chart, which should represent contributor activity for last month.
So could you please create new constant with value 30 and update loadAllContributorsByProjectId

apps/frontend/src/modules/projects/libs/constants/contribution-analytics-date-max-range.constant.ts
```diff
@@ -0,0 +1,3 @@
+const CONTRIBUTION_ANALYTICS_DATE_MAX_RANGE = 30;
```
liza-veis, line 1: this should stay as ANALYTICS_DATE_MAX_RANGE, and for another constant - it's not a max range, it's just an initial range, so let's rename it to ANALYTICS_DEFAULT_DATE_RANGE. So the max range should stay the same - 30 days, but the default preselected value should be 7 days
```suggestion
const ANALYTICS_DATE_MAX_RANGE = 30;
```

liza-veis (changes requested):

liza-veis (approved):

PR #520, opened 2024-09-24
fix: adjusted loader and placeholder to render properly when loading gf-515

Fjortis:
- Adjusted the loader to show when loading. 

- Adjusted placeholder to only show when no items and not loading. 

Notifications with throttle

https://github.com/user-attachments/assets/f3411431-13fb-4f56-a5fc-91a5c28a2791

No notifications

https://github.com/user-attachments/assets/4d28f34f-08b5-4af2-b738-45beae2371a9

### DIFF
--- a/apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
+++ b/apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
@@ -60,7 +60,7 @@ const NotificationsPopover = ({
 
 	const hasNotifications = notifications.length !== EMPTY_LENGTH;
 	const isLoadingMore = dataStatus === DataStatus.PENDING;
-	const shouldShowPlaceholder = !hasNotifications && !isLoadingMore;
+	const isPlaceholderShown = !hasNotifications && !isLoadingMore;
 
 	return (
 		<Popover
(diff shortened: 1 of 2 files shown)


apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
```diff
@@ -59,27 +59,29 @@ const NotificationsPopover = ({
...
-						) : (
+						{hasNotifications
+							? notifications.map((notification) => (
+									<NotificationItem
+										key={notification.id}
+										message={notification.payload}
+										timestamp={formatRelativeTime(notification.createdAt)}
+									/>
+								))
+							: null}
```
GvoFor, line 79: To render nothing we are using this approach
```suggestion
							: <></>}
```

apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
```diff
@@ -59,27 +59,29 @@ const NotificationsPopover = ({
...
+						{shouldShowPlaceholder ? (
 							<p className={styles["empty-placeholder"]}>
 								There is nothing yet.
 							</p>
-						)}
+						) : null}
```
GvoFor, line 84: same

apps/frontend/src/libs/components/header/libs/components/notifications-popover/notifications-popover.tsx
```diff
@@ -59,27 +59,29 @@ const NotificationsPopover = ({
...
-							notifications.map((notification) => (
-								<NotificationItem
-									key={notification.id}
-									message={notification.payload}
-									timestamp={formatRelativeTime(notification.createdAt)}
-								/>
-							))
-						) : (
+						{hasNotifications
+							? notifications.map((notification) => (
```
liza-veis, line 72: hasNotifications is a boolean property, so nothing will be rendered in case of false and we can use && operator. And the same for the condition below
```suggestion
						{hasNotifications && notifications.map((notification) => (
```

liza-veis (changes requested):

liza-veis (approved):

PR #521, opened 2024-09-24
fix: inconsistent data display on analytics table due to date format issue gf-510

sandrvvu:
+ Added getEndOfDay helper.
+ Adjusted date format (omitting time) on frontend and backend.

First render earlier:
image
Now:
image

### DIFF
--- a/packages/shared/src/libs/helpers/get-start-of-day/get-start-of-day.helper.ts
+++ b/packages/shared/src/libs/helpers/get-start-of-day/get-start-of-day.helper.ts
@@ -1,6 +1,6 @@
 import { startOfDay } from "date-fns";
 
-const getStartOfDay = (date: Date): Date => {
+const getStartOfDay = (date: Date | string): Date => {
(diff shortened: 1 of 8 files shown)


packages/shared/src/libs/helpers/get-start-of-day/get-start-of-day.helper.ts
```diff
@@ -1,6 +1,6 @@
 import { startOfDay } from "date-fns";
 
-const getStartOfDay = (date: Date): Date => {
+const getStartOfDay = (date: Date | string): Date => {
```
liza-veis, line 3: we should limit options in our custom helpers and use only one common format for payload, so we either need to pass date or string here. In this case it's better to pass date, not to care about date parsing in this helper. The same for getEndOfDay helper
```suggestion
const getStartOfDay = (date: Date): Date => {
```

liza-veis (changes requested):

liza-veis (approved):

PR #526, opened 2024-09-25
feat: add analytics last sync time to project details page gf-502

AnutkaGn:
When the analytics is set up, remove the setup analytics button from the UI and add the button Setup Again. 
Display the last update time and the user by whom analytics was uploaded.
When the last update was more than 5 hours ago, the label with sync time and user will be red, and on hover, there will be a tooltip.
image


apps/backend/src/modules/projects/project.entity.ts
```diff
@@ -9,40 +9,48 @@ class ProjectEntity implements Entity {
...
 		description,
 		id,
 		lastActivityDate,
+		lastActivityUserName,
 		name,
 	}: {
 		description: string;
 		id: null | number;
 		lastActivityDate: null | string;
+		lastActivityUserName: null | string;
```
liza-veis, line 26: last activity date is a column to track the last commit activity, not analytics sync, for analytics sync we need to save the date and time of sync and the user who is using the api key

```
analyticsLastSyncedAt: null | string;
analyticsLastSyncedByUserId: null | number
```

apps/frontend/src/assets/images/icons/sync.svg
liza-veis, line 1: we need only one icon, and color should be passed using css as for other icons

liza-veis (changes requested):

apps/frontend/src/assets/images/icons/sync-overdue.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M9 3C6.93858 3 5.11881 4.0397 4.03832 5.625H6V7.125H1.5V2.625H3V4.49952C4.36786 2.67875 6.54573 1.5 9 1.5C13.1421 1.5 16.5 4.85786 16.5 9H15C15 5.68629 12.3137 3 9 3ZM3 9C3 12.3137 5.68629 15 9 15C11.0615 15 12.8812 13.9603 13.9617 12.375H12V10.875H16.5V15.375H15V13.5005C13.6321 15.3212 11.4543 16.5 9 16.5C4.85786 16.5 1.5 13.1421 1.5 9H3Z" fill="#ca4925"/>
```
kolibri753, line 2: we agreed on using `fill="currentColor"`

apps/frontend/src/assets/images/icons/sync.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M9 3C6.93858 3 5.11881 4.0397 4.03832 5.625H6V7.125H1.5V2.625H3V4.49952C4.36786 2.67875 6.54573 1.5 9 1.5C13.1421 1.5 16.5 4.85786 16.5 9H15C15 5.68629 12.3137 3 9 3ZM3 9C3 12.3137 5.68629 15 9 15C11.0615 15 12.8812 13.9603 13.9617 12.375H12V10.875H16.5V15.375H15V13.5005C13.6321 15.3212 11.4543 16.5 9 16.5C4.85786 16.5 1.5 13.1421 1.5 9H3Z" fill="#7E8082"/>
```
kolibri753, line 2: same here `fill="#7E8082"` -->  `fill="currentColor"`

packages/shared/src/libs/helpers/get-sync-time/get-sync-time.helper.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type SyncTimeResult = {
+	hoursDifference?: number;
+	label: string;
+};
+
+const getSyncTime = (date: Date, baseDate: Date): SyncTimeResult => {
+	const HOURS_IN_A_DAY = 24;
+	const ONE_HOUR = 1;
+	const ZERO_HOURS = 0;
```
perekotypole, line 11: I think you can take the constants beyond the function
```suggestion
const HOURS_IN_A_DAY = 24;
const ONE_HOUR = 1;
const ZERO_HOURS = 0;

const getSyncTime = (date: Date, baseDate: Date): SyncTimeResult => {
```

apps/frontend/src/pages/project/libs/components/contributors-list/libs/components/sync-time/sync-time.tsx
```diff
@@ -0,0 +1,61 @@
...
+	const lastActivity = lastActivityDate
+		? getStartOfDay(new Date(lastActivityDate))
+		: null;
+
+	const lastUpdateLabel = lastActivity
+		? getSyncTime(lastActivity, currentDate)
+		: null;
+
+	useEffect(() => {
+		if (lastUpdateLabel && lastUpdateLabel.hoursDifference !== undefined) {
```
perekotypole, line 29: Try if it will work like this
```suggestion
		if (lastUpdateLabel?.hoursDifference !== undefined) {
```

apps/backend/src/modules/projects/project.entity.ts
AnutkaGn: I'm a bit confused about where this analyticsLastSyncedByUserId, analyticsLastSyncedAt data should come from

what1s1ove (approved):
lgtm after Lisa's comments

apps/backend/src/modules/projects/project.entity.ts
liza-veis: you take user name correctly, analyticsLastSyncedAt should be the date this log was created. For user created it's better to return it with id:
```
analyticsLastSyncedByUser: null | {
  id: number,
  name: string 
}
```

liza-veis (approved):

PR #539, opened 2024-09-25
fix: incorrect message displayed when all contributors are hidden gf-538

kolibri753:
- added dynamic emptyPlaceholder message based on hasSearch check;
- updated <Table> to use the emptyPlaceholderMessage.

### DIFF
--- a/apps/frontend/src/pages/analytics/libs/constants/constants.ts
+++ b/apps/frontend/src/pages/analytics/libs/constants/constants.ts
@@ -1,3 +1,4 @@
 export { ANALYTICS_DATE_MAX_RANGE } from "./analytics-date-max-range.constant.js";
 export { ANALYTICS_DEFAULT_DATE_RANGE } from "./analytics-default-date-range.constant.js";
 export { ANALYTICS_LOOKBACK_DAYS_COUNT } from "./analytics-lookback-days-count.constant.js";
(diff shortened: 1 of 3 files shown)


apps/frontend/src/pages/analytics/libs/constants/constants.ts
```diff
@@ -1,3 +1,4 @@
 export { ANALYTICS_DATE_MAX_RANGE } from "./analytics-date-max-range.constant.js";
 export { ANALYTICS_DEFAULT_DATE_RANGE } from "./analytics-default-date-range.constant.js";
 export { ANALYTICS_LOOKBACK_DAYS_COUNT } from "./analytics-lookback-days-count.constant.js";
+export { EMPTY_LENGTH } from "@git-fit/shared";
```
liza-veis, line 4: this should be re-exported from src/libs/constants not from this module

liza-veis (changes requested):

what1s1ove (approved):
lgtm after Lisa's comments

apps/frontend/src/pages/analytics/libs/constants/constants.ts
kolibri753: oops, thanks

liza-veis (approved):

PR #542, opened 2024-09-25
fix: Browser Back button not functioning after redirecting to the Analytics page from the Project details page gf-540

s1rserg:
* Browser Back button is now working after redirecting to the Analytics page on contributor item click.
* Added search preselected for contributor name.
* Fixed activity chart not clickable.

### DIFF
--- a/apps/frontend/src/libs/components/chart/chart.tsx
+++ b/apps/frontend/src/libs/components/chart/chart.tsx
@@ -6,11 +6,17 @@ import styles from "./styles.module.css";
 
 type Properties = {
 	data: ChartData;
+	isCursorPointer: boolean;
 };
 
-const Chart = ({ data }: Properties): JSX.Element => {
+const Chart = ({ data, isCursorPointer }: Properties): JSX.Element => {
 	return (
-		<LineChart data={data} height={40} width={120}>
+		<LineChart
+			data={data}
+			height={40}
+			style={{ cursor: isCursorPointer ? "pointer" : "default" }}
+			width={120}
+		>
 			<Line className={styles["line"] as string} dataKey="y" dot={false} />
 		</LineChart>
 	);
--- a/apps/frontend/src/libs/hooks/use-search-filters/use-search-filters.hook.ts
+++ b/apps/frontend/src/libs/hooks/use-search-filters/use-search-filters.hook.ts
@@ -20,7 +20,15 @@ const useSearchFilters = ({
 		: "";
 	const [search, setSearch] = useState<string>(searchParameter);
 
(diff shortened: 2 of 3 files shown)


what1s1ove (approved):

liza-veis (approved):

apps/frontend/src/libs/hooks/use-search-filters/use-search-filters.hook.ts
```diff
@@ -20,7 +20,15 @@ const useSearchFilters = ({
 		: "";
 	const [search, setSearch] = useState<string>(searchParameter);
 
+	const [isInitialLoad, setIsInitialLoad] = useState<boolean>(true);
```
liza-veis, line 23: Are you sure this works as expected? I was thinking maybe the problem in place where we call navigate? like there different methods for navigation, push, replace etc

s1rserg: It works only for the redirection from click on contributor item to analytics page. But for example when user is on contributors page and then clicks on sidebar to get to access management page and tries to click back it doesn't go back. It appears to be related to query parameters. I'm unsure if this qualifies as a bug, as the QA team only mentioned that specific redirection in the ticket, so they may consider the current behavior acceptable for these pages.

apps/frontend/src/libs/components/chart/chart.tsx
```diff
@@ -6,11 +6,17 @@ import styles from "./styles.module.css";
...
 };
 
-const Chart = ({ data }: Properties): JSX.Element => {
+const Chart = ({ data, isCursorPointer }: Properties): JSX.Element => {
 	return (
-		<LineChart data={data} height={40} width={120}>
+		<LineChart
+			data={data}
+			height={40}
+			style={{ cursor: isCursorPointer ? "pointer" : "default" }}
```
GvoFor, line 17: It's better to style through the className

s1rserg: I tried. It doesn't apply it this way.
Image from our dev route:
image

GvoFor: Oh, maybe then let's add `cursor: pointer` style to the `NavLink` with which you wrap chart?

GvoFor: And leave this component without changes

s1rserg: Tried this too. Styles applied to parent elements do not override or cascade down to child elements for the `cursor` property.

yevhenii-hladkov (approved):

PR #559, opened 2024-09-27
fix: Project groups are not deleted from the DB after deleting the project they corresponds to gf-504

GvoFor:
## Features
- added migration which deletes existent unused project groups
- added project group deletion to project deletion flow on service level


apps/backend/src/db/migrations/20240927085349_add_triger_to_remove_unused_project_groups_on_project_delete.ts
kolibri753, line 1: a little typo in name: trigger*

GvoFor: Thanks for noticing

what1s1ove (approved):

apps/backend/src/db/migrations/20240927085349_remove_unused_project_groups_and_add_trigger_to_remove_them_on_project_delete.ts
liza-veis, line 1: can't we just delete projects_to_project_groups by cascade when project is deleted and then delete project groups itself on the service level (because of how we implemented this)

liza-veis (changes requested):

apps/backend/src/db/migrations/20240927085349_remove_unused_project_groups_and_add_trigger_to_remove_them_on_project_delete.ts
GvoFor: Yes, we can. Oh, I get the point. I just was thinking that it is DB responsible to delete relative project group on project deletion, and do it this way. But this could be correct in case we have one to many relationship, but we have many to many, right?

I'll change this

liza-veis: yes, it's correct

liza-veis (approved):

PR #561, opened 2024-09-27
feat: fix quality criteria issues gf-557

Vitaliistf:
Fixed: 
- Inconsistent naming. Folder and file name did not correspond to hook name.
- Spelling mistakes.
- Inconsistent orderBy columns names case. Changed them to camelCase.
- Inconsistent file structure (helper has its own folder).
- Inconsistent icon filling.

Besides fixes that were described in ticket, I've changed case of all table and column names in objection queries (except for raw parts of queries of course).


apps/frontend/src/assets/images/icons/check.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg width="18" height="18" viewBox="0 0 18 18" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
perekotypole, line 1: I have only one question. Why do we need to replace `fill="none"`?

Vitaliistf: It's just for consistency. Either "currentColor" or "none". I chose currentColor because some icons don't have fill in path tag. Maybe it will be not so confusing as it would be if we had none (and no path).

liza-veis (approved):

liza-veis:
Could you also please fix OrderKey from @GvoFor pull request?

Vitaliistf:
> Could you also please fix OrderKey from @GvoFor pull request?

Oh, yeah, sorry, I've changed it but forgot to add to commit😅

what1s1ove (approved):

PR #566, opened 2024-09-27
fix: Name field allows special characters, numbers, Cyrillic in Profile Update gf-562

s1rserg:
* Updated user patch validation schema.
image
image
image


s1rserg:
Regarding trailing whitespaces I can add negative lookahead to name pattern regex to forbid creating/updating usernames which have consecutive whitespaces inside.

what1s1ove (approved):

PR #568, opened 2024-09-27
feat: Add a loading indicator with a tick for the API Key Generation field at the Setup Analytics modal gf-189

perekotypole:
idle view:
image

fulfilled view:
image


apps/frontend/src/pages/project/libs/components/setup-analytics-modal/setup-analytics-modal.tsx
```diff
@@ -129,6 +134,15 @@ const SetupAnalyticsModal = ({
...
 						placeholder="No API key"
+						{...(isKeyGenerated
+							? {
+									leftIcon: (
+										<div className={styles["generated-key-indicator"]}>
+											<Icon height={20} name="check" width={20} />
+										</div>
+									),
+								}
+							: {})}
```
perekotypole, line 145: @liza-veis I'm not sure about this
```
									leftIcon: {
										isKeyGenerated ? <div className={styles["generated-key-indicator"]}>
											<Icon height={20} name="check" width={20} />
										</div> : <></> (or undefined)
									},
```
The code above looks better, but it doesn't work properly. So it's a solution, but I don't like the way the code looks.

liza-veis: we can use && here instead of ?, this will simplify it a bit

perekotypole: @liza-veis I tried, but...
image

liza-veis: in this case, let's leave it as it is

liza-veis (approved):

PR #571, opened 2024-09-27
fix: Three Toast notifications are displayed on attempt to open a project page without a permission gf-567

s1rserg:
* Added has view permission check in order to not load all contributors when user doesn't have neede permissions.
image

### DIFF
--- a/apps/frontend/src/pages/project/project.tsx
+++ b/apps/frontend/src/pages/project/project.tsx
@@ -119,14 +119,38 @@ const Project = (): JSX.Element => {
 	const [contributorToSplit, setContributorToSplit] =
 		useState<ContributorGetAllItemResponseDto | null>(null);
 
+	const hasViewPermission =
+		checkHasPermission(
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/project/project.tsx
```diff
@@ -119,14 +119,38 @@ const Project = (): JSX.Element => {
 	const [contributorToSplit, setContributorToSplit] =
 		useState<ContributorGetAllItemResponseDto | null>(null);
 
+	const hasViewPermission =
+		checkHasPermission(
+			[PermissionKey.VIEW_ALL_PROJECTS, PermissionKey.MANAGE_ALL_PROJECTS],
+			userPermissions,
+		) ||
+		checkIsProjectPermitted({
```
liza-veis, line 127: maybe we can pass array or permissions to checkIsProjectPermitted 🤔

s1rserg: Yeah. It`s local helper, so I just refactored it to receive array of permissions.

PR #572, opened 2024-09-27
fix: Manage Project permission does not allow user to delete a Project group in this specific project on the Manage Access page gf-569

perekotypole:
- replaced `/:id` to `/:projectId/:id` routes in patch and delete project group to identify project for permissions
- added missing "Manage Project" permission to patch project group


packages/shared/src/modules/project-groups/libs/enums/project-groups-api-path.enum.ts
```diff
@@ -1,5 +1,6 @@
 const ProjectGroupsApiPath = {
-	$ID: "/:id",
+	$ID: ":id",
+	$PROJECT_ID_$ID: "/:projectId/:id",
```
liza-veis, line 3: what if project id passed doesn't match project group? this project id should be extracted on the backend side from project group
```suggestion
```

liza-veis (changes requested):

packages/shared/src/modules/project-groups/libs/enums/project-groups-api-path.enum.ts
perekotypole: I simplified it because of this functional
image

The third parameter takes project id. On the other hand, I have to copy the functionality of the whole `checkUserPermissions` function somewhere else after taking the project id from the group,  and somehow pass the permissions from the route. I'd rather add another check to the service, and if project group is not included to the project return `NO_PERMISSION`

liza-veis: in this case let's pass projectId to payload

perekotypole: I just realised that DELETE method doesn't have body and that's why I pass it in params.

perekotypole: and I'm trying to implement it, so it doesn't matter how you pass `project id` (but we still need to take it in router)
```
	public async deleteWithCheckingProject(
		id: number,
		projectId: number,
	): Promise<boolean> {
		const projectGroup = await this.projectGroupRepository.find(id);

		if (!projectGroup) {
			throw new ProjectGroupError({
				message: ExceptionMessage.PROJECT_GROUP_NOT_FOUND,
				status: HTTPCode.NOT_FOUND,
			});
		}

		const { projectId: currentProject } = projectGroup.toObject();

		if (currentProject.id !== projectId) {
			throw new ProjectGroupError({
				message: ExceptionMessage.NO_PERMISSION,
				status: HTTPCode.FORBIDDEN,
			});
		}

		return await this.projectGroupRepository.delete(id);
	}
```

liza-veis: maybe we can remove this hooks for project group and calculate them inside service? I think this will be a more secure and right option

apps/backend/src/libs/helpers/helpers.ts
```diff
@@ -1,3 +1,4 @@
+export { checkUserPermissions as checkUserPermissionsHelper } from "./check-user-permissions.helper.js";
```
liza-veis, line 1: Maybe we can name it as checkProjectPermission instead?

perekotypole: it checks all permissions

perekotypole: it checks all permissions

perekotypole: I changed it to `checkPermissions`, because I just replaced the functional from `checkUserPermissions` hook to not copy the code

liza-veis (approved):

PR #578, opened 2024-09-27
fix: the delete project button is displayed and the deletion is accessible for the manage specific project permission gf-570

Fjortis:
- Removed delete button showing for users with any of the Projectpermissions.
- Adjusted backend to limit deletion for users with Projectpermissions

Test

### DIFF
--- a/apps/frontend/src/pages/project/project.tsx
+++ b/apps/frontend/src/pages/project/project.tsx
@@ -315,11 +315,10 @@ const Project = (): JSX.Element => {
 			projectUserPermissions,
 		});
 
-	const hasManageAllProjectsPermission =
-		checkHasPermission(
-			[PermissionKey.MANAGE_ALL_PROJECTS],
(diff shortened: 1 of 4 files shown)


perekotypole:
Shouldn't we control the access on backend as well?

apps/frontend/src/pages/project/project.tsx
```diff
@@ -378,8 +383,8 @@ const Project = (): JSX.Element => {
 						<div className={styles["project-header"]}>
 							<h1 className={styles["title"]}>{project.name}</h1>
 							<ProjectDetailsMenu
+								hasDeleteProjectsPermissions={hasDeleteProjectsPermissions}
```
liza-veis, line 386:
```suggestion
								hasDeleteProjectsPermission={hasDeleteProjectsPermission}
```

liza-veis (approved):

PR #590, opened 2024-10-02
feat: change script to support more than 1 git repository gf-589

ArtemZag:


scripts/analytics/src/modules/analytics/analytics.service.ts
```diff
@@ -116,4 +122,50 @@ class AnalyticsService {
 	}
 }
 
+function mergeStats(
```
liza-veis, line 125: We keep all helpers separately in a separate folder in libs/helpers, for example
```
/scripts/analytics/src/modules/analytics/libs/helpers/merge-stats/merge-stats.helper.ts
/scripts/analytics/src/modules/analytics/libs/helpers/merge-by-criteria/merge-by-criteria.helper.ts
```
And then they should be re-exported from `/scripts/analytics/src/modules/analytics/libs/helpers/helper.ts`

Take a look at this structure here `/apps/backend/src/libs/plugins/authorization/`

liza-veis: Also, we use arrow functions instead of function declaration, except for migrations because it's generated by default

liza-veis (approved):

yevhenii-hladkov (approved):

